PR #1, opened 2018-01-24
Dockerfiles for .NET Core builds

ravimeda:
These Dockerfiles are used to prepare the images needed for .NET Core official builds that are performed in a Docker container.

Fixes https://github.com/dotnet/core-eng/issues/2440

### DIFF
--- a/src/build.ps1
+++ b/src/build.ps1
@@ -40,9 +40,9 @@ $repoRoot = Split-Path -Path "$PSScriptRoot" -Parent
 & docker run --rm `
     -v /var/run/docker.sock:/var/run/docker.sock `
     -v "${repoRoot}:/repo" `
-    -w /repo/dockerfiles `
+    -w /repo/src `
     $ImageBuilderImageName `
-    build --manifest "/repo/dockerfiles/manifest.json" --path "$DockerfilePath" "$ImageBuilderCustomArgs"
+    build --manifest "/repo/src/manifest.json" --path "$DockerfilePath" "$ImageBuilderCustomArgs"
 
 if ($LastExitCode -ne 0) {
     throw "Failed executing ImageBuilder."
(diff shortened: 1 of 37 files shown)


src/build.ps1
```diff
@@ -0,0 +1,49 @@
...
+
+ExecuteWithRetry docker pull $ImageBuilderImageName
+$repoRoot = Split-Path -Path "$PSScriptRoot" -Parent
+
+& docker run --rm `
+    -v /var/run/docker.sock:/var/run/docker.sock `
+    -v "${repoRoot}:/repo" `
+    -w /repo/dockerfiles `
+    $ImageBuilderImageName `
+    build --manifest "/repo/dockerfiles/manifest.json" --path "$DockerfilePath" "$ImageBuilderCustomArgs"
```
ravimeda: `/repo/dockerfiles/manifest.json` - that needs to be updated.

ravimeda: Replace as -

```powershell
& docker run --rm `
    -v /var/run/docker.sock:/var/run/docker.sock `
    -v "${repoRoot}:/repo" `
    -w /repo/src `
    $ImageBuilderImageName `
    build --manifest "/repo/src/manifest.json" --path "$DockerfilePath" "$ImageBuilderCustomArgs"
```

src/build.ps1
```diff
@@ -0,0 +1,49 @@
+[cmdletbinding()]
```
MichaelSimons: I think this should reside at the root of the repo.

src/manifest.json
```diff
@@ -0,0 +1,328 @@
+{
```
MichaelSimons: I kind of feel like this belongs at the root of the repo as well.

src/build.ps1
ravimeda: Agree.  Will update.

src/manifest.json
ravimeda: Yes, it would make it consistent with other Docker repos.

PR #6, opened 2018-01-26
Add testrunner Dockerfile

ravimeda:
Dockerfile that enables running PowerShell scripts.  This is needed for official builds.

### DIFF
--- a/Dockerfile.linux.testrunner
+++ b/Dockerfile.linux.testrunner
@@ -1,0 +1,7 @@
+# Use this Dockerfile to create a Linux testrunner image
+# Usage: docker run --rm -v /var/run/docker.sock:/var/run/docker.sock testrunner pwsh -File build.ps1 <params....>
+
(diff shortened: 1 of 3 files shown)


MichaelSimons:
Can you also make the `--pull` tweak to the build script that we discussed?

Dockerfile.linux.testrunner
```diff
@@ -0,0 +1,7 @@
+# Use this Dockerfile to create a Linux testrunner image
```
MichaelSimons: Thinking about the name, I think the term "test" is confusing.  I think the more abstract `runner` term is better.  This isn't being used for tests.

ravimeda: Valid point.  I'll update the comment as - `# Use this Dockerfile to create a Linux runner image`

ravimeda:
Yes, I will update `build.ps1` to include pull -
`& docker build -t imagebuilder --pull -f ./Dockerfile.linux.imagebuilder .`

PR #8, opened 2018-01-27
Split ImageBuilderCustomArgs

ravimeda:
ImageBuilder is parsing `ImageBuilderCustomArgs` as a string.   `ImageBuilderCustomArgs` is a collection of argument.  Hence split the string.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -26,7 +26,7 @@ try {
     & docker run --rm `
         -v /var/run/docker.sock:/var/run/docker.sock `
         imagebuilder `
-        build --manifest "manifest.json" --path "$DockerfilePath" "$ImageBuilderCustomArgs"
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -26,7 +26,7 @@ try {
     & docker run --rm `
         -v /var/run/docker.sock:/var/run/docker.sock `
         imagebuilder `
-        build --manifest "manifest.json" --path "$DockerfilePath" "$ImageBuilderCustomArgs"
+        build --manifest manifest.json --path "$DockerfilePath" ${ImageBuilderCustomArgs}.Split()
```
MichaelSimons: I don't think calling split is a good idea.  The ImageBuilderCustomArgs may include quoted values itself.  Splitting in this scenario would not produce the desired behavior.  Why not use InvokeExpression instead?

PR #10, opened 2018-01-31
Symlink post-build hooks

ravimeda:
Symlink post-build hooks to the cleanup script.  Ran the following command under  `src/ubuntu/14.04/cross/hooks` and `src/ubuntu/16.04/cross/hooks`.

```shell
ln -s ../../../build-scripts/build-rootfs-cleanup.sh post-build
```


MichaelSimons:
The PR description could be enhanced a little.  The issue being fixed is that the Unbuntu post-build hooks are currently broken. They were originally defined as symlinks but the link was broken when the files were moved across repos.

PR #30, opened 2018-03-17
Add Fedora27 dockerfile

ianhays:
cc: @dagood @joperezr @livarcocc @janvorli

### DIFF
--- a/src/fedora/27/Dockerfile
+++ b/src/fedora/27/Dockerfile
@@ -1,0 +1,47 @@
+FROM fedora:27
+
+# Workaround to avoid NuGet restore timeout
+RUN dnf upgrade -y nss \
+    && dnf clean all
+
+# Install the base toolchain we need to build anything (clang, cmake, make and the like)
+# this does not include libraries that we need to compile different projects, we'd like
+# them in a different layer.
+RUN dnf install -y \
+        clang \
+        cmake \
+        lldb-devel \
+        llvm-devel \
+        make \
+        python \
+        which \
+    && dnf clean all
+
+# Install tools used by the VSO build automation.
+RUN dnf install -y \
+        git \
+        npm \
+        nodejs \
+        tar \
+        zip \
+    && dnf clean all
+
+RUN npm install -g azure-cli --unsafe-perm
+
+# Dependencies of CoreCLR and CoreFX.
+RUN dnf install -y \
+        glibc-locale-source \
+        libcurl-devel \
+        libgdiplus \
+        libicu-devel \
+        libunwind-devel \
+        libuuid-devel \
+        lttng-ust-devel \
+        openssl-devel \
+        uuid-devel \
+    && dnf clean all
+
+RUN dnf install -y compat-openssl10-devel --allowerasing
+
+# Until Fedora.27-x64 official packages are available, we have to restore the ubuntu ones instead.
+RUN export __PUBLISH_RID=ubuntu.14.04-x64
--- a/src/fedora/27/Dockerfile
+++ b/src/fedora/27/Dockerfile
@@ -44,4 +44,4 @@ RUN dnf install -y \
 RUN dnf install -y compat-openssl10-devel --allowerasing
 
 # Until Fedora.27-x64 official packages are available, we have to restore the ubuntu ones instead.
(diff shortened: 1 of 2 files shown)


src/fedora/27/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+        libgdiplus \
+        libicu-devel \
+        libunwind-devel \
+        libuuid-devel \
+        lttng-ust-devel \
+        openssl-devel \
+        uuid-devel \
+    && dnf clean all
+
+RUN dnf install -y compat-openssl10-devel --allowerasing
```
joperezr: Why not add these dependencies to one of the restores above?

src/fedora/27/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# Until Fedora.27-x64 official packages are available, we have to restore the ubuntu ones instead.
+RUN export __PUBLISH_RID=ubuntu.14.04-x64
```
joperezr: I would use the ‘ENV’ command instead of RUN export, that way the actual environment variable is preserved.

joperezr:
Cc @MichaelSimons

src/fedora/27/Dockerfile
ianhays: I thought it would be cleaner to have the abnormal ones (i.e. the compat installs) in their own command.

src/fedora/27/Dockerfile
ianhays: thanks. Was wondering why the variable wasn't working...

src/fedora/27/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+        libuuid-devel \
+        lttng-ust-devel \
+        openssl-devel \
+        uuid-devel \
+    && dnf clean all
+
+RUN dnf install -y compat-openssl10-devel --allowerasing
+
+# Until Fedora.27-x64 official packages are available, we have to restore the ubuntu ones instead.
+ENV __PUBLISH_RID=ubuntu.14.04-x64
```
joperezr: NIT: I thought the syntax for 'ENV' required a space instead of an '=' but if this has the desired effect and compiles then that should be fine.

ianhays: It can go either way, with or without the equals sign.

src/fedora/27/Dockerfile
dagood: These aren't just commands, they're layers, and you should in general use as few as you can:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

src/fedora/27/Dockerfile
MichaelSimons: https://docs.docker.com/engine/reference/builder/#env - both are supported, when `=` is used you can set more than one environment variable with a single `ENV` instruction.

```
ENV myName="John Doe" myDog=Rex\ The\ Dog \
        myCat=fluffy
```

src/fedora/27/Dockerfile
ianhays: Interesting, thanks for the link. I combined them.

src/fedora/27/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN dnf install -y \
+        glibc-locale-source \
+        libcurl-devel \
+        libgdiplus \
+        libicu-devel \
+        libunwind-devel \
+        libuuid-devel \
+        lttng-ust-devel \
+        uuid-devel \
+        compat-openssl10-devel \
```
MichaelSimons: The standard Dockerfile format is to alphabetize the package lists in order to make it easier to read.  Is there a reason this can't get alphabetized?

ianhays: There is not. Updated.

MichaelSimons:
microsoft/dotnet-buildtools-prereqs:fedora-27-82a3800-20180326211504
microsoft/dotnet-buildtools-prereqs:fedora-24-d770b8b-20180326211504
microsoft/dotnet-buildtools-prereqs:fedora-23-d770b8b-20180326211504

PR #32, opened 2018-03-17
Enable building Ubuntu 18.04 build tools images

rakeshsinghranchi:
Enable building Ubuntu 18.04 build image


src/ubuntu/18.04/debpkg/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apt-get update \
+    && apt-get -y install \
+        debhelper \
+        build-essential \
+        devscripts \
+    && rm -rf /var/lib/apt/lists/*
+
+# liblldb is needed so deb package build does not throw missing library info errors
+# Package resource list is already set in the base Image /etc/apt/sources.list.d/llvm.list
+RUN apt-get update \
```
MichaelSimons: I realize this is the previous pattern used - but is there a specific reason to install liblldb in it's own layer?

src/ubuntu/18.04/debpkg/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM microsoft/dotnet-buildtools-prereqs:ubuntu-18.04
+
+# Install the deb packaging toolchain we need to build debs
+RUN apt-get update \
+    && apt-get -y install \
+        debhelper \
+        build-essential \
```
MichaelSimons: Should be alphabetized.

src/ubuntu/18.04/debpkg/Dockerfile
rakeshsinghranchi: I am not aware of any specific reason. Noticed that base layer installs liblldb-3.9-dev.

src/ubuntu/18.04/debpkg/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM microsoft/dotnet-buildtools-prereqs:ubuntu-18.04TO
```
MichaelSimons: What is `ubuntu-18.04TO`?

src/ubuntu/18.04/debpkg/Dockerfile
MichaelSimons: My preference would be to combine the two.  It will simplify everything - build, pull, and startup time.

rakeshsinghranchi: Yes, Makes sense.

rakeshsinghranchi:
@MichaelSimons , PR tests are failing unrelated to my change. What is needed to re-run the tests ? Also, I don't see ubuntu_18 test, do I need to add https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/netci.groovy#L6 list ?

src/ubuntu/18.04/debpkg/Dockerfile
MichaelSimons: I wasn't proposing the removal of the debpkg image.  I was proposing combining the two layers within the debpkg image.  I think keeping the debian packaging dependencies separated is valuable.

MichaelSimons:
@rakeshsinghranchi - 

> What is needed to re-run the tests

You can trigger rerunning of tests just like any other jenkins based dotnet repo - e.g. `@dotnet-bot test centos_6 please`

> do I need to add ubuntu18 to the list here https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/netci.groovy#L6 list ?

Yes - ubuntu versions were broken out into individual legs because of how long the cross images take to build.  Even though the cross images don't exist for 18.04 yet, I would guess they will eventually get added.

rakeshsinghranchi:
@dotnet-bot test centos_7 please
@dotnet-bot test fedora please

src/ubuntu/18.04/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+        libicu-dev \
+        libkrb5-dev \
+        liblttng-ust-dev \
+        libssl-dev \
+        libunwind8 \
+        libunwind8-dev \
+        uuid-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install the deb packaging toolchain we need to build debs
```
MichaelSimons: In my previous comment, I wasn't proposing the removal of the debpkg image. I was proposing combining the two layers within the debpkg image. I think keeping the debian packaging dependencies separated is valuable.

rakeshsinghranchi:
@MichaelSimons  , I don't have merge permission. Can you please merge ? Also, what is the process around producing the images ?

MichaelSimons:
I kicked off a build and will update the PR with the new image names when the build finishes.

MichaelSimons:
microsoft/dotnet-buildtools-prereqs:ubuntu-18.04-66216d3-20180319211855
microsoft/dotnet-buildtools-prereqs:ubuntu-18.04-debpkg-66216d3-20180319211855

rakeshsinghranchi:
Thanks @MichaelSimons

janvorli:
@rakeshsinghranchi, @MichaelSimons  this dockerfile doesn't create correct image. It installs openssl 1.1 and curl4 dependent on that, since the openssl1.1 is now the default on Ubuntu 18.04. We don't support openssl 1.1 and we require openssl 1.0.
I have created and tested a Dockerfile for Ubuntu 18.04 yesterday when I was creating and testing the bootstrap cli for Ubuntu 18.04. I'll send a PR for it in a minute.

PR #34, opened 2018-03-20
Fix the Ubuntu 18.04 Dockerfile

janvorli:
There are several issues:
* The default openssl in Ubuntu 18.04 is openssl 1.1 that we don't support.
  We need to explicitly install openssl 1.0 and curl depending on that
* CMake package has hard dependency on the curl package with dependency on
  openssl 1.1. So we need to install cmake from a tarball downloaded from
  the cmake site instead.

### DIFF
--- a/src/ubuntu/18.04/Dockerfile
+++ b/src/ubuntu/18.04/Dockerfile
@@ -5,17 +5,17 @@ FROM ubuntu:18.04
 # them in a different layer.
 RUN apt-get update \
     && apt-get install -y wget \
-    && echo "deb http://llvm.org/apt/bionic/ llvm-toolchain-bionic main" | tee /etc/apt/sources.list.d/llvm.list \
-    && wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key | apt-key add - \
     && apt-get update \
     && apt-get install -y \
         clang-3.9 \
-        cmake \
         liblldb-3.9-dev \
         lldb-3.9 \
         llvm-3.9 \
         make \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/* \
+    && wget https://cmake.org/files/v3.10/cmake-3.10.2-Linux-x86_64.tar.gz \
+    && tar -xf cmake-3.10.2-Linux-x86_64.tar.gz --strip 1 -C /usr/local \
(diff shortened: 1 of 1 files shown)


janvorli:
cc: @rakeshsinghranchi, @MichaelSimons - for some reason, I cannot add reviewers and assign the PR to myself.

src/ubuntu/18.04/Dockerfile
```diff
@@ -5,17 +5,17 @@ FROM ubuntu:18.04
 # them in a different layer.
 RUN apt-get update \
     && apt-get install -y wget \
```
MichaelSimons: I think it would make sense to merge the first and second apt-get update and apt-get install calls.

janvorli: I've just mimicked what the other Dockerfiles do. I am fine merging it. Any reason for merging just the first two and not the third one too?

MichaelSimons: With your refactoring, there is no need to retrieve `wget` separately before the other packages.  Additionally there is no point in the second `apt-get update` call because the apt sources hasn't changed.

```
RUN apt-get update \
    && apt-get install -y wget \
    && apt-get update \
    && apt-get install -y \
        clang-3.9 \
        liblldb-3.9-dev \
        lldb-3.9 \
        llvm-3.9 \
        make \
```

Suggestion:
RUN apt-get update \
    && apt-get install -y \
        clang-3.9 \
        liblldb-3.9-dev \
        lldb-3.9 \
        llvm-3.9 \
        make \
        wget \
```

janvorli: Actually, thinking about it more, it seems that if we want to do that, it would make sense to make it as a sweeping change over all the Dockerfiles we have to keep them in line.

janvorli: Ah, I am sorry, I've missed the real point. Ok, I'll do that.

janvorli: Done

MichaelSimons:
microsoft/dotnet-buildtools-prereqs:ubuntu-18.04-f90bc20-20180320152209
microsoft/dotnet-buildtools-prereqs:ubuntu-18.04-debpkg-66216d3-20180320152209

PR #40, opened 2018-04-03
Adding iputils to helix alpine image

joperezr:
Adding iputils dependency so that we fix issue https://github.com/dotnet/corefx/issues/28204

cc: @MichaelSimons


joperezr:
Merging since the failures doesn't seem to be related to my alpine changes and alpine build did pass.

PR #41, opened 2018-04-03
Build liblldb 3.8 from sources

janvorli:
The OpenSUSE 42.3 doesn't have official package liblldb-devel that
we need to build lldb plugin for the lldb that's available for this
distro. So we build the devel library from sources.
This is only build time library, the lldb package contains the runtime
library, so the plugin can work without any extra work on the user
side.

### DIFF
--- a/src/opensuse/42.3/Dockerfile
+++ b/src/opensuse/42.3/Dockerfile
@@ -25,6 +25,58 @@ RUN zypper -n install \
         zip \
     && zypper clean -a
 
+# Build and install liblldb development files
+RUN zypper -n install \
+    doxygen \
+    libedit-devel \
+    libxml2-devel \
+    python-argparse \
+    python-devel \
+    readline-devel \
+    swig && \
+    \
+    wget http://releases.llvm.org/3.8.1/cfe-3.8.1.src.tar.xz && \
+    wget http://releases.llvm.org/3.8.1/llvm-3.8.1.src.tar.xz && \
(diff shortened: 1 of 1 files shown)


janvorli:
CC: @ianhays, @MichaelSimons

ianhays:
Looks good, thanks Jan! We'll need to submit a PR to the build pipeline file with the new dockertag after this is merged and make sure to remove the part where i pass an arg to disable building the plugin.

src/opensuse/42.3/Dockerfile
```diff
@@ -43,3 +95,6 @@ RUN zypper -n install --force-resolution \
 
 # Until OpenSuse.42.3 official packages are available, we have to restore the ubuntu ones instead.
 ENV __PUBLISH_RID=ubuntu.14.04-x64
```
MichaelSimons: You may consider declaring these ENV within one Docker instruction to reduce the number of layers in the resulting image.
```
ENV env1=value1 \
        env2=value2
```

janvorli: Oh, I have not realized that even the env vars create a layer. Will do.

ianhays:
@MichaelSimons good to merge and produce some new tags?

MichaelSimons:
microsoft/dotnet-buildtools-prereqs:opensuse-42.3-e1d0fe9-20180404001623
microsoft/dotnet-buildtools-prereqs:opensuse-42.1-d770b8b-20180404001623

PR #44, opened 2018-04-04
Add idconfig to opensuse 423 build

ianhays:
cc: @janvorli @MichaelSimons

### DIFF
--- a/src/opensuse/42.3/Dockerfile
+++ b/src/opensuse/42.3/Dockerfile
@@ -98,3 +98,5 @@ ENV __PUBLISH_RID=ubuntu.14.04-x64
 # Define variables that the build uses to locate the liblldb.so and the related header.
 ENV LLDB_INCLUDE_DIR=/llvm/include \
     LLDB_LIB_DIR=/llvm/lib64
+
+RUN /sbin/ldconfig
--- a/src/opensuse/42.3/Dockerfile
+++ b/src/opensuse/42.3/Dockerfile
@@ -91,12 +91,11 @@ RUN zypper -n install --force-resolution \
         lttng-ust-devel \
         libopenssl-devel \
         uuid-devel \
-    && zypper clean -a
+    && zypper clean -a \
+    && /sbin/ldconfig
 
 # Until OpenSuse.42.3 official packages are available, we have to restore the ubuntu ones instead.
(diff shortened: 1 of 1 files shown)


src/opensuse/42.3/Dockerfile
```diff
@@ -98,3 +98,5 @@ ENV __PUBLISH_RID=ubuntu.14.04-x64
 # Define variables that the build uses to locate the liblldb.so and the related header.
 ENV LLDB_INCLUDE_DIR=/llvm/include \
     LLDB_LIB_DIR=/llvm/lib64
+
+RUN /sbin/ldconfig
```
janvorli: As @MichaelSimons usually suggests, it would be nice to attach this to the previous RUN to reduce the number of layers.

ianhays: Done.

MichaelSimons:
The following image was produced as a result of this PR:
microsoft/dotnet-buildtools-prereqs:opensuse-42.3-769626d-20180405022038

janvorli:
@MichaelSimons, @ianhays  the image has again the same issue as before, adding the ldconfig to the Dockerfile didn't help. It looks like the docker we are using to build the dockerfile has some quirk that leaves the ldconfig cache in an incorrect state. When I built the image on my Ubuntu or when Ian built it on his windows box, it was working fine.

janvorli:
To be more specific, I have used just `docker build .` command to build it on Ubuntu. Not sure what @ianhays did on Windows.

janvorli:
Yes, I am building from the tip. The docker on my Ubuntu 14.04 is version 1.6.2, build 7c8fca2. I believe I have the latest base image, but let me try with the command arguments you've provided.

janvorli:
Btw, running the ldconfig in the container created from the image you've created fixes the problem. I mean, running clang before that doesn't work - clang complains about missing symbols, after running ldconfig it works.
That's why I've asked Ian to add it to the Dockerfile hoping that it will solve the issue.

janvorli:
@MichaelSimons I've tried to use your `docker build --pull --no-cache -t opensuse-42.3 .` command and it worked ok.

ianhays:
@janvorli maybe we should just try to add it to the setup commands of the build pipeline so we can get the build passing while we try to figure out the docjer issue.

PR #51, opened 2018-05-04
Manually build libcurl for u1804

ianhays:
PTAL: @janvorli @MichaelSimons

### DIFF
--- a/src/ubuntu/18.04/Dockerfile
+++ b/src/ubuntu/18.04/Dockerfile
@@ -33,7 +33,6 @@ RUN apt-get update \
 RUN apt-get update \
     && apt-get install -y \
         gettext \
-        libcurl-openssl1.0-dev \
         libgdiplus \
         libicu-dev \
         libkrb5-dev \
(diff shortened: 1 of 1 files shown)


src/ubuntu/18.04/Dockerfile
```diff
@@ -44,3 +43,35 @@ RUN apt-get update \
         uuid-dev \
     && rm -rf /var/lib/apt/lists/*
 
+# Build and install curl 7.45.0
+RUN wget https://curl.haxx.se/download/curl-7.45.0.tar.lzma && \
```
MichaelSimons: I think it would be best to place the `&&` as the beginning of the lines to be consistent with the rest of the Dockerfile.  The `&&` at the front is a common Dockerfile practice.

ianhays: Fixed.

ianhays:
@MichaelSimons Done and ready to merge :)

MichaelSimons:
The following image was built as a result of this PR:

microsoft/dotnet-buildtools-prereqs:ubuntu-18.04-ea8a500-20180507175506

PR #57, opened 2018-07-24
Add RHEL 7.5 Dockerfile and infrastructure

dagood:
Add @dleeapho's RHEL7.5 Dockerfile. To support builds, add `-RunPrivileged` build.ps1 arg and upgrade image-builder to get https://github.com/dotnet/docker-tools/pull/109.

Unlike `microsoft/dotnet-buildtools-prereqs`, the new `dotnet-buildtools-prereqs` repo in the manifest doesn't get pushed to public Dockerhub. There is now a second build definition [dotnet-buildtools-prereqs-docker-rhel](https://devdiv.visualstudio.com/DevDiv/_build/index?context=allDefinitions&path=\dotnet-docker&definitionId=9629&_a=completed) that uses a registered Red Hat machine to build and push to a private Azure Container Registry. (RHEL images can't be pushed to public repos.)

We plan to use this for CI in source-build: https://github.com/dotnet/source-build/issues/415.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,19 @@ The Dockerfiles in this repository are used for building the .NET Core product. 
 
 ## Where are the published images
 
-The images produced from the Dockerfiles are published to the [microsoft/dotnet-buildtools-prereqs](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/) Docker Hub repository.
+Each image produced from the Dockerfiles is published based on its repository in the [manifest](#manifest):
 
-- [Most recent tags](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/tags/)
-- [Full list of tags](https://registry.hub.docker.com/v1/repositories/microsoft/dotnet-buildtools-prereqs/tags)
+- [microsoft/dotnet-buildtools-prereqs](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/) is a Docker Hub repository.
+  - [Most recent tags](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/tags/)
+  - [Full list of tags](https://registry.hub.docker.com/v1/repositories/microsoft/dotnet-buildtools-prereqs/tags)
+- **dotnet-buildtools-prereqs** is a repository in the *DotnetDocker* private Azure Container Registry.
+  - This repository is used to host images that can't be published to Docker Hub.
 
 ## How to identify an image
 
-The tag format used by an image is `microsoft/dotnet-buildtools-prereqs:<linux-distribution-name>-<version>-<variant>-<dockerfile-commit-sha>-<date-time>`
+The tag format used by an image is `<repository>:<linux-distribution-name>-<version>-<variant>-<dockerfile-commit-sha>-<date-time>`
 
+- `<repository>` - name of the repository this image is part of, almost always `microsoft/dotnet-buildtools-prereqs`
 - `<linux-distribution-name>` - name of the Linux distribution the image is based on
 - `<version>` - version of the Linux distribution
 - `<variant>` - name describing the specialization purpose of the image.  Often special dependencies are needed for certain parts of the product.  It can be beneficial to separate these dependencies into a separate Dockerfile/image.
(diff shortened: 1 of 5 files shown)


build.ps1
```diff
@@ -2,7 +2,12 @@
 param(
     [string]$DockerfilePath = "*",
     [string]$ImageBuilderCustomArgs,
-    [switch]$CleanupDocker
+    [switch]$CleanupDocker,
+
+    # Adds "--privileged" to the "docker run" command. In unknown circumstances we encounter on the
+    # RHEL build agent, the container must be run with --privileged in order to access the shared
+    # docker socket. See https://stackoverflow.com/a/36614457 for an untried potential alternative.
+    [switch]$RunPrivileged
```
MichaelSimons: Question:  Do you think there is more value in making this generic/flexible - e.g. [string]$RunCustomArgs?

src/rhel/7.5/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+        llvm-toolset-7-lldb-devel.x86_64 \
+        llvm-toolset-7-llvm.x86_64 \
+        llvm-toolset-7-llvm-libs.x86_64 \
+        wget \
+        which \
+        make \
+    && yum clean all
+
+# Needed by coreclr build to discover llvm-toolset-7 clang, and includes.
+ENV PATH "/opt/rh/llvm-toolset-7/root/usr/bin:${PATH}"
```
MichaelSimons: Consider: declaring both ENV with one instruction/layer.  Be careful of the syntax if you do, it is different - see https://github.com/dotnet/dotnet-docker/blob/master/2.1/sdk/stretch/amd64/Dockerfile#L29

dagood: I've only made some small formatting changes to this file so far since dleeapho had success with it, but I'll see about this. My concern is `${PATH}`: reading the Docker manual (very) pessimistically makes me think it won't work with `FOO=BAR` syntax:

> https://docs.docker.com/engine/reference/builder/#env
> 
> ...The first form, ENV <key> <value>, will set a single variable to a value. The entire string after the first space will be treated as the <value> - including whitespace characters. **The value will be interpreted for other environment variables, so quote characters will be removed if they are not escaped.**
> 
> **The second form**, ENV <key>=<value> ..., allows for multiple variables to be set at one time. Notice that the second form uses the equals sign (=) in the syntax, while the first form does not. Like command line parsing, quotes and backslashes can be used to include spaces within values.

But it's easy to see if it worked and I have a local build set up, so I'll just try it. It seems like it would be very silly for it not to work.

dagood:
Yep, I'll update docs.

README.md
```diff
@@ -4,15 +4,19 @@ The Dockerfiles in this repository are used for building the .NET Core product.
...
 
-The images produced from the Dockerfiles are published to the [microsoft/dotnet-buildtools-prereqs](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/) Docker Hub repository.
+Each image produced from the Dockerfiles is published based on its repository in the [manifest](#manifest):
 
-- [Most recent tags](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/tags/)
-- [Full list of tags](https://registry.hub.docker.com/v1/repositories/microsoft/dotnet-buildtools-prereqs/tags)
+- [microsoft/dotnet-buildtools-prereqs](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/) is a Docker Hub repository.
+  - [Most recent tags](https://hub.docker.com/r/microsoft/dotnet-buildtools-prereqs/tags/)
+  - [Full list of tags](https://registry.hub.docker.com/v1/repositories/microsoft/dotnet-buildtools-prereqs/tags)
+- **dotnet-buildtools-prereqs** is a repository in the *DotnetDocker* private Azure Container Registry.
```
dagood: @MichaelSimons Do you know if there's a good way to link to the resource? (Without confusing people with 404s, etc.?)

MichaelSimons: I'm not aware of one.

dagood:
(I almost forgot:)

The following image was built at a result of merging this PR:
dotnetdocker.azurecr.io/dotnet-buildtools-prereqs:rhel-7.5-b4560f6-20180725153037

PR #46, opened 2018-04-05
Upgrade npm (and node) to get new registry cert

dagood:
The new version seems to have node on PATH, so nodejs-legacy might not be needed anymore.

`--unsafe-perm=true` prevents this:

```
gyp WARN EACCES attempting to reinstall using temporary dev dir "/usr/lib/node_modules/azure-cli/node_modules/fibers/.node-gyp"
gyp WARN EACCES user "nobody" does not have permission to access the dev dir "/usr/lib/node_modules/azure-cli/node_modules/fibers/.node-gyp/8.11.1"
```

The `-f` in `npm cache clean -f` makes the clean happen. `npm cache clean` [isn't needed anymore for "normal" scenarios](https://docs.npmjs.com/cli/cache#details), so it was changed to just error out. It's still valid to run the command to free up disk space, though, which is what we're doing.

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/43

PR #64, opened 2018-11-02
Add Fedora 29 Dockerfile with procps for pkill/ps

dagood:
Remove azure-cli and npm installs: this fails, and I can't find them being used.

Remove NSS upgrade: NSS isn't installed/used by default in 29.

---

Making this new image to enable source-build smoke-test during CI: https://github.com/dotnet/source-build/issues/814.


dagood:
OpenSUSE CI consistently failing, trying to bring in fixes from https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/46. Didn't bring in the node upgrade yet, since not sure if that's needed in this case or how it'll work with `zypper`.

PR #65, opened 2018-11-05
Add missing logic from build yml to push images

MichaelSimons:
The current build definition was not pushing images to Docker Hub.  Updated it to include the necessary logic to enable this.

### DIFF
--- a/.vsts-pipelines/dotnet-buildtools-prereqs.yml
+++ b/.vsts-pipelines/dotnet-buildtools-prereqs.yml
@@ -6,6 +6,12 @@ jobs:
     demands: agent.os -equals linux
   workspace:
     clean: all
+  variables:
+    imageBuilderCustomArgs: >
+      --push
+      --username $(dockerRegistry.userName)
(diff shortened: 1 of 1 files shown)


.vsts-pipelines/dotnet-buildtools-prereqs.yml
```diff
@@ -6,12 +6,18 @@ jobs:
     demands: agent.os -equals linux
   workspace:
     clean: all
+  variables:
+    imageBuilderCustomArgs: >
+      --push
+      --username $(dockerRegistry.userName)
+      --password $(BotAccount-dotnet-dockerhub-bot-password)
+      $(imageBuilder.queueBuildArgs)
```
dseefeld: Why did this change from buildArgs to queueBuildArgs?

MichaelSimons: I renamed this to make it more clear that these are build args defined when the build was queued.

PR #73, opened 2019-02-07
Add linux musl arm64 build image.

jashook:
This adds an ubuntu x64 image that can cross build linux arm64 musl. Note that it has to build ld.gold which takes quite some time. This is required to link correctly when building arm64 with the crossrootfs.

### DIFF
--- a/src/ubuntu/16.04/cross/arm64-alpine/Dockerfile
+++ b/src/ubuntu/16.04/cross/arm64-alpine/Dockerfile
@@ -10,13 +10,14 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Workaround to build a functioning ld.gold which can link for linux-musl arm64
-RUN cd /tmp && git clone https://github.com/richfelker/musl-cross-make.git
-
-RUN echo "TARGET = aarch64-alpine-linux-musl" >> /tmp/musl-cross-make/config.mak
-RUN echo "OUTPUT = /usr" >> /tmp/musl-cross-make/config.mak
-RUN echo "BINUTILS_CONFIG=--enable-gold=yes --program-prefix=aarch64-alpine-linux-musl-" >> /tmp/musl-cross-make/config.mak
-
-RUN cd /tmp/musl-cross-make && make
-RUN cd /tmp/musl-cross-make && make install
+RUN cd /tmp \
+    && wget https://ftp.gnu.org/gnu/binutils/binutils-2.27.tar.gz \
+    && tar -xf binutils-2.27.tar.gz \
+    && cd binutils-2.27 \
+    && ./configure --disable-werror --target=aarch64-alpine-linux-musl --prefix=/usr --libdir=/lib --disable-multilib --with-sysroot=aarch64-alpine-linux-musl --enable-gold=yes --program-prefix=aarch64-alpine-linux-musl- \
+    && make \
+    && make install \
+    && cd .. \
+    && rm -r *
 
 ADD rootfs.arm64.tar crossrootfs
--- a/src/ubuntu/16.04/cross/arm64-alpine/Dockerfile
+++ b/src/ubuntu/16.04/cross/arm64-alpine/Dockerfile
@@ -1,0 +1,22 @@
+FROM microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-crossdeps
+
+# Install binutils-aarch64-linux-gnu
+#
+# Also install bison. This is required to build musl-cross-make
+RUN apt-get update \
+    && apt-get install -y \
+        binutils-aarch64-linux-gnu \
+        bison \
+    && rm -rf /var/lib/apt/lists/*
+
+# Workaround to build a functioning ld.gold which can link for linux-musl arm64
+RUN cd /tmp && git clone https://github.com/richfelker/musl-cross-make.git
+
+RUN echo "TARGET = aarch64-alpine-linux-musl" >> /tmp/musl-cross-make/config.mak
+RUN echo "OUTPUT = /usr" >> /tmp/musl-cross-make/config.mak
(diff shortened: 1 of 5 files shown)


jashook:
@janvorli @MichaelSimons ptal

/cc @wfurt @safern @RussKeldorph

janvorli:
@jashook I have experimented a bit and found that instead of building the whole musl-cross-make from which we need just the ld.gold, we can actually build just the binutils (from https://ftp.gnu.org/gnu/binutils/binutils-2.27.tar.gz), which is much smaller. The build of that is 4 times faster than the build of the musl-cross-make and also produces less unnecessary stuff.
So if you want,  you can replace the code that builds the musl-cross-make by downloading the binutils, unpacking it, changing current directory into the unpacked one and then running
```
./configure --disable-werror --target=aarch64-alpine-linux-musl --prefix=/usr --libdir=/lib --disable-multilib --with-sysroot=aarch64-alpine-linux-musl --enable-gold=yes --program-prefix=aarch64-alpine-linux-musl-
make
sudo make install
```
I was trying to test the full docker image build locally, but my docker is having issues find the base image:
```
Error response from daemon: manifest for microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-crossdeps not found
```
But I have added a suggestion on what I believe should work.

src/ubuntu/16.04/cross/arm64-alpine/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+#
+# Also install bison. This is required to build musl-cross-make
+RUN apt-get update \
+    && apt-get install -y \
+        binutils-aarch64-linux-gnu \
+        bison \
+    && rm -rf /var/lib/apt/lists/*
+
+# Workaround to build a functioning ld.gold which can link for linux-musl arm64
+RUN cd /tmp && git clone https://github.com/richfelker/musl-cross-make.git
```
janvorli: Hmm, multiline suggestions don't work. 
Here is what you can use instead. Please note that it is preferred practice to use a single RUN command instead of multiple ones as each RUN creates a layer in the resulting image.
```
# Workaround to build a functioning ld.gold which can link for linux-musl arm64
RUN cd /tmp \
    && wget https://ftp.gnu.org/gnu/binutils/binutils-2.27.tar.gz \
    && tar -xf binutils-2.27.tar.gz \
    && cd binutils-2.27 \
    && ./configure --disable-werror --target=aarch64-alpine-linux-musl --prefix=/usr --libdir=/lib --disable-multilib --with-sysroot=aarch64-alpine-linux-musl --enable-gold=yes --program-prefix=aarch64-alpine-linux-musl- \
    && make \
    && make install \
    && cd .. \
    && rm -r *
```

jashook: HI Jan, I had to rebuild several times and there were issues during the build. I kept the run commands separate to allow for restore points; however, thinking further I do not believe many people, if any build the images out of this repo manually.

janvorli: That's the way I also usually do stuff in Dockerfile. I just convert it to the minimal number of RUN commands when it is ready for PR. @MichaelSimons advised me on keeping a minimal muber of RUN commands a long time ago.

jashook: Makes sense thanks!

jashook:
As it is now this change times out the ubuntu_16 build. So hopefully with the change the build time is better. Testing locally now. Also @janvorli you can build the crossdep Dockerfile locally so that you can pull it to use for the cross build.

I have been using build.ps1 to build.

```sh
# Sudo is required to run the --privileged docker container.
sudo pwsh
./build.ps1 -DockerfilePath "src/ubuntu/16.04/crossdeps"
./build.ps1 -DockerfilePath "src/ubuntu/16.04/cross/arm64-alpine"
```

jashook:
Building locally now. Will update when/if the new image successfully builds.

janvorli:
> you can build the crossdep Dockerfile locally

Ah, good point, I haven't realized that.

jashook:
The image build is **significantly** faster with the change. Thank you Jan.

jashook:
Verified a clean build.

```sh
./build.sh arm64 checked cross -clang5.0
```

jashook:
@dotnet-bot test alpine

jashook:
Opensuse is taking a while but it has been untouched. @MichaelSimons ptal

src/ubuntu/16.04/cross/arm64-alpine/Dockerfile
MichaelSimons: As Jan pointed out it is general best practice to reduce the number of layers.  Because of the way docker utilizes a union file system and layers, you must do this in scenarios where you want to cleanup temporary artifacts.  If not, you end up caring the deleted files in a base layer anyways which defeats the purpose.  Docker does provide the ability to consolidate all image layers into one which is useful in some scenarios.  I agree though that often times during development iterations, it is beneficial to break the layers apart.

MichaelSimons:

microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-10fcdcf-20190207192950
microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-debpkg-d770b8b-20190207192950
microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-coredeps-d770b8b-20190207192950
microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-crossdeps-10fcdcf-20190207192950
microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-cross-arm64-e435274-20190207192950
microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-cross-arm64-alpine10fcdcf-20190207192950
microsoft/dotnet-buildtools-prereqs:ubuntu-16.04-cross-10fcdcf-20190207192950

PR #76, opened 2019-02-08
Adding new docker image for debian arm32 with preinstalled libgpiod

joperezr:
cc: @MichaelSimons 

This will add a debian arm32 image that will come with libgpiod library preinstalled. This will be required by the dotnet/iot repo which will have to build a shim for this library, so we need to be able to get the built library out in order to be able to cross compile the shim for arm distros.

FYI: @buyaa-n @joshfree


src/debian/iot/arm32v7/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM arm32v7/debian:latest
+
+# Dependencies for downloading and running the installation script of libgpiod
+RUN apt-get update \
+    && apt-get install -y \
+            wget \
```
MichaelSimons: Docker best practices suggest alphabetizing apt-get package lists.  It helps the readability.  This is definitely the case for some of our amd64 Dockerfiles where the number of packages is very large.

MichaelSimons:

microsoft/dotnet-buildtools-prereqs:debian-iot-arm32v7-be5b37d-20190211200049

PR #81, opened 2019-02-19
Adding libgpiod dependency to ubuntu 16.04 cross

joperezr:
cc: @MichaelSimons @safern 

Initially what we wanted for the dotnet/iot repo was to use two different containers for their build. One would be the iot container (with arm architecture) which had the libgpiod dependency, and then the other would be the cross container where we would build the repo after copying libgpiod from the first container. 

This plan won't work for now as Azure pipelines (where dotnet/iot builds) won't support arm containers yet, so in the meantime, we need to have a cross container which already has libgpiod installed. I will contact Azure devops to get that support, and once it works, I'll come back and undo this change and use the more reproducible process instead.


MichaelSimons:
@[joperezr](/joperezr) - Can you explain how you were trying to copy the lib out of the ARM container in our AMD64 build environment?  I would expect the following to work.

```
docker create --name iot microsoft/dotnet-buildtools-prereqs:debian-iot-arm32v7-be5b37d-20190211200049
docker cp iot:/usr/local/lib/libgpiod.so .
docker container rm -f iot
```

Regarding this PR.  I am a little hesitant about this approach because of the checked in tarball even thought the desire is for a quick temporary solution.

joperezr:
> Can you explain how you were trying to copy the lib out of the ARM container in our AMD64 build environment?

I tried two different things, first I tried having a build phase that all it would do was to use Azure DevOps built-in container support to run in this arm container, and would only run one step which would simply be to add libgpiod assets to the build artifacts. That didn't work because DevOps will fail to initialize the container given that the target architecture is ARM. The error I was getting was:
```
##[error]Docker exec fail with exit code 1
##[debug]System.InvalidOperationException: Docker exec fail with exit code 1
   at Microsoft.VisualStudio.Services.Agent.Worker.ContainerOperationProvider.StartContainerAsync(IExecutionContext executionContext, ContainerInfo container)
   at Microsoft.VisualStudio.Services.Agent.Worker.ContainerOperationProvider.StartContainersAsync(IExecutionContext executionContext, Object data)
   at Microsoft.VisualStudio.Services.Agent.Worker.JobExtensionRunner.RunAsync()
   at Microsoft.VisualStudio.Services.Agent.Worker.StepsRunner.RunStepAsync(IStep step, CancellationToken jobCancellationToken)
```

Next thing, I tried just having a script that would do something very similar to what you outlined above, where instead of using the built in support for docker, I would just call docker myself to create the container and grabbed the assets out, but I get another failure because looks like Azure DevOps requires a very specific version of node to be installed on the image in order for it to be able to create the container, which is why I went with this as my third option.

> I am a little hesitant about this approach because of the checked in tarball even thought the desire is for a quick temporary solution.

I also don't love this approach, although the tarball was generated by using the container that is also checked in to this repo, so I guess another option instead of checking it in would be to do something similar to what you do with the cross arm and x86 tarballs where you generate them at build time out of the other container. What would you think about that?

joperezr:
Closing this as I will go with @MichaelSimons idea of copying assets without starting container that we discussed offline.

Issue #82, opened 2019-02-20
Host images on mcr.microsoft.com


MichaelSimons:
All Microsoft images are to be hosted on MCR.  The image repo should be

`mcr.microsoft.com/dotnet-buildtools/prereqs`

Since these images are engineering infrastructure for the dotnet open source project, they should be hosted in a community repo on Docker Hub versus a verified publisher repo.

MichaelSimons:
All images on microsoft/dotnet-buildtools-prereqs should be copied to mcr.  This will allow existing usages to all be moved to utilize mcr.

PR #83, opened 2019-02-20
Update build definition to push images to MCR

MichaelSimons:
Related to #82


MichaelSimons:
FYI - I updated the readme to reflect the images will now be available on MCR.

PR #89, opened 2019-02-27
make sure /root/commands is accessible to other users

wfurt:
Needed to start helix client. 

Can you please rebuild and post another image @MichaelSimons ? Build is failing now as helixbot cannot execute common scripts.


MichaelSimons:
The following image was built as a result of merging this PR.

mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.9-helix-ce7c948-20190227232728

PR #96, opened 2019-03-11
AMR32 & ARM64 Dockerfiles for Debian 9, Ubuntu 16.04 and Ubuntu 18.04

ulisesh:


ulisesh:
cc @jashook @wfurt @safern

manifest.json
```diff
@@ -375,6 +375,32 @@
             }
           ]
         },
+        {
+          "platforms": [
+            {
+              "architecture": "arm",
+              "dockerfile": "src/ubuntu/16.04/helix/arm32v7",
```
jashook: Following the alpine convention the path would be:

src/ubuntu/16.04/arm32v7/helix/

src/debian/9/helix/arm32v7/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:debian-9-arm32v7
```
jashook: Seems like this from should just be the local path instead of the mcr location.

manifest.json
ulisesh: I followed alpine 3.6 convention src/alpine/3.6/ForHelix/amd64

@MichaelSimons, can you comment on what is the expected convention?

jashook: On closer look this looks correct thanks!

wfurt:
we should not run tests as root. I would suggest to add something like #88 
Some corefx tests will fail without it.

src/debian/9/helix/arm32v7/Dockerfile
ulisesh: > 
> 
> In general why are we pulling in the build packages? I think eventually we may end up doing native builds natively on these machines, but hopefully not on helix. These dependencies will inflate our image sizes.

I wrongly assumed the images you created had product dependencies and not build dependencies. I will changes Helix images to use base images from docker hub, thanks for pointing this out

ulisesh:
> we should not run tests as root. I would suggest to add something like #88
> Some corefx tests will fail without it.

Addressed this in the latest commit

manifest.json
```diff
@@ -510,7 +603,10 @@
               "dockerfile": "src/ubuntu/18.04/arm64v8",
               "os": "linux",
               "tags": {
-                "ubuntu-18.04-arm64v8-$(System:DockerfileGitCommitSha)-$(System:TimeStamp)": {}
+                "ubuntu-18.04-arm64v8-$(System:DockerfileGitCommitSha)-$(System:TimeStamp)": {},
+                "ubuntu-18.04-arm64v8": {
```
MichaelSimons: Please revert these unnecessary manifest changes.  Since you aren't adding a dependency on these images the local tag is not needed.

ulisesh: Reverted

src/ubuntu/18.04/helix/arm64v8/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+        build-essential \
+        cmake \
+        clang \
+        gdb \
+        lldb-3.9 \
+        git \
+        gcc \
+        python-pip \
+    && rm -rf /var/lib/apt/lists/* \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
```
MichaelSimons: From a readability perspective I think it would be good to add a blank line before this (e.g. `    \`) so that it doesn't get overlooked with the package installation/cleanup.

src/debian/9/helix/arm64v8/Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM arm64v8/debian:9
+
+# Install Helix Dependencies
+
+RUN apt-get update && \
+    apt-get install -y \
+        locales \
```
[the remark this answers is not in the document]
ulisesh: Done, will take in consideration next time

ulisesh:
@MichaelSimons is openSUSE failure expected?

MichaelSimons:
@ulisesh - yes, see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/97

ulisesh:
> @ulisesh - yes, see #97

Thanks, looks like all other checks passed

MichaelSimons:
The following images were produced as a result of merging this PR:

```
mcr.microsoft.com/dotnet-buildtools/prereqs:debian-9-helix-arm64v8-0a0ebdd-20190312215438
mcr.microsoft.com/dotnet-buildtools/prereqs:debian-9-helix-arm32v7-0a0ebdd-20190312215452
mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-helix-arm64v8-0a0ebdd-20190312220351
mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-helix-arm32v7-0a0ebdd-20190312215532
mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-helix-arm64v8-0a0ebdd-20190312221438
mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-helix-arm32v7-0a0ebdd-20190312215551
```

PR #113, opened 2019-04-10
Refactor pipeline to support PR builds

mthalman:
Refactors the pipeline logic to support PR builds.
This includes changes for both PR and official pipelines that splits them apart such that they each are responsible for building a specific Linux distro.  For example, the dotnet-buildtools-prereqs-pr-fedora.yml pipeline will be triggered when a change is made to src/fedora folder and it will only build the images in that folder.


mthalman:
> core-setup is utilizing [this check](https://github.com/dotnet/core-setup/blob/master/eng/jobs/bash-build.yml#L118) that I think should be considered here as it seem like it would be a way to eliminate the need for separate build yamls for official and ci build.
> 
> ```
> ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
> ```
> Thoughts?

Yes, that definitely cleans things up.  I've made changes to reflect this.  The condition I use only checks whether it's public.  I figured that we'd still want to run everything for internal PRs.

.vsts-pipelines/dotnet-buildtools-prereqs-alpine.yml
```diff
@@ -0,0 +1,23 @@
...
+    - src/alpine/*
+
+variables:
+- template: ../variables/common.yml
+
+jobs:
+  - template: jobs/build-image-legs.yml
+    parameters:
+      imageBuilderPath: src/alpine/*
+      isPullRequestBuild: $(isPullRequestBuild)
```
MichaelSimons: Missing line ending.

.vsts-pipelines/variables/common.yml
```diff
@@ -0,0 +1,3 @@
+variables:
+- ${{ if ne(variables['System.TeamProject'], 'public') }}:
+  - template: common-official.yml
```
MichaelSimons: I feel like separating the common-official variables isn't worthwhile.  The main detractor is that they contain groups which means you aren't going to be able to use the short hand syntax here if variables are ever added.  Having the common-official variables here would make that more obvious.

.vsts-pipelines/dotnet-buildtools-prereqs.yml
```diff
@@ -2,35 +2,9 @@ trigger: none
 pr: none
```
MichaelSimons: What is the expected use case for this build pipeline going forward?

mthalman: After thinking about it, it may not really be necessary.  The only benefit it would have is if we needed to manually run a build and have it target a more specific path than one of the other pipelines.  I could go either way on this.

MichaelSimons: My vote would be to proactively clean it up.  If it turns out that there is a use case for it, it would be simple to add back.

PR #117, opened 2019-04-12
Define a single uber PR build pipeline

mthalman:
It was determined that we can't make use of the path filter capability of AzDO pipelines when targeting GitHub repos.  This means for PR verification we can't make use of the OS-specific pipeline files that made use of the path filter.  Instead, we'll need to have a single pipeline that covers all the OS types.

These changes define a new PR yml pipeline and makes use of a matrix that defines the src paths to all the OS types.  This is basically a poor man's version of the functionality that Image Builder provides with it's matrix generation.  This was just a quick implementation to get things up and running before migrating to use the matrix generation.


mthalman:
> Can you log an issue to track updating the PR build to use the generate matrix functionality?

#118

PR #119, opened 2019-04-12
Delete netci.groovy

mthalman:
No longer needed now that we've migrated to AzDO builds.  Repo has already been updated to remove the CI webhooks.


mthalman:
> LGTM
> 
> I believe this needs to be deleted as well - https://github.com/dotnet/dotnet-ci/blob/master/data/repolist.txt#L58
> 
> cc @chcosta - FYI

Yep, I've submitted a PR to make that change already.

PR #128, opened 2019-05-01
Fix ubuntu trigger paths to not use wildcard for version folder name

mthalman:
The build triggers aren't working for Ubuntu, apparently because a wildcard is used in the folder name.  Changing it to be explicit for the version name instead.  This is technically the right thing to do anyway in case we were to ever add support for additional versions with the same major version, in which case we'd want separate pipelines for each.

### DIFF
--- a/.vsts-pipelines/dotnet-buildtools-prereqs-ubuntu14.yml
+++ b/.vsts-pipelines/dotnet-buildtools-prereqs-ubuntu14.yml
@@ -4,14 +4,14 @@ trigger:
     - master
   paths:
     include:
-    - src/ubuntu/14*
+    - src/ubuntu/14.04/*
 pr:
   branches:
     include:
     - master
   paths:
     include:
-    - src/ubuntu/14*
+    - src/ubuntu/14.04/*
 
 variables:
 - template: variables/common.yml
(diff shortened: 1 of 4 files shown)


.vsts-pipelines/dotnet-buildtools-prereqs-ubuntu14.yml
```diff
@@ -4,14 +4,14 @@ trigger:
     - master
   paths:
     include:
-    - src/ubuntu/14*
+    - src/ubuntu/14.04/*
 pr:
```
MichaelSimons: This comment is orthogonal to these changes.  Should these pr triggers be removed?  They are a little mis-leading.

mthalman: I could go either way.  They still can apply if we were to ever do a pull request on the internal repo.

MichaelSimons: My preference is to clean up dead/unused code paths as they are a maintenance tax and add complexity to the code base.  In this particular case, it adds confusion as to why this is needed in addition to https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/.vsts-pipelines/dotnet-buildtools-prereqs-pr.yml.  If needed in the future it would be easy to add back.

PR #129, opened 2019-05-01
Add ARM legs to PR validation

MichaelSimons:


wfurt:
are we sure we have enough capacity on ARM systems?

MichaelSimons:
There are currently three ARM leg failures but those are existing issues orthogonal to these changes.

MichaelSimons:
@wfurt - the reason for the failures on the earlier run is related to a config issue with the build agents that has been addressed by DDFUN.

PR #131, opened 2019-05-01
create helixbot user in consistent way

wfurt:
This is replacement for  #130
To make behavior more consistent (and easy to clone) we
1. use adduser consistently
2. use same groups on all OSes
3. use explicit options to avoid dependency on implicit behavior

For each OS group I verified that helixbot is created,  has home directory and and can sudo without password.

BTW it seems like my repo is polluted. I only realized it when created PR. squash merge should take care of it as the diff is what I wanted.


wfurt:
I'll watch it. For now, I plan to update only Ubuntu  19.
I want to do few more modifications to add packages so Process tests do run instead of skipping. I'll do more testing than.

src/fedora/28/helix/amd64/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN python -m pip install --upgrade pip==19.0.3 && \
 ENV LANG=en-US.UTF-8
 
 # create helixbot user and give rights to sudo without password
-RUN /usr/sbin/useradd -G adm -s /bin/sh helixbot; \
+RUN /usr/sbin/adduser --disabled-password --gecos '' --uid 1001 --shell /bin/bash --ingroup adm helixbot
```
MichaelSimons: FYI - the Fedora CI legs are failing.  
```suggestion
RUN /usr/sbin/adduser --disabled-password --gecos '' --uid 1001 --shell /bin/bash --ingroup adm helixbot; \
```

wfurt: fixed. The docker file explicitly refers to mcr but it is failing to find dependent image. 

```
docker build .
Step 1/6 : FROM mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-30
manifest for mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-30 not found
```

MichaelSimons: How are you building?  You need to have the prereqs built (e.g. mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-30)  - the build script handles this for you.
  
```
\build.ps1 -DockerfilePath "fedora/30*"
```

wfurt:
I did another update and decided to use '&&' instead of ';'
That allows to propagate failures from  adduser
Its seems like verification is stuck because of Azure outage.

PR #74, opened 2019-02-07
Add missing NUMA development library

janvorli:
This library is needed to enable the coreclr binaries to be NUMA aware.
We have found today that this was accidentally missing for all of our build docker images.

PR #133, opened 2019-05-01
Revert changes to add libnuma-dev to select arm32 images

MichaelSimons:
These changes were recently introduced in #74.  libnuma-dev is not available in the [ubuntu 16.04](https://packages.ubuntu.com/xenial/libnuma-dev) and [debian 9](https://packages.debian.org/stretch/libnuma-dev) arm32 default package feeds.  I would like to revert these changes to unblock the builds.

### DIFF
--- a/src/debian/9/arm32v7/Dockerfile
+++ b/src/debian/9/arm32v7/Dockerfile
@@ -10,7 +10,6 @@ RUN apt-get update \
             libicu-dev \
             libkrb5-dev \
             liblttng-ust-dev \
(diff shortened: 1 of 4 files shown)


MichaelSimons:
@janvorli - not a problem - we should have had pr validation running on arm to catch these types of things.  I have since [enabled the arm runs](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/129) on this repo.  This is something that was recently made possible by DDFUN.

PR #143, opened 2019-05-18
Fix Alpine 3.6 docker image

janvorli:
Due to some shuffles / image changes in the Alpine repo, the image got broken
again. This change fixes that.

### DIFF
--- a/src/alpine/3.6/amd64/Dockerfile
+++ b/src/alpine/3.6/amd64/Dockerfile
@@ -20,6 +20,7 @@ RUN apk add --no-cache \
         libtool \
         libunwind-dev \
         linux-headers \
+        llvm \
         make \
         openssl \
         openssl-dev \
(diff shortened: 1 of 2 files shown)


MichaelSimons:
This is also affecting arm64.

```
ERROR: unsatisfiable constraints:‌
  llvm5 (missing):‌
    required by: world[llvm5]
```

Any thoughts on upgrading to a supported version of Alpine?

janvorli:
> Any thoughts on upgrading to a supported version of Alpine?

That would mean reducing Linux distros we can run on. The version we build is not just for Alpine, but for any distro based on MUSL.
Moreover, it would not help. The lldb is in the edge/testing branch for all versions of Alpine, so the same issue would arise even if we were building on Alpine 3.9.

Let me fix the arm64 too.

MichaelSimons:
My comment on upgrading the version was somewhat orthogonal to this particular fix.  My concern stems from 3.6 reaching EOL on 2019-05-01.

janvorli:
Ah, I didn't know about that. We should look into it then. We might still be blocked by the issue with inability to build on newer Alpine Linux versions that we've hit in the past.

janvorli:
@MichaelSimons can you please share the image names containing this change?

MichaelSimons:

```
mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.6-05bedd1-20190521001804
mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.6-WithNode-cfdd435-20190521001804
mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.6-helix-09ca40b-20190521001758
mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.6-arm64v8-bcdc5e3-20190521001712
```

janvorli:
@MichaelSimons I've tried to build Alpine on 3.9 and it works again. The issue that we've hit on 3.8 got fixed. I just needed to do a little change to the version of the dotnet cli that is installed at the beginning (the 2.1.504 didn't support openssl1.1 - there is no openssl1.0 on 3.9 -  while the 2.1.604 does).
In addition to that, I've found that binaries built on 3.9 still work even on 3.6 (with a small change to the ICU shim to start scanning available ICU libraries from an earlier version). This is thanks to the fact that unlike glibc, MUSL based distros doesn't use versioned symbols.

PR #144, opened 2019-05-20
Install Ubuntu build pre-reqs in pipeline

mthalman:
The build machines require some additional packages installed on them in order to resolve some Docker issues for Ubuntu.

### DIFF
--- a/.vsts-pipelines/jobs/build-images.yml
+++ b/.vsts-pipelines/jobs/build-images.yml
@@ -22,12 +22,15 @@ jobs:
         --username $(acr.userName)
         --password $(BotAccount-dotnet-docker-acr-bot-password)
         $(imageBuilder.queueArgs)
+      dockerRunOptions: -e DOCKER_REPO=$(acr.server)/$(publishRepoPrefix)dotnet-buildtools/prereqs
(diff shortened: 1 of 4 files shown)


.vsts-pipelines/jobs/build-images.yml
```diff
@@ -22,12 +22,15 @@ jobs:
         --username $(acr.userName)
         --password $(BotAccount-dotnet-docker-acr-bot-password)
         $(imageBuilder.queueArgs)
+      dockerRunOptions: -e DOCKER_REPO=$(acr.server)/$(publishRepoPrefix)dotnet-buildtools/prereqs
```
MichaelSimons: I think it would be good to add a comment that calls out why this is needed.  It may help prevent accidental removal in the future.

.vsts-pipelines/steps/install-build-prereqs.yml
```diff
@@ -0,0 +1,4 @@
+steps:
```
MichaelSimons: What was the reasoning on separating this into a template versus in-lining it?

mthalman: Leftover from my experimentation of trying to get a better conditional execution of the task.  Added it back inline.

PR #151, opened 2019-06-13
Add arm32v7 alpine 3.9 Dockerfile

krwq:
cc: @janvorli 

except for numactl-dev missing (replaced with building from sources), it is a copy of arm64 dockerfile

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -673,19 +673,6 @@
         {
           "platforms": [
             {
-              "architecture": "arm64",
-              "dockerfile": "src/alpine/3.9/arm64v8",
-              "os": "linux",
-              "tags": {
-                "alpine-3.9-arm64v8-$(System:DockerfileGitCommitSha)-$(System:TimeStamp)": {}
-              },
-              "variant": "v8"
-            }
-          ]
-        },
-        {
-          "platforms": [
-            {
               "architecture": "arm",
               "dockerfile": "src/alpine/3.9/arm32v7",
               "os": "linux",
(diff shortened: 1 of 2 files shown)


mthalman:
This new Dockerfile needs to be referenced in the manifest.json file at the root of the repo in order for it to get built.  Add this snippet next to the existing src/alpine/3.9/arm64v8 reference:
```
        {
          "platforms": [
            {
              "architecture": "arm",
              "dockerfile": "src/alpine/3.9/arm32v7",
              "os": "linux",
              "tags": {
                "alpine-3.9-arm32v7-$(System:DockerfileGitCommitSha)-$(System:TimeStamp)": {}
              },
              "variant": "v7"
            }
          ]
        },
```

janvorli:
@krwq the numactl-dev is not needed for arm32, you can just remove it. We do the same for linux-arm. Enabling NUMA on 32 bit systems doesn't make sense.

krwq:
@janvorli, thanks, I didn't know what NUMA is

manifest.json
```diff
@@ -683,6 +683,19 @@
...
+              "architecture": "arm",
+              "dockerfile": "src/alpine/3.9/arm32v7",
+              "os": "linux",
+              "tags": {
+                "alpine-3.9-arm32v7-$(System:DockerfileGitCommitSha)-$(System:TimeStamp)": {}
+              },
+              "variant": "v7"
+            }
+          ]
+        },
```
mthalman: Nit: Could you actually move this up above the 3.9 ARM64 entry?  That'll keep things consistent where ARM32 is listed before ARM64 for everything.

mthalman:
Failed build looks to be a random failure unrelated to the change.  I'll go ahead and merge the changes.

mthalman:
Images produced by this change:

mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.9-arm32v7-2612d91-20190613183347

PR #156, opened 2019-06-24
Integrate eng common folder

mthalman:
Related to https://github.com/dotnet/docker-tools/issues/160


eng/pipelines/dotnet-buildtools-prereqs-ubuntu16.yml
```diff
@@ -10,7 +10,9 @@ pr: none
...
 - template: variables/common.yml
 - name: imageBuilder.pathArgs
-  value: src/ubuntu/16.04*
+  value: --path 'src/ubuntu/16.04*'
 
 stages:
-  - template: stages/build-repo.yml
+  - template: ../common/templates/stages/build-test-publish-repo.yml
+    parameters:
+      linuxAmdBuildJobTimeout: 120
```
MichaelSimons: I am surprised more legs didn't require a longer timeout. (e.g. the ubuntu 14.04 leg)

manifest.json
```diff
@@ -246,6 +266,7 @@
             {
               "dockerfile": "src/ubuntu/14.04",
               "os": "linux",
+              "osVersion": "trusty",
```
MichaelSimons: The difference in versioning scheme here is unfortunate.  The Dockerfile paths and FROM statements are using the version number syntax while the osVersion is using the version code name.  Is it possible to unify these?  Can the osVersion be ubuntu14.04?

mthalman: Sure.  Also updated xenial to ubuntu16.04.

eng/pipelines/dotnet-buildtools-prereqs-ubuntu16.yml
mthalman: Ubuntu 14.04 actually runs pretty quick: ~30 mins.  I did however determine that CentOS and openSUSE needed to have their timeouts extended.

Issue #69, opened 2018-12-18
How do I see the latest built and published image for a given Dockerfile?


dagood:
This is the best way I can think of right now to find the latest version of a specific image I want to use:

1. Find the dockerfile I'm interested in on GitHub.
2. Open the full list of tags (https://registry.hub.docker.com/v1/repositories/microsoft/dotnet-buildtools-prereqs/tags).
3. Search for the initial part of the name (like ubuntu-18.04) using the convention in the readme.
4. Pick the one with the latest date. Unfortunately the tags are sorted by name, and commit hash is in the name before the build date. Besides making a tool, the most efficient way I can think to do find latest is put the images in a text editor, edit out the commit, sort again, and see which image corresponds to the latest.

Is there any status on something like publishing this info to dotnet/versions or the Arcade BAR?

/cc @MichaelSimons @dleeapho

tmat:
For devs who do not know about how we build these docker images it would be useful to have a starting page that lists [all distros and platforms Core CLR supports](https://docs.microsoft.com/en-us/dotnet/core/linux-prerequisites?tabs=netcore2x#supported-linux-versions) and where to find the latest container that we use in CI to test them.

MichaelSimons:
I agree this would be valuable.  It is pretty low in the priority stack rank at the moment however.

MichaelSimons:
Plan for implementation.  The build will generate the MCR tags yaml, utilize the MCR tag doc generator to create the tags details and insert this into the readme.  The build will them push the updated readme back to the repo.

mthalman:
The original plan for this was to have the latest list of image tags listed in the readme that was automatically updated after each build.  Due to implementation complexity, we're backing off that plan to something simpler for now.  We'll instead provide a link in the readme to the [image info file](https://github.com/dotnet/versions/blob/master/build-info/docker/image-info.dotnet-dotnet-buildtools-prereqs-docker-master.json) that contains metadata about the latest images that have been published, including the list of tags.  It is human readable and listed alphabetically by the path to the Dockerfile in the repo.

PR #163, opened 2019-08-01
Update pipeline to make use of all stages

mthalman:
The build pipeline for this repo had been configured to only run the build stage.  When it built the images, it would push them directly to the public location rather than staging.  Since there were no tests to execute, the test stage was skipped.  And the publish stage was skipped because it had already pushed the images directly to the public location.

But skipping the publish stage meant that we weren't publishing the image info data for these images.  And we need that data as part of the work for #69.  These changes update the pipeline so that all 3 stages get executed.  Tests are just a no-op since there are no tests for these images.

One key part of these changes is that it makes use of the `image-info` parameter of the copyAcrImages command.  This parameter was added in https://github.com/dotnet/docker-tools/pull/239 and is necessary because the image tags contain dynamic values.

### DIFF
--- a/eng/common/templates/jobs/publish.yml
+++ b/eng/common/templates/jobs/publish.yml
@@ -15,6 +15,10 @@ jobs:
       publicSourceBranch: master
   steps:
   - template: ../steps/init-docker-linux.yml
+  - template: ../steps/download-build-artifact.yml
+    parameters:
+      targetPath: $(Build.ArtifactStagingDirectory)
+      artifactName: image-info
   - script: >
       $(runImageBuilderCmd) copyAcrImages
       $(stagingRepoPrefix)
@@ -26,6 +30,7 @@ jobs:
       --os-type '*'
       --architecture '*'
       --repo-prefix $(publishRepoPrefix)
+      --image-info $(artifactsPath)/image-info.json
       $(imageBuilder.pathArgs)
       $(imageBuilder.commonCmdArgs)
     displayName: Copy Images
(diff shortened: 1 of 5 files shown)


eng/common/templates/jobs/publish.yml
```diff
@@ -26,6 +30,7 @@ jobs:
       --os-type '*'
       --architecture '*'
       --repo-prefix $(publishRepoPrefix)
+      --image-info $(artifactsPath)/image-info.json
```
MichaelSimons: Thoughts on making this optional and only used in this repo?

mthalman: Unless there's a need for it, I'd prefer to keep it consistent.  I think we should strive for no special conditions unless they're really necessary so that we can more easily determine whether a change breaks something.  I want to avoid situations where the pipeline works in one GH repo but not in another one.

MichaelSimons: I don't have a strong argument against that.  If this is the direction, then is there any value in having the CopyAcrImagesCommand support having no image-info.json?  It seems like dead code that is adding complexity albeit minor, to the codebase.

mthalman: Keeping it in there simplifies running the command locally.  So I think there's value in that.

mthalman: And it might be a good fallback to have just in case things aren't working properly when making use of the image info.  If we're trying to get a release out the door and the use of image info is causing errors for whatever reason, we can quickly change the pipeline to not pass in the image info.

PR #169, opened 2019-08-06
Change docker-testrunner base image from stretch to stretch-slim

MichaelSimons:
This reduces the size of the image by ~70 mb on disk which is ~13%.  For as many times as the image is pulled this seems worth doing.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -151,11 +151,11 @@
         {
           "platforms": [
             {
-              "dockerfile": "src/debian/stretch/docker-testrunner",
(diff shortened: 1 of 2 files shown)


mthalman:
What prevents us from using Alpine instead?

MichaelSimons:
The main reason is the channel of PS we are using is not officially supported on Alpine.

mthalman:
These changes produced the following image:
```
mcr.microsoft.com/dotnet-buildtools/prereqs:debian-stretch-slim-docker-testrunner-d61254f-20190807161111
```

PR #171, opened 2019-08-07
enabling dumps for windows docker

epananth:
Enabling dumps for nano

https://github.com/dotnet/core-eng/issues/7038

### DIFF
--- a/src/nanoserver/1803/helix/amd64/Dockerfile
+++ b/src/nanoserver/1803/helix/amd64/Dockerfile
@@ -26,5 +26,15 @@ RUN python -m pip install --upgrade pip \
                              six==1.12.0 \
                              virtualenv==16.5.0
 
+RUN pwsh -Command \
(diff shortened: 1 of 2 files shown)


epananth:
@MattGal you are correct, I reran it without the layer, it worked!

src/nanoserver/1803/helix/amd64/Dockerfile
```diff
@@ -24,7 +24,16 @@ RUN python -m pip install --upgrade pip \
                              pyopenssl==19.0.0 \
                              requests==2.21.0 \
                              six==1.12.0 \
-                             virtualenv==16.5.0
+                             virtualenv==16.5.0 \
+    && pwsh -Command \
+    New-Item -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\' -Name 'dotnet.exe' -Force -ErrorAction SilentlyContinue ; \
```
mthalman: For readability, I would indent all of these Powershell commands so it's clear they're all a continuation of the call to `pwsh -Command`.  Same comment applies to the other file.

epananth: Sure!

Issue #187, opened 2019-09-19
Ubuntu 16.04 Cross Images unable to build CoreCLR


jkoritzinsky:
The docker images for Ubuntu 16.04 for cross building against 14.04 ARM, 16.04 ARM64, and Alpine ARM64 fail to build CoreCLR due to what looks like a misconfigured/incorrectly generated rootfs.

This broke at some point between 2019-05-28 (the newest image CoreCLR is currently using) and 2019-08-14. 

Sample build here: https://dev.azure.com/dnceng/public/_build/results?buildId=360236&view=logs&j=7df7e7c1-9be0-5cba-8a92-3fe105a7dcf9

This is blocking us from upgrading our images to the ones built on 2019-09-18 with the changes in #185

jkoritzinsky:
cc: @MichaelSimons @mthalman @jashook @janvorli

MichaelSimons:
@mthalman - it looks like the changes to onboard this repo onto the common Docker build infra lost the fix for https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/144 - see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/156/files#diff-fbb8382709c1eeda44e133299b934e0dL34.

MichaelSimons:
@jkoritzinsky - I patched the build and rebuilt the 16.04 images.  Can you validate?

ubuntu-16.04-coredeps-d9d81d0-20190919190059
ubuntu-16.04-crossdeps-cfdd435-20190919190059
ubuntu-16.04-cross-arm64-cfdd435-20190919190059
ubuntu-16.04-cross-arm64-alpine-406629a-20190919190059
ubuntu-16.04-cross-23cacb0-20190919190059
ubuntu-16.04-cross-14.04-23cacb0-20190919190059

jkoritzinsky:
@MichaelSimons those images worked. Should I just use those ones or should I wait for #188 to get in to master?

MichaelSimons:
@jkoritzinsky - you can use those images if they work for you (but I saw you made another PR).

jkoritzinsky:
Yeah I realized that I missed one image in my last change.

MichaelSimons:
Closing this issue since the Ubuntu 16.04 cross images were addressed.  I will log a new issues to solve the long term issues within the build infrastructure.

PR #188, opened 2019-09-19
Adding temporary workaround for issue-187

MichaelSimons:
This is a patch to workaround #187


MichaelSimons:
Merging to unblock builds.

PR #189, opened 2019-09-20
Install modern CMake on Ubuntu 16.04 cross-build images.

jkoritzinsky:
I missed that the cross images aren't based on the regular Ubuntu 16.04 image.

This PR combined with the fix for #187 will enable CoreCLR to move up to CMake 3.14.

cc: @MichaelSimons

### DIFF
--- a/src/ubuntu/16.04/crossdeps/Dockerfile
+++ b/src/ubuntu/16.04/crossdeps/Dockerfile
@@ -2,7 +2,12 @@ FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-coredeps
 
 # Install the base toolchain we need to build anything (clang, cmake, make and the like).
 RUN apt-get update \
-    && apt-get install -y apt-transport-https ca-certificates gnupg software-properties-common wget \
+    && apt-get install -y \
+        apt-transport-https \
+        ca-certificates \
+        gnupg \
+        software-properties-common \
+        wget \
     && wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc | apt-key add - \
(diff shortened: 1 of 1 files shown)


src/ubuntu/16.04/crossdeps/Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-coredeps
 
 # Install the base toolchain we need to build anything (clang, cmake, make and the like).
 RUN apt-get update \
+    wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc | apt-key add - && \
```
MichaelSimons: This syntax doesn't look right,  I was expecting a `&&` after the initial `apt-get update`.  Is the initial call to `apt-get update` even necessary?  Please put the `&&` at the beginning of the lines as this is the convention typically used within the Dockerfiles of this repo.

jkoritzinsky:
I've cleaned it up per your feedback.

src/ubuntu/16.04/crossdeps/Dockerfile
```diff
@@ -2,6 +2,10 @@ FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-coredeps
 
 # Install the base toolchain we need to build anything (clang, cmake, make and the like).
 RUN apt-get update \
+    && apt-get install -y apt-transport-https ca-certificates gnupg software-properties-common wget \
```
MichaelSimons: Could you format this as done below where each package is on a separate line?  This helps the readability so that you can more easily pick out which packages are included in the image.  Additionally wget can be removed below since it is now included here.

PR #197, opened 2019-10-18
Unify clang in containers used to build .NET Core to version 9

janvorli:
For CentOS, we need to build clang 9 and lldb 9 from sources. On CentOS 6,
it is more tricky, as we need to build a recent version of swig to enable
building clang 9.
For Ubuntu, we can install it from packages, but there is another issue.
The lldb-9 package has broken dependencies and cannot be installed.
So we install lldb-6.0 instead, which is sufficient for the diagnostics
repo build.


janvorli:
cc: @jashook

src/centos/6/Dockerfile
```diff
@@ -4,20 +4,22 @@ FROM centos:6
 
 RUN yum install -y \
         centos-release-SCL \
+        scl-utils-build \
```
MichaelSimons: Can this be alphabetized?  Same comment applies in other places as well.

janvorli: Done

janvorli:
There is something flaky about the ubuntu 16.04 image build. It builds fine locally.

janvorli:
Ah, I can see the problem now. The Ubuntu 16.04 image build is not failing, but rather another image that is based on that (mlnet). There is a problem with libomp-dev installation. I need to look into that.

janvorli:
@MichaelSimons the centos 6 and 7 images build times out in about 73% of building of the clang, so that will obviously never succeed in the CI.
The arm64v8 debian 10 fail seems to be some infra issue.

mthalman:
@janvorli - Because of the longer build time now, please update the timeout for the CentOS 6 and 7 builds here (probably add another hour, 180 minutes total):
* https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/eng/pipelines/dotnet-buildtools-prereqs-centos6.yml#L18
* https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/eng/pipelines/dotnet-buildtools-prereqs-centos7.yml#L18.

PR #200, opened 2019-10-29
Ubuntu 19.10 Docker image

jcagme:
Fixes https://github.com/dotnet/core-eng/issues/8181 which is meant to create an Ubuntu 19.10 Docker image which installs the `helix-scripts` wheel


src/ubuntu/19.10/helix/amd64/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+    python -m pip install helix-scripts --extra-index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-staging/pypi/simple
+
+# create helixbot user and give rights to sudo without password
+RUN /usr/sbin/adduser --disabled-password --gecos '' --uid 1000 --shell /bin/bash --ingroup adm helixbot && \
+    chmod 755 /root && \
+    echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers
+
+USER helixbot
+
+RUN python -m virtualenv --no-site-packages /home/helixbot/.vsts-env
```
mthalman: Please add a trailing newline.

jcagme:
> There also needs to be a change to the official pipelines. By default, there are separate YAML pipelines for each major version of Ubuntu but the "Ubuntu 19" folder is explicitly referencing the 19.04 folder (see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/eng/pipelines/dotnet-buildtools-prereqs-ubuntu19.yml#L7). Now that you've added a 19.10 folder, something will need to change. Either the include path will need to be updated to account for 19.04 and 19.10 or a new pipeline should be added just for 19.10. Which way to go depends on whether the Dockerfiles between 19.04 and 19.10 are likely to be updated together. If they're likely to updated together in a single PR, it makes sense to have a single pipeline. Otherwise, there should be a separate pipeline.

I don't have enough context on how things are normally updated to know whether to create a new pipeline or keep one. @MattGal, thoughts?

eng/pipelines/dotnet-buildtools-prereqs-ubuntu19.yml
```diff
@@ -4,13 +4,13 @@ trigger:
...
     include:
-    - src/ubuntu/19.04/*
+    - src/ubuntu/*
 pr: none
 
 variables:
 - template: variables/common.yml
 - name: imageBuilder.pathArgs
-  value: --path 'src/ubuntu/19.04*'
+  value: --path 'src/ubuntu/*'
```
mthalman: There are already YAML pipelines defined for 14, 16, and 18.  This should explicitly only include 19*.

```suggestion
  value: --path 'src/ubuntu/19*'
```

eng/pipelines/dotnet-buildtools-prereqs-ubuntu19.yml
```diff
@@ -4,13 +4,13 @@ trigger:
     - master
   paths:
     include:
-    - src/ubuntu/19.04/*
+    - src/ubuntu/*
```
mthalman:
```suggestion
    - src/ubuntu/19*
```

PR #203, opened 2019-11-06
Update CentOS images to install CMake 3.15.5.

jkoritzinsky:
The minimum CMake version for the consolidated repository is going to be CMake 3.15.5 to account for some fixes in CMake for their VS generators.

@MichaelSimons can you also rebuild the Ubuntu 16.04 image, the Ubuntu 16.04 crossdeps image, and the various images that depend on those images? They need to pull an updated CMake version from the package feed.

### DIFF
--- a/src/centos/6/Dockerfile
+++ b/src/centos/6/Dockerfile
@@ -45,9 +45,9 @@ RUN yum install -y \
     && \
     yum clean all
 
-# Build and install CMake 3.15.3
+# Install CMake 3.15.5
 
-RUN wget -O cmake-install.sh https://github.com/Kitware/CMake/releases/download/v3.15.3/cmake-3.15.3-Linux-x86_64.sh && \
+RUN wget -O cmake-install.sh https://github.com/Kitware/CMake/releases/download/v3.15.5/cmake-3.15.5-Linux-x86_64.sh && \
     bash ./cmake-install.sh --skip-license --exclude-subdir --prefix=/usr/local && \
     rm ./cmake-install.sh
 
--- a/src/centos/7/Dockerfile
+++ b/src/centos/7/Dockerfile
@@ -44,9 +44,9 @@ RUN yum install -y \
     && \
     yum clean all
 
-# Build and install CMake 3.15.3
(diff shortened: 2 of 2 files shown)


MichaelSimons:
@jkoritzinsky - There looks to be a failure while building the ubuntu-16.04 cross images while running build-rootfs.sh.  see https://dev.azure.com/dnceng/public/_build/results?buildId=417916&view=logs&j=7e9b14cd-47a0-562c-e7e8-7f0ef272e133

```
ERROR: unsatisfiable constraints:
  so:libpython3.7m.so.1.0 (missing):
```

jkoritzinsky:
It looks like that might have been related to the lldb 9.0 change. This PR itself only changes the CentOS 6 and 7 Dockerfiles.

cc: @janvorli any ideas why there's dependency issues with lldb 9.0?

MichaelSimons:
@jkoritzinsky, yes I agree it is not related to the changes in this PR.  I bring it up because of the request to rebuild the Ubuntu 16.04 images.

janvorli:
It seems the lldb 9 got there by a chance due to the fact that LLDB is in the edge/testing branch, so it keeps getting rolling updates. Unfortunately lldb is not in any stable branches at all, so almost every time we rebuild, it breaks us. I will check if we can fix it again somehow, or maybe get rid of the lldb from the alpine image completely. That could cause trouble for the stuff built in the diagnostics repo but I think we don't produce packages of those tools for Alpine anyways. @mikem8361 am I right?

mikem8361:
For the alpine images the diagnostics repo does need the lldb-devel (include files) installed so it can build the plugin even though lldb still doesn't work that well though I might be wrong about 9.0 lldb. 

So removing lldb completely would break the diagnostics repo when/if it upgrades to the new images.

janvorli:
Thanks @mikem8361 for the details. I'll try to fix it then.

MichaelSimons:
I opened an [issue](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/204) for the failing ubuntu-16.04 cross build.

PR #205, opened 2019-11-06
Update build-rootfs to clone arcade instead

ViktorHofer:
The core-setup /cross folder will go away therefore we are cloning arcade instead.


ViktorHofer:
Merging to unblock CI here as the cross folder isn't present anymore in core-setup.

PR #206, opened 2019-11-15
Remove Dockerfiles for Windows Server 1803

mthalman:


MichaelSimons:
cc @MattGal - FYI, 1803 is EOL.  As a result, we are dropping build support for 1803 images.

PR #236, opened 2020-01-22
Fix using CMake on alpine 3.9 docker images

jkoritzinsky:
The version of CMake in the edge and 3.11 Alpine repositories are built against a new version of libstdc++ that exposes new symbols. As a result, CMake from 3.11+ Alpine won't run on <=3.10 Alpine. Change the cmake install to install from the 3.10 Alpine repository as a result.

This change is to support https://github.com/dotnet/runtime/issues/2030.

cc: @mthalman @janvorli

### DIFF
--- a/src/alpine/3.9/amd64/Dockerfile
+++ b/src/alpine/3.9/amd64/Dockerfile
@@ -32,7 +32,7 @@ RUN apk add --no-cache \
         which \
         zlib-dev
 
-RUN apk -X http://dl-cdn.alpinelinux.org/alpine/v3.11/main add --no-cache \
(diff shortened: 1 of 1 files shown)


mthalman:
@jkoritzinsky - Did you want to wait for a review from @janvorli before this gets merged?

jkoritzinsky:
We can merge this in without his review.

janvorli:
I've added a comment yesterday, but obviously forgotten to push the send comment and I haven't realized that. My comment was that we need to  make sure that the stuff we build using this works on Alpine 3.8. My worry was that cmake depends on 10 other packages and might pull in something that would break backwards compatibility.

PR #245, opened 2020-01-30
Add Python 3 to CentOS and Alpine images for Mono

### LINKED ISSUE #244: Add Python 3 to images
Mono needs Python 3 for a bunch of our scripts. I attempted to enforce this in configure.ac, but it looks like the Linux images are all missing it (see https://github.com/dotnet/runtime/pull/2372).

CoffeeFlux:
Fixes #244 

I'm not very familiar with Docker, so if there's a better way to do any of this please let me know.

### DIFF
--- a/src/alpine/3.9/amd64/Dockerfile
+++ b/src/alpine/3.9/amd64/Dockerfile
@@ -26,6 +26,8 @@ RUN apk add --no-cache \
         openssl-dev \
         paxctl \
         python \
+        python3 \
+        python3-dev \
         shadow \
         sudo \
         util-linux-dev \
--- a/src/alpine/3.9/arm32v7/Dockerfile
+++ b/src/alpine/3.9/arm32v7/Dockerfile
@@ -27,6 +27,8 @@ RUN apk add --no-cache \
         openssl-dev \
         paxctl \
         python \
+        python3 \
+        python3-dev \
         shadow \
         sudo \
         util-linux-dev \
--- a/src/alpine/3.9/arm64v8/Dockerfile
+++ b/src/alpine/3.9/arm64v8/Dockerfile
@@ -27,6 +27,8 @@ RUN apk add --no-cache \
         openssl-dev \
         paxctl \
         python \
+        python3 \
+        python3-dev \
         shadow \
         sudo \
         util-linux-dev \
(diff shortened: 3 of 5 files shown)


directhex:
Do we want to change our Alpine image vendor from Alpine to the wider Docker community? I don't know the answer to that.

CoffeeFlux:
I think we do that for the helix images already, unless I'm misunderstanding. See https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/6298b198185e12926ad3c9ad55bb6d00674190c6/src/alpine/3.9/helix/amd64/Dockerfile#L1

directhex:
:+1:

mthalman:
Why are only the Alpine 3.9 Dockerfiles being updated?  Alpine 3.9 is actually EOL now.

CoffeeFlux:
As far as I can tell, the other ones are just for Helix and already have Python 3.7. I need it on the ones used for building Mono on the dotnet/runtime CI.

directhex:
https://github.com/dotnet/runtime/blob/master/eng/pipelines/common/platform-matrix.yml#L90-L95

mthalman:
Ah, never mind.  Alpine 3.9 isn't EOL until Jan. 2021.  I was getting it confused with our own support of Alpine 3.9 for our .NET images.

src/alpine/3.9/amd64/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.9
+FROM python:3.7-alpine3.9
```
MichaelSimons: I'm not seeing this listed as being a supported tag any longer.  Because of that, I don't think we should be taking a dependency on it.  The consequence of this is the image would not be updated as patches are made to the alpine:3.9 image.

MichaelSimons: It is beneficial to keep alignment between the Dockerfiles for the various architectures.  By this I mean the base images should align between this Dockerfile and the [corresponding arm Dockerfile](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/245/files#diff-81ce923e0871c5773a95153cf8e42270).

CoffeeFlux: Good to know, thanks. Updated.

Issue #92, opened 2019-02-28
Add build image for RPM packages


dagood:
We have e.g. `ubuntu-14.04-debpkg` for Debian packages, we need something similar to build RPM packages. This would let Core-Setup and others use an actively supported build image when producing RPM packages.

I think we just need to create a new `centos-7-rpmpkg` Dockerfile that extends the CentOS Dockerfile with:

```
RUN yum clean all \
    && yum install -y \
        gcc \
        make \
        rpm-build \
        ruby-devel \
        sudo \
    && gem install --no-ri --no-rdoc fpm \
    && yum clean all
```

This isn't blocking me.

MichaelSimons:
@dagood - Go ahead and create a PR please.  It is expected that the product teams manage the Dockerfiles they need specific to their scenarios.

dagood:
I plan on it when I get a chance to validate it and go through with consuming it in the Core-Setup build infra, or someone from e.g. Core-SDK or ASP.NET Core might want to do this too. Is it bad to open this kind of tracking issue here?

MichaelSimons:
Logging the issue here is great.  I just wanted to clarify expectation with everyone. 😺

dagood:
This is also useful for dev builds, would be worth tackling even if we don't go and *immediately* put it in the official build. /cc @NikolaMilosavljevic

MichaelSimons:
Fixed by #252.

PR #252, opened 2020-02-18
Dockerfile for RPM packaging, based on CentOS 7

NikolaMilosavljevic:
Fix for issue https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/92

### DIFF
--- a/src/centos/7/rpmpkg/Dockerfile
+++ b/src/centos/7/rpmpkg/Dockerfile
@@ -1,0 +1,11 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
+
+# Install the rpm packaging toolchain we need to build rpms
+RUN yum install -y \
+        gcc \
+        make \
+        rpm-build \
+        ruby-devel \
+        rubygems \
+        sudo \
+    && gem install --no-document fpm
--- a/src/centos/7/rpmpkg/Dockerfile
+++ b/src/centos/7/rpmpkg/Dockerfile
@@ -1,6 +1,6 @@
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
 
-# Install the rpm packaging toolchain we need to build rpms
+# Install the rpm packaging toolchain required to build rpms
 RUN yum clean all \
     && yum install -y \
         gcc \
--- a/src/centos/7/rpmpkg/Dockerfile
+++ b/src/centos/7/rpmpkg/Dockerfile
@@ -1,11 +1,12 @@
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
 
 # Install the rpm packaging toolchain we need to build rpms
-RUN yum install -y \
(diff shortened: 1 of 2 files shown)


src/centos/7/rpmpkg/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
+
+# Install the rpm packaging toolchain we need to build rpms
+RUN yum install -y \
+        gcc \
+        make \
+        rpm-build \
+        ruby-devel \
+        rubygems \
+        sudo \
```
MichaelSimons: Should this be doing a `yum clean all` at the end?

src/centos/7/rpmpkg/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
+
+# Install the rpm packaging toolchain we need to build rpms
+RUN yum install -y \
+        gcc \
+        make \
```
MichaelSimons: `make` and `sudo` are already installed in the base image.

src/centos/7/rpmpkg/Dockerfile
NikolaMilosavljevic: I'm not sure - I've used this file as model: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/src/centos/7/mlnet/Dockerfile

@dagood you had specified "yum clean" commands in the suggested solution - issue https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/92. Are those commands needed?

src/centos/7/rpmpkg/Dockerfile
NikolaMilosavljevic: I'll remove them. Didn't seem to make a difference, but want to follow the best practices.

src/centos/7/rpmpkg/Dockerfile
dagood: It's a guideline, but yes, the `yum` cache data is not useful for the image user and should be cleaned up in the same `RUN` statement as the install commands. The user either won't run any more `yum` commands and have no use of a cache, but even if they do install more packages, they will probably find the cache data is old and useless.

src/centos/7/rpmpkg/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
+
+# Install the rpm packaging toolchain we need to build rpms
```
MichaelSimons: Nit - consider avoiding `we` in the comment - maybe something like `Install the rpm packaging toolchain required to build rpms

NikolaMilosavljevic: Fixed.

PR #256, opened 2020-03-04
Update to build a 16.04 rootfs

jashook:
14.04 is eol, we should not be building the rootfs using it, even if it is just for the crossbuild.


jashook:
@jkoritzinsky @trylek ptal

jkoritzinsky:
cc: @janvorli

MichaelSimons:
I agree 14.04 should be removed.  That being said how is the new proposed image different than the one produced [here](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/manifest.json#L303)?

jashook:
I will confirm; however, I believe that one does not include the rootfs for x86 and armhf

jashook:
Nope, they are the same. I will remove the 16.04 folder.

jashook:
Ok this should be ready for review now, addressing @MichaelSimons comment

PR #262, opened 2020-03-11
Create gcc-7 symlink available in PATH

am11:
Related to https://github.com/dotnet/runtime/issues/33346

### DIFF
--- a/src/centos/7/Dockerfile
+++ b/src/centos/7/Dockerfile
@@ -92,3 +92,6 @@ RUN wget ftp://sourceware.org/pub/binutils/snapshots/binutils-2.29.1.tar.xz && \
     rm -r llvmbuild && \
     rm -r llvm-9.0.0.src && \
     rm -r binutils-2.29.1
+
+# Create gcc-7 symlink available in PATH
+RUN ln -s /opt/rh/devtoolset-7/root/usr/bin/gcc /usr/local/bin/gcc-7


am11:
@safern, please take a look. :)
GCC 9 is not strictly necessary, CoreCLR builds with GCC v5 up to v9. Current donet/net master has some errors when building with GCC, but that is a running problem for which we adding the proposed pipeline (i am in the middle of fix for both GCC7 or GCC9 for runtime PR).
If there are no objections and mono is also ok with GCC 9, we can go ahead and use the current latest version 9, instead of 2+ years old version (as we are using the latest Clang v9 from LLVM toolchain).

src/centos/7/Dockerfile
```diff
@@ -78,17 +78,23 @@ RUN wget ftp://sourceware.org/pub/binutils/snapshots/binutils-2.29.1.tar.xz && \
...
     cd .. && \
     rm -r llvmbuild && \
     rm -r llvm-9.0.0.src && \
     rm -r binutils-2.29.1
+
+# Add gcc in PATH
+ENV PATH /opt/rh/devtoolset-9/root/usr/bin:$PATH
+
+# Clear yum cache to save ~150 MB
+RUN yum clean all
```
mthalman: I'm not understanding why this is needed.  The yum installation is done way up top and it already runs a clean command at that point.  Why is it necessary at this point?

mthalman:
@am11 - Please investigate the build failure.  Ignore the `StandardError has not been redirected` message; that's an infrastructure issue.  But the errors above that are relevant.

src/centos/7/Dockerfile
am11: I have reverted this change. I was calculating size impact of container at this point and comparing it what I had when running this container interactively on localhost; didn't realized that we are already doing the cleanup.

safern:
> GCC 9 is not strictly necessary, CoreCLR builds with GCC v5 up to v9. Current dotnet/runtime master has some errors when building with GCC, but that is a running problem for which we adding the proposed pipeline (i am in the middle of fix for both GCC7 or GCC9 for runtime PR).
If there are no objections and mono is also ok with GCC 9, we can go ahead and use the current latest version 9, instead of 2+ years old version (as we are using the latest Clang v9 from LLVM toolchain).

t seems reasonable to me, but who might have other insights on this. @janvorli @jkotas @dotnet/runtime-infrastructure

jkotas:
The GCC version has to work for Mono. I assume there was a reason why @akoeplinger added this specific version of GCC and not the latest.

src/centos/7/mlnet/Dockerfile
```diff
@@ -3,6 +3,9 @@ FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
 RUN yum install -y \
         perl-Data-Dumper
 
+# Sets the library path to pickup openmp
```
safern: Why is this needed?

am11: Actually this was the primary reason for this PR to add GCC’s based it in PATH (so we can cleanup CLR_{CC, CXX} from pipeline YAML, which we are using in runtime PR as a temporary workaround). While at it, I thought of revving the version from v7 to v9, and that became the title. If @akoeplinger and co. suggest that we should stick to v7, I will remove the rest of the changes except this.

safern: But this is just moving this from the bottom of the file to line 6. This was already part of this file. Also, this dockerfile is for the mlnet repo, we don't use this to build the runtime repo.

am11: Ah, sorry, I misread the filename (was on my phone and thought it is about PATH export line ..  GH is not showing the next line) 😄 
In the previous CI run, where it tries to compile mlnet with the just-compiled Clang (using GCC 9 devtoolset in the base layer), it failed to locate system libs. The error was:

```sh
    /usr/bin/ld: cannot find -lstdc++
```

I moved this line above in the anticipation, that it will pick up the correct LD_LIBRARY_PATH this time; now waiting for LLVM build to finish to see the effect.

am11: Still failed. It is actually failing to build openmp (not mlnet) after compiling llvm project:
```
Step 4/4 : RUN wget http://releases.llvm.org/3.9.1/openmp-3.9.1.src.tar.xz &&     mkdir -p llvm-3.9.1.src/openmp &&     tar -xf openmp-3.9.1.src.tar.xz --strip 1 -C llvm-3.9.1.src/openmp &&     rm openmp-3.9.1.src.tar.xz &&         mkdir llvmbuild &&     cd llvmbuild &&     cmake         -DCMAKE_CXX_COMPILER=/usr/local/bin/clang++         -DCMAKE_C_COMPILER=/usr/local/bin/clang         -DCMAKE_BUILD_TYPE=Release         -DLLVM_LIBDIR_SUFFIX=64        -DLLVM_ENABLE_EH=1         -DLLVM_ENABLE_RTTI=1         ../llvm-3.9.1.src/openmp     &&     make -j $(($(getconf _NPROCESSORS_ONLN)+1)) &&     make install &&     cd .. &&     rm -r llvmbuild &&     rm -r llvm-3.9.1.src
 ---> Running in b487c9579b28
[snip]
    /usr/bin/ld: cannot find -lstdc++
```

am11:
The LLVM v9 built with GCC v9 toolchain was missing some targets. This is my conclusion after diffing various parts of 15K+ lines of CI logs. Not sure why only changing the toolchain version would compile different LLVM targets (from same source package of LLVM).

Therefore, I have repurposed this PR to only create a symlink for `gcc-7`. This way it is visible in PATH and nothing else gets a conflict; which was our primary (totally optional) need for CoreCLR GCC CI leg, which is already green by means of environment variables.

In future, it would probably be good to understand gcc9-compiled llvm9 is different than gcc7-compiled one. (or maybe use prebuilt llvm9 toolset from package manager, assuming they have compiled with all the features required by dotnet/runtime: https://git.centos.org/rpms/llvm/blob/676fa747a0dc8819e6509eb3947cb14714dc8b09/f/SPECS/llvm.spec#_207).

akoeplinger:
> Mono. I assume there was a reason why @akoeplinger added this specific version of GCC and not the latest.

I don't think we added a specific version of GCC, it just came as a dependency of some other tool.
We force clang in dotnet/runtime for the mono build at the moment.

am11:
Ideally, if "compiler installation" step is a separate docker layer for clang and gcc, there will be no conflicts in LD_LIBRARY_PATH and other configurations, and we will have clean environments for each with ability to switch versions independently.

PR #267, opened 2020-03-23
Add Fedora 31 build image.

crummel:
This is mostly a copy of the Fedora 30 image with some packages that are no longer available removed.  I was able to build source-build on this image so the missing packages do not seem to be required for any core product builds.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -178,6 +178,19 @@
           }]
         },
         {
+          "platforms": [{
+            "dockerfile": "src/fedora/31/amd64",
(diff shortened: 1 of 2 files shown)


mthalman:
You also need to add an entry to the manifest file to reference this new Dockerfile.

See example: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/09ec75772538159cfc508bab041c882ff09b0012/manifest.json#L159

crummel:
Thanks, fixed.

PR #268, opened 2020-03-23
Add the correct cmake version

jashook:
Copy/paste issue removed the correct cmake version


jashook:
@mthalman ptal

jashook:
This is ready to be merged imo

PR #270, opened 2020-03-26
Add new Debian and Ubuntu source-build images with updated CMake for source-build.

crummel:
I thought that it was probably appropriate to split out these source-build images since we're installing stuff from the backports-sloppy feed and downloading specific tarballs.  If these changes are fine in the main images I can move them there instead.


MichaelSimons:
I agree these changes don't belong in the main images.  The question I have however is if there is value in basing the SB images off of the main images or factor out the commonality into a new shared base image.  Do you want SB to pick up any of the other changes that may be made to the main images?

crummel:
That's a good point, I'll refactor these into being based off the main images.  Thanks!

manifest.json
```diff
@@ -181,6 +181,18 @@
...
           ]
         },
+        {
+          "platforms": [
+            {
+              "dockerfile": "src/debian/stretch/source-build",
+              "os": "linux",
+              "osVersion": "stretch",
+              "tags": {
+                "debian-stretch-source-build-$(System:DockerfileGitCommitSha)-$(System:TimeStamp)": {}
```
mthalman: For this tag and the other 18.04 tag, can you be sure to swap the position of the commit SHA and timestamp?  See #275

Issue #272, opened 2020-03-27
Builds failing because of "no space left on device"


akoeplinger:
After merging https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/271 I saw some of the builds now consistently failing with "no space left on device" errors.

Example: https://dev.azure.com/dnceng/internal/_build/results?buildId=575510&view=results (I retriggered four times).

@mthalman @MichaelSimons any ideas? I assume because that build runs on the Azure Pipelines hosted pool which has limited disk space and adding one more image caused it to run out.

Can we somehow clean temporary files between each image build?

mthalman:
I attempted to run a build on a self-hosted Linux agent that we have but that ran out of disk space as well.  Continuing to investigate.

MichaelSimons:
@mthalman was able to workaround the disk space issue by using a different build pool with larger drive space.  You can find the new images from #271 documented at [Latest tags](https://github.com/dotnet/versions/blob/master/build-info/docker/image-info.dotnet-dotnet-buildtools-prereqs-docker-master.json).

I found there was an issue with the rootfs cleanup logic that freed up a significant amount of space during the build.  This issue was addressed in #274.  Closing.

akoeplinger:
Thank you 🎉

PR #274, opened 2020-03-27
Fix broken rootfs cleanup logic

MichaelSimons:
I discovered the logic that cleans up the rootfs tarballs was no longer working because of various refactoring over time.  For the ubuntu 18.04 leg this equates to 4.68 GB on tarballs.

I also increased the ubuntu leg timeout.

Related to #272

### DIFF
--- a/eng/pipelines/dotnet-buildtools-prereqs-ubuntu18.yml
+++ b/eng/pipelines/dotnet-buildtools-prereqs-ubuntu18.yml
@@ -21,6 +21,6 @@ variables:
 stages:
   - template: ../common/templates/stages/build-test-publish-repo.yml
     parameters:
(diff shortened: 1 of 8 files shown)


MichaelSimons:
Hold off on reviewing this, I found more issues.

MichaelSimons:
This is ready for review.

MichaelSimons:
From my investigation, all of the cross builds were calling post-build correctly.  The issue is the files are supposed to be defined as file links so there is only one copy of the source in the repo.  When I modified the shared file, I discovered all of the 18.04 files were not file links and therefore didn't pick up my change.  This was most likely due to coping the files from 16.04 (maybe on a Windows machine).

PR #284, opened 2020-04-01
Add locales to ubuntu 16.0.4

jaredpar:
The MSBuild `Exec` task will use `export LC_ALL=en_US.UTF-8` as a
part of the `bash` invocation. When ubuntu lacks the `locales`
package this will cause a warning to be emitted from the `Exec`
usage and the outer MSBuild invocation. Adding the `locales`
package here to get our builds warning free as this makes it
significantly easier to manage our build reliability

https://github.com/microsoft/msbuild/issues/4194
https://github.com/dotnet/runtime/issues/34280

### DIFF
--- a/src/ubuntu/16.04/Dockerfile
+++ b/src/ubuntu/16.04/Dockerfile
@@ -19,6 +19,8 @@ RUN apt-get update && \
             llvm-9 \
             make \
             python-lldb-6.0 \
+            locales \
+            locales-all \
             sudo && \
     apt-get clean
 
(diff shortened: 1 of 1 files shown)


jaredpar:
@wfurt @ViktorHofer @directhex

src/ubuntu/16.04/Dockerfile
```diff
@@ -19,6 +19,7 @@ RUN apt-get update && \
             llvm-9 \
             make \
             python-lldb-6.0 \
+            locales \
```
mthalman, line 22: Please place this alphabetically in the list of packages.

jaredpar:
> Note that Alpine does not have locale and I saw similar warnings

@ViktorHofer and @safern brought this up and it seems we've moved to a new image where we're no longer hitting this anymore. Will definitely update if we hit that again though.

PR #310, opened 2020-05-06
Update Ubuntu Dockerfile to configure system-wide locale

mstfbl:
Related to [ML .NET Issue #5093](https://github.com/dotnet/machinelearning/issues/5093)

The ML .NET Ubuntu Docker instance needs a system-wide locale set. We are re-enabling our ONNX Runtime tests to run on Linux, and this requires installing the language-pack-en package,as noted [here](https://github.com/microsoft/onnxruntime/blob/master/README.md#system-language).


safern:
cc: @MichaelSimons @mthalman

PR #330, opened 2020-06-12
Define ARM64 version of testrunner

mthalman:
We need an ARM64 version of testrunner so that it can be run on ARM machines.

ARM builds of PowerShell are not available in the Debian package feeds so it is downloaded directly from the GitHub releases URL.  

Defined a multi-arch tag for testrunner to make it simpler to reference.  To avoid issues with needing to have the same timestamp value across multiple build jobs, I decided to just remove that segment of the tag.  Also had to remove the commit SHA from the tag because Image Builder doesn't support that built-in variable in shared tags since there's no build context from which to get a commit.  This uniqueness doesn't really seem necessary for how little this image gets updated.  Let me know what you think.  Otherwise, I'll need to add a custom variable to provide uniqueness.

Related to https://github.com/dotnet/docker-tools/issues/486

### DIFF
--- a/src/debian/stretch-slim/docker-testrunner/arm64v8/Dockerfile
+++ b/src/debian/stretch-slim/docker-testrunner/arm64v8/Dockerfile
@@ -1,0 +1,28 @@
+# Dockerfile used to create a testrunner image that can perform Docker operations.
+# Usage:  docker run --rm -v /var/run/docker.sock:/var/run/docker.sock testrunner pwsh -File xyz.ps1
+
+FROM arm64v8/debian:stretch-slim
+
+# Install Docker
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        gnupg2 \
+        libicu57 \
+        software-properties-common \
+    && curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - \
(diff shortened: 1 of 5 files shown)


src/debian/stretch-slim/docker-testrunner/arm64v8/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+FROM arm64v8/debian:stretch-slim
+
+# Install Docker
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        gnupg2 \
+        libicu57 \
```
MichaelSimons, line 13: Why is this required in the arm64 variant but not x64?

mthalman: It is required.  But since the x64 version installs PowerShell via a package, libicu gets installed as a dependency automatically.  In the arm64 variant, we're installing PowerShell manually, so we need to explicitly include libicu.

mthalman:
> I think the shared tag needs to be unique per change. The components in the Docker image can have breaking changes which we should be able roll-out in stages.

Added a UniqueId to the tag, just as was done for [Image Builder](https://github.com/dotnet/docker-tools/pull/553).

PR #350, opened 2020-07-27
Add Docker image for PPC64 cross-compile

directhex:
This targets a cross of Debian Unstable, because there's no easy way right now to produce a crossrootfs to target an RPM-based distribution (CentOS 7 would be preferred), and PPC64 is an unofficial Debian port without any stable release candidate.

s390x would be a nicer big-endian target to work with, but we have PPC inside corpnet, whilst we'd need to use externally hosted test hardware for s390x

Ref: https://github.com/dotnet/arcade/pull/5841
Ref: https://garagehackbox.azurewebsites.net/hackathons/2107/projects/91897

### DIFF
--- a/src/ubuntu/18.04/cross/ppc64/Dockerfile
+++ b/src/ubuntu/18.04/cross/ppc64/Dockerfile
@@ -1,0 +1,9 @@
+FROM ubuntu-18.04-crossdeps
+
+# Install binutils-powerpc64-linux-gnu
+RUN apt-get update \
+    && apt-get install -y \
+        binutils-powerpc64-linux-gnu \
+    && rm -rf /var/lib/apt/lists/*
+
+ADD rootfs.ppc64.tar crossrootfs
--- a/src/ubuntu/18.04/cross/ppc64/Dockerfile
+++ b/src/ubuntu/18.04/cross/ppc64/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu-18.04-crossdeps
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-crossdeps
 
 # Install binutils-powerpc64-linux-gnu
 RUN apt-get update \
--- a/manifest.json
+++ b/manifest.json
@@ -452,6 +452,18 @@
         {
           "platforms": [
             {
+              "dockerfile": "src/ubuntu/18.04/cross/ppc64",
+              "os": "linux",
+              "osVersion": "bionic",
+              "tags": {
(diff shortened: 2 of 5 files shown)


directhex:
This'll probably fail horridly until the Arcade PR is merged

mthalman:
In order for this to get built, it needs to be added to the manifest.  See this as an example: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/3ab9f445cb7ee3aa4e70f92a108cf1589da4a461/manifest.json#L515

directhex:
Ah. I was looking in the yaml, didn't realise there was a second place things were defined

src/ubuntu/18.04/cross/ppc64/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu-18.04-crossdeps
```
mthalman, line 1:
```suggestion
FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-crossdeps
```

directhex:
Argh. Yes, my fault. I had locally removed the registry prefix, whilst testing, and forgot to change it back

mthalman:
@directhex - Any ideas why the call to the build-rootfs.sh script worked since you expected it to fail due to the pending dotnet/arcade#5841 change?

directhex:
I think the fallback behaviour in the script on unknown params is x64 behaviour, so it probably just created an x64 crossrootfs. I'll check the logs.

directhex:
Yeah, more or less, but ARM.

```
Using mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-crossdeps to set up cross-toolset for ppc64 for sid
13e0d77c3d695ed5c0431647d30cd90b3352f3d5552154c867590d05dfb633ef
Using mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-crossdeps to clone arcade to fetch scripts used to build cross-toolset
Cloning into '/scripts'...
Running build-rootfs.sh
I: Running command: debootstrap --arch armhf --foreign xenial /rootfs/ppc64 http://ports.ubuntu.com/
```

The issue is in build-rootfs.sh, which picks a dumb default on unknown params, instead of failing

mthalman:
Ok, so this shouldn't be merged until dotnet/arcade#5841 is merged, right?

directhex:
Definitely, definitely not.

directhex:
Mostly I opened this PR so I wouldn't forget to do it later. Wish it had failed on the "bad" rootfs!

MichaelSimons:
@directhex, I see dotnet/arcade#5841 was closed as not a priority at this time.  What is the impact on this PR?  Should this remain open or should it be closed until it raises in priority?

PR #363, opened 2020-08-26
Disable building of ARM32 Helix images

mthalman:
Disabling these images to workaround #359 and unblock builds.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -578,20 +578,6 @@
         {
           "platforms": [
             {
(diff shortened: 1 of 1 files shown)


mthalman:
> The changes look good. It makes me wonder if we should add a disable metadata flag to better accommodate scenarios in which we want to temporarily not build an image. Thoughts?

Same thought occurred to me. 👍

PR #370, opened 2020-09-18
Fix repo path to install cross build prereqs

mthalman:
The path had been hardcoded but the name of the repo actually varies based on the name of the pipeline, of which there are several for this repo.


mthalman:
cc @dagood

MichaelSimons (approved):

PR #392, opened 2020-12-27
Update devtoolset-7 to devtoolset-9

am11:
Currently, old gcc v7 is used to custom compile llvm-toolchian v9 (this llvm-toolchain is used for official builds). Same gcc version is used in a (single) leg that runtime repo has for gcc build validation.

This PR updates gcc to v9x (currently at 9.3), which continues to compile llvm-toolchain v9, and makes it possible to validate runtime build with recent gcc version.

### DIFF
--- a/src/centos/7/Dockerfile
+++ b/src/centos/7/Dockerfile
@@ -11,7 +11,7 @@ RUN yum install -y \
         "perl(Time::HiRes)" \
         autoconf \
         automake \
-        devtoolset-7-gcc* \
+        devtoolset-9-gcc* \
         doxygen \
         gdb \
         git \
@@ -78,7 +78,7 @@ RUN wget ftp://sourceware.org/pub/binutils/snapshots/binutils-2.29.1.tar.xz && \
     \
     mkdir llvmbuild && \
     cd llvmbuild && \
-    scl enable devtoolset-7 'cmake \
+    scl enable devtoolset-9 'cmake \
         -DCMAKE_BUILD_TYPE=Release \
         -DLLVM_LIBDIR_SUFFIX=64\
         -DLLVM_ENABLE_EH=1 \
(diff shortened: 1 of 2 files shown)


mthalman (approved):

am11:
cc @janvorli

am11:
Thank you. If there is a retention policy in effect for these containers, just so you know, runtime repo's master branch is now using this tag `centos-7-20201227183837-5fe0e50`.

directhex:
The unintended consequences from this are kicking my butt 😬

am11:
> The unintended consequences from this 

What are those?

directhex:
We're now linking any C++ code built in the CentOS 7 VM against libstdc++9, which is not ABI compatible with the libstdc++ 8 in any of our Ubuntu 18.04-based "cross" VMs. This results in a link-time error:

```
undefined reference to `std::_Sp_make_shared_tag::_S_eq(std::type_info const&)'
```

Note the missing _S_eq struct element between https://github.com/gcc-mirror/gcc/blob/releases/gcc-8/libstdc++-v3/include/bits/shared_ptr_base.h#L501-L513 and https://github.com/gcc-mirror/gcc/blob/releases/gcc-9/libstdc++-v3/include/bits/shared_ptr_base.h#L501-L515

So all runtime wasm builds have been broken since the weekend, when we updated the llvm-project yaml to build on centos-7-20210714125435-9b5bbc2 instead of centos-7-bfcd90a-20200121150017

I'm still trying to figure a workaround.

am11:
We haven't seen linkage problems with coreclr's C++ based libraries (this image is in use for about 11 months). Perhaps something to compare configs with.

However, if you want, you can safely revert a0b6b0a (back to devtoolset-7), because as of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/504, we are using debian-based gcc docker for gcc leg in runtime repo.

Curious, what is the benefit of upgrading from centos-7-bfcd90a-20200121150017 to centos-7-20210714125435-9b5bbc2?

directhex:
I've switched to an older docker tag for now.

Frankly, updating was just to keep llvm-project aligned with runtime.

PR #393, opened 2020-12-28
add openssl executable to Centos 8 helix image

wfurt:
at least one of the runtime tests depends on openssl app to determined ciphers openssl was compiled with. This change adds 1M in size. 

contributes to https://github.com/dotnet/runtime/issues/972


MattGal (approved):
LGTM, also tested locally (this is why I tend to leave in the huge list of dependencies when making new images :) )

PR #409, opened 2021-02-16
Adding V8 engine to webassembly container in order to run trimming tests

joperezr:
cc: @safern @mthalman @akoeplinger @MichaelSimons 

Contributes to https://github.com/dotnet/runtime/issues/39274

dotnet/runtime trimming tests are executing during build-time (not executed in Helix) and today we are not running them on the wasm vertical due to the fact that our webassembly build container doesn't have v8 installed in order to be able to execute webassmebly app bundles. This PR will add the v8 engine to the webassembly container, and a subsequent PR in dotnet/runtime will use this new container to add a wasm leg for our trimming tests.

### DIFF
--- a/src/ubuntu/18.04/webassembly/Dockerfile
+++ b/src/ubuntu/18.04/webassembly/Dockerfile
@@ -24,3 +24,13 @@ RUN git clone https://github.com/emscripten-core/emsdk.git ${EMSDK_PATH} \
     && ./emsdk install ${EMSCRIPTEN_VERSION}-upstream \
     && ./emsdk activate ${EMSCRIPTEN_VERSION}-upstream \
     && chmod -R 777 ${EMSDK_PATH}
+
+# Install V8 Engine to be able to execute Trimming Tests
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
+    && apt install -y nodejs \
+    && npm install jsvu -g \
+    && export PATH="${HOME}/.jsvu:${PATH}" \
+    && jsvu --os=linux64 --engines=javascriptcore,spidermonkey,v8 \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV PATH="/root/.jsvu:${PATH}"


safern (approved):
LGTM

src/ubuntu/18.04/webassembly/Dockerfile
```diff
@@ -24,3 +24,13 @@ RUN git clone https://github.com/emscripten-core/emsdk.git ${EMSDK_PATH} \
     && ./emsdk install ${EMSCRIPTEN_VERSION}-upstream \
     && ./emsdk activate ${EMSCRIPTEN_VERSION}-upstream \
     && chmod -R 777 ${EMSDK_PATH}
+
+# Install V8 Engine to be able to execute Trimming Tests
+RUN curl -sL https://deb.nodesource.com/setup_12.x | bash - \
+    && apt install -y nodejs \
```
mthalman, line 30: Installing nodejs shouldn't be necessary.  It's already installed in the base image at https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/f339b3b50c34cf1e4116ea7a94f25369debaee0c/src/ubuntu/18.04/coredeps/Dockerfile#L7

joperezr: True but for v8 engine we will need npm 10+ which will only be available after the previouse curl command. After running that, the `apt install nodejs` will install the new npm which will be able to install the v8 engine. With the nodejs that is preinstalled in the container we can't install v8.

mthalman (approved):

src/ubuntu/18.04/webassembly/Dockerfile
akoeplinger: @joperezr we don't need to install v8 from npm/jsvu, for the helix queue we just download the binary from blob storage: https://dev.azure.com/dnceng/internal/_git/dotnet-helix-machines?path=%2Fartifacts%2Flinux-v8-engine%2Flinux-v8-engine.sh&_a=contents&version=GBmaster

PR #415, opened 2021-02-23
Add Docker images for legacy Mono CI

directhex:
Legacy Mono has been kicked off Jenkins, so we need to use AzDO. Right now we're using existing Helix images and just have a step in our .yml to provision the extra packages we need on every build, but this is obviously a better technical solution.

The ARMv7 image is intended to allow us to run builds on our self-hosted ARM64 pool - according to the docs, correct use of LABEL allows a user-provided (distro-provided) Node binary, for cases where the VSTS-agent-provided Node binary might not work.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -664,7 +664,10 @@
               "os": "linux",
               "osVersion": "bionic",
               "tags": {
-                "ubuntu-18.04-helix-arm32v7-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {}
+                "ubuntu-18.04-helix-arm32v7-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "ubuntu-18.04-helix-arm32v7": {
+                    "isLocal": true
+                }
               },
               "variant": "v7"
             }
@@ -678,7 +681,10 @@
               "os": "linux",
               "osVersion": "bionic",
               "tags": {
-                "ubuntu-18.04-helix-arm64v8-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {}
+                "ubuntu-18.04-helix-arm64v8-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "ubuntu-18.04-helix-arm64v8": {
+                    "isLocal": true
+                }
               },
               "variant": "v8"
             }
--- a/manifest.json
+++ b/manifest.json
@@ -664,10 +664,7 @@
               "os": "linux",
               "osVersion": "bionic",
               "tags": {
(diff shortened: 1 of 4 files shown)


src/ubuntu/18.04/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+# Install additional Mono build dependencies
+RUN apt-get update \
+    && apt-get install -qq -y \
+        autoconf \
+        libtool \
+        automake \
+        build-essential \
+        gettext \
+        wget \
+        bc \
```
mthalman, line 14: Please list these in alphabetical order.  Same applies to the other Dockerfiles.

mthalman (approved):

manifest.json
```diff
@@ -664,7 +664,10 @@
               "os": "linux",
               "osVersion": "bionic",
               "tags": {
-                "ubuntu-18.04-helix-arm32v7-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {}
+                "ubuntu-18.04-helix-arm32v7-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "ubuntu-18.04-helix-arm32v7": {
+                    "isLocal": true
```
akoeplinger, line 669: this change seems unrelated?

src/ubuntu/18.04/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-helix-amd64
```
akoeplinger, line 1: I think we shouldn't derive from the Helix image since there's a lot of things we don't need from there in mono/mono, let's just copy whatever we need from there and inline here.

manifest.json
directhex: It's used for the FROM calls

src/ubuntu/18.04/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+        tzdata \
+        unzip \
+        wget \
+    && rm -rf /var/lib/apt/lists/* \
+    \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+
+ENV LANG en_US.utf8
+
+# create helixbot user and give rights to sudo without password
```
akoeplinger, line 32: let's use a different username since we won't be using this image for helix

akoeplinger (approved):

PR #418, opened 2021-02-24
Add a Docker image for Windows builds of Mono

directhex:
The docker image nightmares are made from.

As part of our "no more Jenkins" exercise, we need to build legacy Mono on AzDO. Some of our software requirements are... esoteric to say the least (for example, we package with WiX which requires .NET 3.5; the build system is Automake and we have adapted to Cygwin's quirks - starting over with MSys2 or another UNIX environment would mean basically starting over on Windows building, mid-servicing-cycle)

This Docker file roughly implements the build script from our old Jenkins cloud VM initialization. Some particularly gross hacks are present - likely with good reason.

I've been able to run a build locally with this image, but I haven't been able to test for adverse interactions between the image and the VSTS agent, so some iteration might need to happen later.


src/windows/2004/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,76 @@
+# escape=`
```
akoeplinger, line 1: what is this?

src/windows/2004/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+        Remove-Item -Recurse -Force %TEMP%\*
+
+# Download and install python
+RUN curl -SL --output %TEMP%\python2-x64.msi https://www.python.org/ftp/python/2.7.18/python-2.7.18.amd64.msi `
+    && powershell -Command `
+        Start-Process %TEMP%\python2-x64.msi -Wait -ArgumentList '/qn /norestart TARGETDIR=C:\python' ; `
+        Remove-Item -Recurse -Force %TEMP%\*
+
+# Download and install Mono
+RUN curl -SL --output %TEMP%\mono32.msi https://download.mono-project.com/archive/6.12.0/windows-installer/mono-6.12.0.107-gtksharp-2.12.45-win32-0.msi `
```
akoeplinger, line 19: we can remove this I think, it was used before we had a working monolite which is downloaded during the build. I've never actually seen it being used since we used to install an older Mono that wasn't compatible anyway.

src/windows/2004/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+USER ContainerAdministrator
+
+# Download and install cmake
+RUN curl -SL --output %TEMP%\cmake-x64.msi https://cmake.org/files/v3.9/cmake-3.9.6-win64-x64.msi `
+    && powershell -Command `
+        Start-Process %TEMP%\cmake-x64.msi -Wait -ArgumentList '/qn /norestart' ; `
+        Remove-Item -Recurse -Force %TEMP%\*
+
+# Download and install python
+RUN curl -SL --output %TEMP%\python2-x64.msi https://www.python.org/ftp/python/2.7.18/python-2.7.18.amd64.msi `
```
akoeplinger, line 13: do you remember what we're using this for? the Mono build requires python3 now and we're installing it via cygwin (since it's only used by the automake build afaik).

src/windows/2004/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+        Remove-Item -Recurse -Force %TEMP%\*
+
+# Download and install Mono
+RUN curl -SL --output %TEMP%\mono32.msi https://download.mono-project.com/archive/6.12.0/windows-installer/mono-6.12.0.107-gtksharp-2.12.45-win32-0.msi `
+    && powershell -Command `
+        Start-Process %TEMP%\mono32.msi -Wait -ArgumentList '/qn /norestart' ; `
+        Remove-Item -Recurse -Force %TEMP%\*
+
+# Download and install git
+RUN curl -SL --output %TEMP%\git-64bit.exe https://github.com/git-for-windows/git/releases/download/v2.30.1.windows.1/Git-2.30.1-64-bit.exe `
```
akoeplinger, line 25: I don't think we need this _inside_ the container since we'd be cloning on the AzDO host right?

src/windows/2004/mono/amd64/Dockerfile
directhex: No idea! I initially started with a Helix Server Core image, which used it. I assume it's responsible for changing the line break escape from backslash to backtick (given how problematic backslash is on Windows)

Let me check the docs.

https://docs.docker.com/engine/reference/builder/#escape

Yup!

src/windows/2004/mono/amd64/Dockerfile
directhex: I don't remember, no. I can yank it.

src/windows/2004/mono/amd64/Dockerfile
directhex: I think this was to provide `mcs` in absence of Monolite, back in the day

src/windows/2004/mono/amd64/Dockerfile
directhex: It's not entirely clear to me from reading the build logs whether git is used from the host side or container side - the "Initialize containers" step happens _before_ checkout

I know we use git during the build for version number generation & submodule sync, but I assume that's a second copy with Cygwin.

akoeplinger: yeah we shouldn't need this

src/windows/2004/mono/amd64/Dockerfile
```diff
@@ -1,76 +1,58 @@
...
+    && powershell -Command `
+        Start-Process %TEMP%\cygsetup.exe -Wait -ArgumentList '--quiet-mode --wait --root c:\cygwin --local-package-dir c:\cygwin --site http://mirrors.kernel.org/sourceware/cygwin/ --packages autoconf,automake,bison,gcc-core,gcc-g++,mingw64-i686-runtime,mingw64-i686-binutils,mingw64-i686-gcc-core,mingw64-i686-gcc-g++,mingw64-i686-pthreads,mingw64-i686-w32api,mingw64-i686-zlib,mingw64-x86_64-runtime,mingw64-x86_64-binutils,mingw64-x86_64-gcc-core,mingw64-x86_64-gcc-g++,mingw64-x86_64-pthreads,mingw64-x86_64-w32api,mingw64-x86_64-zlib,libtool,make,python,gettext-devel,gettext,intltool,libiconv,pkg-config,git,curl,wget,libxslt,bc,patch,cmake,libxml2-devel,libiconv-devel,zlib-devel,liblzma-devel,libssl-devel,bsdcpio,unzip,wget,p7zip,nohup,yasm' `
+    && C:\cygwin\bin\bash.exe -c 'for i in `/bin/find / -xdev -type l`; do j=`/bin/readlink $i`; /bin/rm -f $i; /bin/ln -sf $j $i; done' `
+    && powershell -Command `
+        Remove-Item -Recurse -Force c:\cygwin\http* ; `
+        Remove-Item -Recurse -Force %TEMP%\*
+
+# Setup PATH
+RUN setx /M PATH "%PATH%;C:\Program Files\CMake\bin;C:\python;C:\cygwin\bin"
```
akoeplinger, line 56: we no longer install python:

```suggestion
RUN setx /M PATH "%PATH%;C:\Program Files\CMake\bin;C:\cygwin\bin"
```

akoeplinger (approved):

src/windows/2004/mono/amd64/Dockerfile
MichaelSimons: This is the relevant snippet from the doc @directhex references

> Setting the escape character to ` is especially useful on Windows, where \ is the directory path separator.

I have seen a number of people get stumped by Docker build failures as a result of the default escape character when building Windows Dockerfiles.  I recommend to everyone to change the escape character in Windows Dockerfiles.

directhex: You have no idea what nightmares I had with AOT cross compiler builds on Windows caused by an escaped *space* in an msbuild exec call

directhex:
HOW SLOW are the machines doing these image builds, slash, how fast is Ryzen 5000 on my personal PC? Because _3 hours in_ it's still not past step 5

directhex:
clearly it stalled during the .net 3 install, but i can't repro that locally 🤔

src/windows/2004/mono/amd64/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# escape=`
+FROM mcr.microsoft.com/windows:2004
```
mthalman, line 2: You've confirmed that `mcr.microsoft.com/windows/servercore:2004` doesn't meet your needs?  It's smaller and preferable if it works.

directhex: VS Build Tools fails to install with an inscrutable error in hcsshim::ImportLayer when I use servercore as the base

directhex:
And now it's continually failing due to disk space.

Hrm

Wonder if I can figure out how to make the VC build tools work on server core...

mthalman:
Let me kill this build agent and get a fresh one.

mthalman:
Still having issues in the build:

```
Step 4/9 : RUN curl -SL --output %TEMP%\vcbuildtools.exe https://go.microsoft.com/fwlink/?LinkId=691126     && powershell -Command         Start-Process %TEMP%\vcbuildtools.exe -Wait -ArgumentList '/full /quiet /norestart' ;         Remove-Item -Recurse -Force %TEMP%\*
 ---> Running in 1b43977eeb1e
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 3210k  100 3210k    0     0  3210k      0  0:00:01 --:--:--  0:00:01 25.0M
re-exec error: exit status 1: output: time="2021-02-25T17:33:24Z" level=error msg="hcsshim::ImportLayer - failed failed in Win32: The system cannot find the path specified. (0x3)" error="hcsshim::ImportLayer - failed failed in Win32: The system cannot find the path specified. (0x3)" importFolderPath="C:\\ProgramData\\docker\\tmp\\hcs889691927" path="\\\\?\\C:\\ProgramData\\docker\\windowsfilter\\77269554e333d8b4fb20b3e1b7749ac6216aa5e89363085f794d671d5a31b69c"
hcsshim::ImportLayer - failed failed in Win32: The system cannot find the path specified. (0x3)
```

directhex:
Fascinating. That's the issue I get with server core, but not with regular server, locally

One thing I've determined is the build tools install is actually _successful_ - the error is because some file being created during the install is incompatible with Docker's overlay file system handling (I had the same issue with Cygwin, which was creating NTFS junction points). I'll keep investigating.

akoeplinger:
I've seen a few reports that this also happens with 1) running out of disk space in the container, which you can supposedly workaround by passing `--storage-opt size=120G` or 2) some bugs on older Windows releases.

Can we try the first one? We might also be able to switch to the server core image if it turns out that's the issue.

akoeplinger:
@directhex interestingly the build tools install worked in the previous build where we hit the .NETFx3 issue: https://dev.azure.com/dnceng/public/_build/results?buildId=1010896&view=logs&j=a798c2b2-96b5-5d1e-2090-066e854eb5f7&t=7e5bed0e-2df7-5275-7456-302bac5aaf6e&l=345

directhex:
@akoeplinger I'm trying to figure out how to get usable information out of Docker (there's a 6 year old bug still open that they forgot to implement log messages w/ `debug` priority). The problem between VC build tools installing fine on regular Windows (locally) but not on Server Core (locally) is a problem when docker tries to pack up the changes in that RUN into a layer. There are no new junction points created, so it's something else

directhex:
So here's the weirdest difference between my Server Core & regular builds: CRT_X64_Store. That package doesn't get installed on the bad Server Core run. The 32-bit package is, though. 🤷‍♂️

directhex:
An important thing in engineering is knowing when to throw in the towel and stop throwing good time after bad. That time is now. I'll find another way to implement this, avoiding Docker.

PR #437, opened 2021-04-09
Change ubuntu cross arm alpine to use 3.13 rootfs

janvorli:
Due to the breaking change on Alpine 3.13 w.r.t. time_t type
size, we will only support ARM on Alpine >= 3.13.
This change updates the ROOTFS that the cross build uses
to be the one of Alpine 3.13.

I have also created cross image for Ubuntu 16.04, as we should
be cross building on 16.04 rather than 18.04 to get better coverage
of Linux distros where cross-tools like crossgen can be run.
Ideally, we would use CentOS 7, but we cannot run crossbuild on
that one.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -312,6 +312,18 @@
         {
           "platforms": [
             {
+              "dockerfile": "src/ubuntu/16.04/cross/arm-alpine",
+              "os": "linux",
+              "osVersion": "xenial",
+              "tags": {
+                "ubuntu-16.04-cross-arm-alpine-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {}
+              }
+            }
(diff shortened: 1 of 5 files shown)


mthalman:
You'll need to update the manifest.json file to register the new Dockerfile being added.  Basically, you just need to copy this snippet and adjust the values accordingly: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/5d87b804d306ea13a42b6bd02cd4841d7d2eff90/manifest.json#L300-L311

janvorli:
Thanks @mthalman,  I've forgotten to add that file to my commit, I have made that change locally.

mthalman (approved):

PR #447, opened 2021-05-11
mlnet cross targeting images

michaelgsharp:
Cross targeting images for ML.Net arm and arm64.


src/ubuntu/18.04/mlnet/cross/arm64/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+ADD rootfs.arm64.tar crossrootfs
+
+# Install openmp support with Clang
+RUN apt-get update \
+    && apt-get install -y \
+        clang-3.9 \
+        libomp5 \
+        libomp-dev
```
mthalman, line 18: Any reason this can't be combined with the `apt-get install` from above?

mthalman (approved):

PR #463, opened 2021-06-11
[win/ltsc2019/webassembly] Reduce the image size

radekdoulik:
The CI build is failing with:

    System.IO.IOException: There is not enough space on the disk. : 'C:\__w\1\s\artifacts\log\Release\Build.binlog'

Lets try to save some space. It looks like
the `Microsoft.VisualStudio.Component.VC.Tools.x86.x64` component pulls
enough dependencies for us. I was also able to build it without
`Microsoft.VisualStudio.Workload.ManagedDesktopBuildTools` component
locally. Let see if it will be enough. The size is reduced by cca 1.3GB.

### DIFF
--- a/src/windowsservercore/ltsc2019/webassembly/Dockerfile
+++ b/src/windowsservercore/ltsc2019/webassembly/Dockerfile
@@ -81,3 +81,5 @@ RUN curl -SL --output vs_buildtools.exe https://aka.ms/vs/16/release/vs_buildtoo
     --add Microsoft.VisualStudio.Component.Windows10SDK.17763 \
 || IF "%ERRORLEVEL%"=="3010" EXIT 0) \
     && del /q vs_buildtools.exe
+
+RUN del /q /f /s %TEMP%\*
--- a/src/windowsservercore/ltsc2019/webassembly/Dockerfile
+++ b/src/windowsservercore/ltsc2019/webassembly/Dockerfile
@@ -45,21 +45,21 @@ ENV EMSDK_PATH="C:\emscripten\emsdk"
 
 RUN mkdir %EMSCRIPTEN_PATH% \
     && cd %EMSCRIPTEN_PATH% \
-    && git clone https://github.com/emscripten-core/emsdk.git %EMSDK_PATH%
-RUN cd %EMSDK_PATH% \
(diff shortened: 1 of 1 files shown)


mthalman:
One other thing I realized is that I think VS install files/folders are leftover in %TEMP%.  Would probably save some space cleaning that stuff up too.

src/windowsservercore/ltsc2019/webassembly/Dockerfile
```diff
@@ -81,3 +81,5 @@ RUN curl -SL --output vs_buildtools.exe https://aka.ms/vs/16/release/vs_buildtoo
     --add Microsoft.VisualStudio.Component.Windows10SDK.17763 \
 || IF "%ERRORLEVEL%"=="3010" EXIT 0) \
     && del /q vs_buildtools.exe
+
+RUN del /q /f /s %TEMP%\*
```
mthalman, line 85: Rather than having this in a separate RUN instruction, add this as part of the previous RUN instruction after deleting vs_buildtools.exe.  Doing that will save on overall image size.  By having a separate RUN instruction that deletes from TEMP it means the previous image layer kept the complete contents of TEMP and actually carries that with the image. So that's not a best practice.

src/windowsservercore/ltsc2019/webassembly/Dockerfile
```diff
@@ -81,3 +81,5 @@ RUN curl -SL --output vs_buildtools.exe https://aka.ms/vs/16/release/vs_buildtoo
     --add Microsoft.VisualStudio.Component.Windows10SDK.17763 \
 || IF "%ERRORLEVEL%"=="3010" EXIT 0) \
```
mthalman, line 82: Indent this to align with everything else

src/windowsservercore/ltsc2019/webassembly/Dockerfile
radekdoulik: I wonder whether it would make sense to squash the image in the end?

radekdoulik: Ah, the squash option is experimental and requires docker daemon with experimental features enabled. So I guess it doesn't apply here.

mthalman (approved):

PR #481, opened 2021-07-19
[dotnet\runtime WASM] Added typescript

Daniel-Genkin:
Added global typescript to the WebAssembly dockerfile.

### DIFF
--- a/src/ubuntu/18.04/webassembly/Dockerfile
+++ b/src/ubuntu/18.04/webassembly/Dockerfile
@@ -19,6 +19,9 @@ RUN wget https://cmake.org/files/v3.17/cmake-3.17.0-Linux-x86_64.tar.gz \
 # WebAssembly build needs working UTF-8 locale
 RUN locale-gen en_US.UTF-8
 
+# WebAssembly build needs typescript
+RUN npm i -g typescript
+
 # Install Emscripten toolchain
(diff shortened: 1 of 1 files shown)


src/ubuntu/18.04/webassembly/Dockerfile
```diff
@@ -19,6 +19,9 @@ RUN wget https://cmake.org/files/v3.17/cmake-3.17.0-Linux-x86_64.tar.gz \
 # WebAssembly build needs working UTF-8 locale
 RUN locale-gen en_US.UTF-8
 
+# WebAssembly build needs typescript
+RUN npm i -g typescript
```
radekdoulik, line 23: Should we also add `/usr/local/lib/node_modules/typescript/bin` to the PATH?

Daniel-Genkin: When I tested locally this was enough. I think the way npm works is it just adds a shortcut into the npm path folder to the various installed packages. So effectively it adds it automatically if I understand correctly

radekdoulik: I checked and it indeed adds the link to `/usr/local/bin` so we should be fine.

radekdoulik (approved):

mthalman (approved):

PR #485, opened 2021-07-26
Add the lldb python support to Alpine 3.12/3.13

mikem8361:
Needed to run the diagnostics SOS tests. Adding lldb-dev only allowed it to be built not run.

/cc: @hoyosjs

### DIFF
--- a/src/alpine/3.12/helix/amd64/Dockerfile
+++ b/src/alpine/3.12/helix/amd64/Dockerfile
@@ -24,12 +24,14 @@ RUN apk update && \
         libffi \
         libffi-dev \
         linux-headers \
+        lldb \
         lldb-dev \
         llvm \
         make \
         openssl \
         openssl-dev \
         py-cffi \
+        py3-lldb \
         python3 \
         python3-dev \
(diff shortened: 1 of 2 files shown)


src/alpine/3.12/helix/amd64/Dockerfile
```diff
@@ -24,12 +24,14 @@ RUN apk update && \
         libffi \
         libffi-dev \
         linux-headers \
+        lldb \
```
mthalman, line 27: This is redundant since `lldb-dev` is already listed and `lldb` is a dependency of `lldb-dev`.  Same applies to the Alpine 3.13 Dockerfile.

mikem8361: I know it was redundant but I wanted to be clear that lldb itself is needed to run tests instead of just for building.

mikem8361: Is leaving it ok with you?

mthalman (approved):

PR #488, opened 2021-07-28
small fixes for Fedora34 helix image

wfurt:
1) add openssl binary. Seems to be missing now from default and some tests depend on it.
2) execute msquic build directly. The top wrapper would download extra copy of dotnet to start Arcade and then build it twice - once for x64 and once for x86 - and that is not needed.


mthalman (approved):

mthalman:
Published tags:
`mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-34-20210728124700-5883dc8`
`mcr.microsoft.com/dotnet-buildtools/prereqs:fedora-34-helix-20210728124700-4f64125`

PR #490, opened 2021-07-31
Add support for alpine 3.14

am11:


src/alpine/3.14/WithNode/amd64/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.14 as base
+FROM node:lts-alpine
+COPY --from=base / /
+RUN apk update && apk add --no-cache --virtual yarn
```
am11, line 4: @safern, it has been a while (2018) since the previous setup that was done for WithNode to build node.js 10 from source. For this new image, I have used LTS version from node.js official alpine image in multistage, since multistage is available since docker 17. It cuts down the CI time by 35+ minutes. Let me know if there is reason to keep using the old approach for 3.14 as well.

src/alpine/3.14/amd64/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        util-linux-dev \
+        which \
+        zlib-dev
+
+RUN apk -X http://dl-cdn.alpinelinux.org/alpine/v3.10/main add --no-cache \
+        cmake && \
+    apk -X http://dl-cdn.alpinelinux.org/alpine/edge/main add --no-cache \
+        userspace-rcu-dev \
+        lttng-ust-dev \
+        numactl-dev
```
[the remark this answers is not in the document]
am11: Agreed, also userspace-rcu-dev is a dependency of lttng-ust-dev. I have double checked the packages at https://pkgs.alpinelinux.org/packages?branch=v3.14 and updated the files so we run a single apk add command.

src/alpine/3.14/WithNode/amd64/Dockerfile
am11: On a second thought, the node.js from alpine's main repo is also the latest LTS: 14.17 https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.14. I have simplified it to install node.js and yarn from packages.

mthalman (approved):

src/alpine/3.14/WithNode/amd64/Dockerfile
wfurt: Do we know where that image is used? It feels like it would be better to use package from the eco-system as that what others would likely to do.

am11: It is used for label `com.azure.dev.pipelines.agent.handler.node.path` which is documented here: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops#tell-the-agent-about-nodejs

In 2018 node 10 was the latest LTS, today LTS is 14, which is what I have used.

> it would be better to use package from the eco-system

The distro package (from main) is 14.17, I have just used the one from Alpine packages (like everything else in this container). I think "official" in the context of a distro is distro package, which is generally recommended way to install software in their respective ecosystems.

mthalman:
Is this waiting for further reviewers to weigh in?

wfurt (approved):
LGTM

PR #499, opened 2021-09-04
Add missing tzdata in alpine 3.13 and 3.14 amd64

am11:
@ViktorHofer, this one got missed in the refactoring, which is why the ICU test was failing to find `Europe/London` timezone (amd64-only issue).

### DIFF
--- a/src/alpine/3.13/WithNode/amd64/Dockerfile
+++ b/src/alpine/3.13/WithNode/amd64/Dockerfile
@@ -1,34 +1,9 @@
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.13
-RUN apk update
 
-ENV NODE_VERSION=14.16.1
-
-### INSTALL NODE
-RUN apk add --no-cache nodejs
-
-ENV YARN_VERSION=1.10.1
-
-RUN apk add --no-cache --virtual .build-deps-yarn \
-        gnupg \
-        tar \
-        && for key in \
-            6A010C5166006599AA17F08146C2130DFD2497F5 \
-        ; do \
-            gpg --keyserver hkps://keyserver.ubuntu.com --recv-keys "$key" \
-            || gpg --keyserver hkps://keys.openpgp.org --recv-keys "$key"; \
-        done \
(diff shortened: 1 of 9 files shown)


am11:
In second commit, cleaned up some sub-dependencies which we don't (directly) need and unified the files so they "look" a like (would make it easier in future to deduplicate this code when we switch to templates like https://github.com/dotnet/dotnet-docker/blob/44483abdb1bd260048d8efc6eb56bc56b66e8ea2/eng/dockerfile-templates/runtime/6.0/Dockerfile.linux).

src/alpine/3.13/WithNode/amd64/Dockerfile
```diff
@@ -1,34 +1,9 @@
...
-            && curl -fsSLO --compressed "https://yarnpkg.com/downloads/$YARN_VERSION/yarn-v$YARN_VERSION.tar.gz" \
-            && curl -fsSLO --compressed "https://yarnpkg.com/downloads/$YARN_VERSION/yarn-v$YARN_VERSION.tar.gz.asc" \
-            && gpg --batch --verify yarn-v$YARN_VERSION.tar.gz.asc yarn-v$YARN_VERSION.tar.gz \
-            && mkdir -p /opt \
-            && tar -xzf yarn-v$YARN_VERSION.tar.gz -C /opt/ \
-            && ln -s /opt/yarn-v$YARN_VERSION/bin/yarn /usr/local/bin/yarn \
-            && ln -s /opt/yarn-v$YARN_VERSION/bin/yarnpkg /usr/local/bin/yarnpkg \
-            && rm yarn-v$YARN_VERSION.tar.gz.asc yarn-v$YARN_VERSION.tar.gz \
-            && apk del .build-deps-yarn
+RUN apk update && apk add --no-cache nodejs yarn
```
mthalman, line 3: Recommended to list out each package name on its own line.

mthalman (approved):

src/alpine/3.13/WithNode/amd64/Dockerfile
am11: Updated.

am11:
@ViktorHofer, if there is no other feedback could this be merged? I'll update runtime PR to fix the culture failing tests.

PR #501, opened 2021-09-08
Downgrade to node.js v10.24.1 for ExtractFiles task

am11:
This is a workaround for https://github.com/microsoft/azure-pipelines-tasks/issues/15261.

cc @ViktorHofer


am11:
cc @wfurt, @safern, learned it the hard way that _this_ was the reason to keep node.js to v10 rather than the latest LTS .. 😁

mthalman (approved):
Thanks for documenting in the Dockerfile why this pinned version is needed.

am11:
Good to merge? I'll update runtime PR once this is merged. :)

mthalman (approved):

PR #503, opened 2021-09-09
Use node.js from a shared location

am11:
Upon consuming the new images, it turned out `helixbot_azpcontainer` (the user created during `Initialize containers` step) does not have access to `/root`, although the logs [do suggest](https://dev.azure.com/dnceng/public/_build/results?buildId=1353034&view=logs&j=174348cd-3455-59d8-c7f7-32c969b0807d&t=8a37b7b6-cca0-42a9-93a4-6c86eb692846&l=45) as if it has same access as root..

Simpler way is to just download the version we need in `/usr/share/node/` directory, and symlink the binary at `/usr/local/bin/node`.

### DIFF
--- a/src/alpine/3.13/WithNode/amd64/Dockerfile
+++ b/src/alpine/3.13/WithNode/amd64/Dockerfile
@@ -4,18 +4,19 @@ RUN apk update && \
     apk add --no-cache \
         yarn
 
-# We need to pin node.js to v10.24.1, because ExtractFiles AzDO task is currently failing with ENOBUFS
-# see https://github.com/microsoft/azure-pipelines-tasks/issues/15261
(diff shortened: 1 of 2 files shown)


mthalman (approved):

am11:
`/usr/bin/node` was already occupied by node.js v14 (LTS), which is now getting installed as dependency of `yarn`. I have used `/usr/local/bin/node` for symlink, as `/usr/local/bin` has precedence over `/usr/bin` in PATH.

am11:
Please merge, it will unblock runtime PR, hopefully this time fr. @ViktorHofer 🤣

ViktorHofer (approved):

PR #504, opened 2021-09-10
Add a separate image for gcc 11

am11:
GNU compiler toolchain is used in runtime for native build validation. The LLVM toolchain is default one to build the shipping product.

Currently we are using gcc9 from `centos-7` image, which we install from YUM packages when building the prereq image for two purposes:
1. to compile custom llvm toolchain
2. to use it in for the said validation leg in runtime repo

Therefore, it is non-trivial to upgrade gcc independently, as first we need to wait fro `devtoolset` to pick up the GNU version, then we need to make sure our custom llvm toolchain's build is intact.

This PR adds gcc 11 leg based on official library layer (https://hub.docker.com/_/gcc), which uses Debian 11 (buillseye) for gcc 11. In future, we can add `src/debian/{version}/gcc/Dockerfile` when needed, without affecting anything else.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1127,6 +1127,19 @@
           "platforms": [
             {
               "architecture": "amd64",
+              "dockerfile": "src/debian/11/gcc11",
+              "os": "linux",
+              "osVersion": "bullseye",
+              "tags": {
+                "debian-11-gcc11-amd64-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {}
+              }
+            }
+          ]
+        },
+        {
+          "platforms": [
+            {
+              "architecture": "amd64",
               "dockerfile": "src/debian/11/helix/amd64",
               "os": "linux",
               "osVersion": "bullseye",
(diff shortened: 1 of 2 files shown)


am11:
cc @janvorli, @hoyosjs, @wfurt
I have tested it locally, will send a PR in runtime to switch gcc leg to this container once we have the new image produced.

am11:
Failures are in unrelated legs (Ubuntu certificates are expired)

mthalman (approved):

am11:
Rebased with main.
Ubuntu error went away, since it was fixed by https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/505.
Speaking of _moving targets_, between yesterday and today, Fedora 34 regressed due to pip marked as optional dependency... opened #506 with the fix.

am11:
All green. Lets snapshot it now before it regresses. 😅

@hoyosjs, @janvorli, this will potentially fix all 2565 warnings currently we are hitting in gcc leg, enable us to unsuppress `-Wall` in the JIT and hopefully will close https://github.com/dotnet/runtime/issues/33541.

hoyosjs:
LGTM if the runtime properly builds on this container. Thanks @am11!

am11:
Thanks! @mthalman, if it LGTYT, please merge.

PR #507, opened 2021-09-13
Use Alpine 3.13 (now default) for cross-build image

am11:
cross-build of arm32 is currently using 3.13 and arm64 is using 3.9. We have updated rootfs script to set 3.13 as a minimum version https://github.com/dotnet/arcade/pull/7849. This PR aims to rebuild both arm and arm64 cross-build Alpine images, so 3.13 is used consistently for alpine on all supported architectures.

### DIFF
--- a/src/ubuntu/16.04/cross/arm-alpine/hooks/pre-build
+++ b/src/ubuntu/16.04/cross/arm-alpine/hooks/pre-build
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../build-scripts/build-rootfs.sh ubuntu-16.04 alpine3.13 arm lldb3.9
+$SCRIPTPATH/../../../../build-scripts/build-rootfs.sh ubuntu-16.04 alpine arm lldb3.9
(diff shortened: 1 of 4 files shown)


am11:
@mthalman since it is the change in upstream (arcade) repo which we want to exercise, but arm64 will use the cached image. 
Is there a way to force rebuild arm64 images as well part of CI run?

mthalman:
Because the Dockerfile hasn't changed, it'll end up pulling the already published version of the built Dockerfile instead of building it (and running the shell script).  To actually build it, you could make a simple whitespace change to the relevant Dockerfiles to verify the build results and then revert the change before merging.

am11:
Ah ok, for now I have toggled the values (since alpine == alpine3.13 as of now). We can align them in next cycle.

am11:
@mthalman, unfortunately the new image picked up the layer from cache, arm64 is still on 3.9 .. 😞

```sh
$ URN=mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-16.04-cross-arm64-alpine-20210923132506-78f7860
$ docker run -it  $URN cat /crossrootfs/arm64/etc/os-release | grep VERSION
VERSION_ID=3.9.6
```

We would need to add a dummy step (`RUN echo foo`) somewhere on top of these Dockerfiles to force the rebuild (whitespace changes don't count) ..

mthalman:
Oh, right, forgot about that.  I can run the build with caching disabled.  No worries.

am11:
Thanks a lot! It is fixed in `ubuntu-16.04-cross-arm64-alpine-20210923140502-78f7860`.
arm and arm64 files are aligned in #517 to avoid confusion.

PR #546, opened 2021-12-07
Eliminate OS-specific pipelines

### LINKED ISSUE #535: Proposal: Eliminate OS-specific pipelines
There are currently 15 pipeline pairs defined in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/tree/main/eng/pipelines; one pair for each OS. As more Dockerfiles for new OS versions are added, these pipelines need to updated; similarly when the Dockerfiles for a given OS version are removed. There's just a lot of maintenance involved here.

Originally, there was only one pipeline which built all the Dockerfiles. This became problematic because many of the build legs took very long (several hours). So when making a change to a Dockerfile that was built by a build leg that finished quickly, you'd still have to wait hours for the rest of the build to be successful. By defining OS-specific pipelines, it split apart the building of Dockerfiles so that a change to an Alpine Dockerfile would only trigger the Alpine-based pipeline to run and not end up building the long-running CentOS Dockerfiles, for example.

All this was implemented prior to having the image caching feature, however. Now that we have image caching, the impact of using a single pipeline to build everything is less because it wouldn't actually have to build Dockerfiles which don't need to be built.

Doing this would also negate the proposed work for https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/357.

mthalman:
This deletes all the OS-specific pipelines and just keeps two left:
* dotnet-buildtools-prereqs: for official builds
* dotnet-buildtools-prereqs-pr: for PR builds

This also updates the pipeline to remove the test placeholder which was a no-op. This results in less build jobs being run because it skips the PreBuildValidation job.

Fixes #535


mthalman:
Closing this in favor of a different approach that doesn't require as many build jobs by default.

PR #547, opened 2021-12-07
Add an Ubuntu 20.04 image for building against Raspbian 10

directhex:
Raspberry Pi is an extremely popular ecosystem of single board computers based on SoCs from Broadcom. Whilst the latest generation of hardware has a reasonable ARMv8 multi-core CPU, a major selling point of the Pi is that the same OS image and apps work on all hardware - which means building for the lowest common denominator (ARMv6, VFPv2, hard float). It isn't possible to simply use the existing ARMv7 crossrootfs and alter the compiler flags - something in the environment is baked in and results in SIGILL and SIGSEGV on actual Pi hardware.

This image is based on Ubuntu 20.04 (rather than our common Ubuntu 18.04 base for cross builds) because Clang in Ubuntu 18.04 has a bug which specifically affects us - An error introduced in 9.0.0 and fixed in 9.0.1 causes the Clang codegen to treat VFPv2 as having 32 general-purpose registers instead of the correct count of 16. The bug causes a mix of SIGILL and crashes in seemingly impossible places. Using Ubuntu 20.04 as a base instead was deemed the best long-term solution to the availability of a newer version of Clang.


directhex:
Ref https://github.com/dotnet/runtime/issues/7764

mthalman (approved):

Issue #357, opened 2020-08-18
PR validation shouldn't have to rebuild all Dockerfiles when a new image is added


MichaelSimons:
Each time a PR is made to add a new image, the manifest has to be updated which triggers all of the Dockerfiles in the repo to run through PR validation.  This wastes a fair amount of compute as there are a lot of Dockerfiles in this repo and some take a very long time to build (e.g. ubuntu cross).  Ideally PR validation should only have to build a small subset of the repo.

Today there are separate pipelines defined for each distro/version as seen [here](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/tree/master/eng/pipelines).  These pipelines build the images specific to the distro/version the pipeline is scoped to.  The problem is the [dotnet-buildtools-prereqs-docker pipeline](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/eng/pipelines/dotnet-buildtools-prereqs-eng.yml).  This gets triggered whenever there is a change to things like the engineering infra and the manifest.

To solve this problem, I speculate it is possible to break apart the single manifest into parts that are scoped to distro/version associated with each pipeline.  The manfiest files would then reside within the distro/version folder.  The pr and build triggers would then work as desired mainly the entire set of repo Dockerfiles would not run through PR validation when a new image is added.  Additionally the current behavior for building/testing only the Dockerfiles that are edited would be maintained.

MichaelSimons:
In theory we believe this work but needs some testing/investigation to prove it out.

PR #552, opened 2021-12-10
Consolidate pipelines into a per-OS grouping

mthalman:
This consolidates the pipelines to be grouped on a per-OS basis rather than an OS-version basis. This pattern was already done for Alpine Linux but now it is applied to all operating systems. In addition, it also splits apart the manifest into separate per-OS sub-manifests.

This has the benefit of requiring less maintenance since new versions of an OS will be handled by the existing pipeline. And it also limits the amount of build jobs by not having to build _everything_ in the repo.

Closes https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/535
Closes https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/357


MichaelSimons (commented):
Looks great.  I like the ease in maintenance this creates.  Couple suggestions:

1. The readme could use a couple updates to provide clarify around the manifest pattern as well as calling out that when a new distro is added, pipelines will need to be created.
2. Consider posting an announcement for this change and cc the top contributors - Matt, Jan, Jo, Tomas, etc.

MichaelSimons (approved):

PR #562, opened 2022-01-17
Add windowsservercore-20H2-helix-webassembly image

radekdoulik:
Also bump git and node versions


mthalman:
20H2 is going to be EOL in August. Do we really need new dependencies on it?

radekdoulik:
> 20H2 is going to be EOL in August. Do we really need new dependencies on it?

I need something as replacement for removed 2004 webassembly image. So I based the new one on 20H2, which already has the amd64 image.

Thinking more about it I can try to prepare LTSC2022 version.

radekdoulik:
Replaced by #563

PR #563, opened 2022-01-18
Add windowsservercore-ltsc2022-helix images

radekdoulik:
2 images for amd64 and webassembly


radekdoulik:
@mthalman do you think I should also remove 20h2 amd64 image?

mthalman:
> @mthalman do you think I should also remove 20h2 amd64 image?

I don't think that's necessary at this point.

mthalman (approved):

PR #570, opened 2022-02-01
Update to use 1ES hosted pool for Linux amd64

mthalman:
cc @mmitche


mmitche (approved):

MichaelSimons:
Why is this change needed versus pulling in the latest common templates - https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/567?

mthalman:
> Why is this change needed versus pulling in the latest common templates - #567?

Ah, you're right. I was thinking that it was using the base, not the dotnet version, of the stages template. So yes, I'll close this and defer to the changes in #567.

PR #571, opened 2022-02-10
Enable IBM POWER platform

directhex:
64-bit little endian Linux only, at this point in time.

### DIFF
--- a/src/ubuntu/18.04/coredeps/Dockerfile
+++ b/src/ubuntu/18.04/coredeps/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/powershell:lts-ubuntu-18.04	
+FROM mcr.microsoft.com/powershell:lts-ubuntu-18.04
 
 # Install tools used by the VSO build automation.
 RUN apt-get update \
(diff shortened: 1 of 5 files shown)


mthalman:
> Why did that fail? Why would apt-add-repository work during the crossdeps image build, but the key not work for the apt update in the cross image?

Because the base image was never built. The infrastructure got a cache hit for it meaning that it saw that its base image is up-to-date and the Dockerfile hasn't changed since it was last published. So it's using what is actually a very old base image. It all stems from https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/c9b3b3d562b88d9f907b8432ae83080c23d77707/src/ubuntu/18.04/coredeps/Dockerfile#L1

That tag hasn't been updated in 2 years. I would suggest updating it to something a bit more modern.

src/ubuntu/18.04/coredeps/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/powershell:6.2.1-ubuntu-18.04
+FROM mcr.microsoft.com/powershell:lts-ubuntu-18.04
```
mthalman, line 1: nit: Remove extra whitespace at the end

mthalman (approved):

PR #575, opened 2022-02-15
Increase Ubuntu build timeout

mthalman:
The previous build timeout update in #573 wasn't sufficient. The build is still timing out after 3 hrs. Increasing to 4 hrs.


directhex:
Eventually we're gonna find the magic number which is sufficient for a wholesale rebuild of all ubuntu-coredeps' dependencies. Maybe this one is it!

PR #578, opened 2022-02-23
Take the timeout up to five hours

directhex:
This whole adventure has highlighted a fragility in how we deal with changes lower down the stack which impact many many images requiring many many rebuilds. Obviously on a normal day, a five hour timeout is way too long and indicates a problem, but occasionally it's needed. Is there a better way to codify that?

As discussed on Teams, the build needs 3h50 to complete, but needs more than 10 minutes to upload the results. So with this, we're giving a nice chunky 1h10 for the upload. Plenty of time. Let's finally get the repo green again.

### DIFF
--- a/eng/pipelines/dotnet-buildtools-prereqs-ubuntu-pr.yml
+++ b/eng/pipelines/dotnet-buildtools-prereqs-ubuntu-pr.yml
@@ -24,6 +24,6 @@ stages:
   parameters:
     internalProjectName: ${{ variables.internalProjectName }}
     publicProjectName: ${{ variables.publicProjectName }}
-    linuxAmdBuildJobTimeout: 240
+    linuxAmdBuildJobTimeout: 300
     customBuildInitSteps:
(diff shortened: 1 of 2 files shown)


mthalman (approved):

mthalman:
> This whole adventure has highlighted a fragility in how we deal with changes lower down the stack which impact many many images requiring many many rebuilds. Obviously on a normal day, a five hour timeout is way too long and indicates a problem, but occasionally it's needed. Is there a better way to codify that?

This is all stems from having a common root Dockerfile from which the others are based on. Because of that the infrastructure lumps all of them into one build job. We have discussed in the past supporting the breaking apart of such large graphs into separate build jobs. This would allow the root Dockerfile to be built in one job and then other branches of the graph that all depend on that build job could be split into their own separate build jobs. This hasn't been something we've seriously considered to this point because the ROI isn't worth it. I think we'll just have to wait and see how things look with this and go from there.

PR #600, opened 2022-04-13
new attempt to build Fedora35

wfurt:
with Fedora 36  just few days away this does not make sense.
This is just probe to see  if new OSes are buildable again.  replaces  #509


wfurt:
This was mostly check if we can avoid now failures we saw in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/509. I will eventually work out new Helix image with MsQuic.

PR #601, opened 2022-04-13
Add RHEL 8 Dockerfile for source-build.

crummel:


mthalman:
@crummel - This PR is marked as draft and has been around a while. Is any further work going to happen here or can it be closed?

crummel:
This can be closed, we're using CentOS as our family representative.

PR #602, opened 2022-04-14
Build Fedora 36 with msquic

wfurt:
This may not be final but should be enough to give it try with runtime. 
It seems like the infrastructure issues blocking upgrade to 35 are resolved so I decided to jump to 36 instead (as it should be release in few days) 

This is significantly different from the 34 with regard to msquic. Back then there were no published package and everything was in flux. For now I'm using already published binaries. 

- 2.x is current normal. Build and published by MsQuic team. There is still some problem with signature so for now it uses nogpgcheck. This will be used by test for main/7.0
- 2.0 is incompatible with .NET 6 so I also bring the 1.9 so we can test side-by-side installations and have forward version for release/6.0

I still use `dnf` to find best compatible version e.g. it should pick up 1.9.1 and other security fixes. But I use rpm directly so I can have both versions at the same time

```
[helixbot@a11a8df5b513 lib64]$ rpm -qa| grep quic
libmsquic-2.0.2-1.x86_64
libmsquic-1.9.0-1.x86_64
```

and 
```
helixbot@a11a8df5b513 lib64]$ ls -al libmsquic.*
-rwxr-xr-x 1 root root 2278504 Oct 22 19:17 libmsquic.lttng.so
-rw-r--r-- 1 root root 9532512 Apr  4 17:28 libmsquic.lttng.so.2.0.2
-rwxr-xr-x 1 root root 1539632 Oct 22 19:18 libmsquic.so
-rw-r--r-- 1 root root 6508088 Apr  4 17:28 libmsquic.so.2
-rw-r--r-- 1 root root 6508088 Apr  4 17:28 libmsquic.so.2.0.2
```

cc: @omajid @nibanks


ManickaP:
The advantage of building msquic ourselves was that we were able to consume the latest greatest before its officially published. Since it takes manual intervention and still has many issues (signing atm), wouldn't it be better to continue building it ourselves for our tests?

Also my layout for 1.9 and 2.0 locally is like this:
```
lrwxrwxrwx 1 root root      24 Mar 24 10:55 /usr/lib/libmsquic.lttng.so.1 -> libmsquic.lttng.so.1.9.0
-rwxr-xr-x 1 root root 6744640 Mar 24 10:55 /usr/lib/libmsquic.lttng.so.1.9.0
-rwxr-xr-x 1 root root 7198856 Apr  8 18:08 /usr/lib/libmsquic.lttng.so.2.1.0
lrwxrwxrwx 1 root root      18 Mar 24 10:55 /usr/lib/libmsquic.so.1 -> libmsquic.so.1.9.0
-rwxr-xr-x 1 root root 5547816 Mar 24 10:55 /usr/lib/libmsquic.so.1.9.0
lrwxrwxrwx 1 root root      18 Apr  8 18:09 /usr/lib/libmsquic.so.2 -> libmsquic.so.2.1.0
-rwxr-xr-x 1 root root 5795616 Apr  8 18:09 /usr/lib/libmsquic.so.2.1.0
```

I did fix [SONAME in 1.9 msquic branch](https://github.com/ManickaP/msquic/commit/a82c80f8770c4759cb087308637a83e3bdac296f) and do packaging myself though.

wfurt:
We don't need to `.so.1` imho as .NET6 will happily load '.so' and it does not conflict with 2.0. 
I'm still not sure about updating strategy. We will probably not rebuild the container on every change. 
I'm hoping we will figure out how to consume bits from MsQuic pipeline at some point.

src/fedora/36/amd64/Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM registry.fedoraproject.org/fedora:36
+
+# Install the base toolchain we need to build anything (clang, cmake, make and the like)
```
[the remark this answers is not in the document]
wfurt: good point AFAIK we don't need any of this to run Helix. (@MattGal to confirm) So this is probably question for @mthalman if we want to add everything for consistency/legacy or if we want to keep the images small e.g. keeping only what we need.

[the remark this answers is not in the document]
MattGal: You need a valid python 3, the rest is leftovers from copy-paste stuff.  Remove all you like; anything that's automatically present won't change the layer anyways, though.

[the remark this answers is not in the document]
MattGal: Oops deleted previous comment.  This isn't a Helix image, it's a build image by the name; so yeah this should be what you need to run the build.

src/fedora/36/helix/amd64/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+# Install Helix Dependencies
+
+RUN dnf install -y \
+        openssl \
+        python3 \
+        python3-devel \
+        libatomic \
+        libffi-devel \
+        redhat-rpm-config \
+        sudo && \
```
MattGal, line 12: these are the Helix dependencies.  Stuff like openSSL, python, and sudo are absolutely needed but if you're certain that they're already there enough to risk the several hours putting them back, it's fine to remove them; as noted it will only impact this one image anyways.

MattGal (approved):

src/fedora/36/helix/amd64/Dockerfile
wfurt: How about the pip & virtualenv? Can we get them via system packages instead of installing them directly?

MattGal: Pip's the package manager and it's not installed in this docker image by default:
``` 
C:\scratch>docker run -it registry.fedoraproject.org/fedora:36 bash
[root@5daa616fd751 /]# python3 -m pip
/usr/bin/python3: No module named pip
```
So yes, helix still needs this to be present for later install commands below.

wfurt: There is. I explicitly remove it bellow with `RUN dnf remove -y python3-pip && \` because it was conflating with the custom logic.

mthalman (approved):

mthalman:
@wfurt - Is this ready to merge?

wfurt:
yes, it slipped from my radar @mthalman. I'll make updates once we have signed msquic packages available.

PR #606, opened 2022-04-20
upgrade llvm from 3.9 to 9

wfurt:
upgrade llvm to avoid coreclr crashes.
related to https://github.com/dotnet/runtime/issues/66970

curl is already on 7.45.0 so I removed the custom build and added distro package

### DIFF
--- a/src/ubuntu/18.04/amd64/Dockerfile
+++ b/src/ubuntu/18.04/amd64/Dockerfile
@@ -5,20 +5,20 @@ FROM ubuntu:18.04
 # them in a different layer.
 RUN apt-get update \
     && apt-get install -y \
-        clang-3.9 \
+        clang-9 \
         gdb \
-        liblldb-3.9-dev \
-        lldb-3.9 \
-        llvm-3.9 \
+        liblldb-9-dev \
+        lldb-9 \
+        llvm-9 \
         locales \
         make \
-        python-lldb-3.9 \
+        python-lldb-9 \
         sudo \
         wget \
     && rm -rf /var/lib/apt/lists/* \
-    && wget https://cmake.org/files/v3.10/cmake-3.10.2-Linux-x86_64.tar.gz \
-    && tar -xf cmake-3.10.2-Linux-x86_64.tar.gz --strip 1 -C /usr/local \
-    && rm cmake-3.10.2-Linux-x86_64.tar.gz
+    && wget https://cmake.org/files/v3.23/cmake-3.23.1-linux-x86_64.tar.gz \
+    && tar -xf cmake-3.23.1-linux-x86_64.tar.gz --strip 1 -C /usr/local \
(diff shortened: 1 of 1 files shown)


src/ubuntu/18.04/amd64/Dockerfile
```diff
@@ -56,6 +56,7 @@ RUN apt-get update \
     && apt-get install -y \
         autoconf \
         automake \
+        curl \
```
janvorli, line 59: I think we should keep building the curl from sources. There was a reason for that, I think it was to get libcurl built with the right APIs.

wfurt: in the very old days, we used curl directly in HttpClient and it did matter what SSL backend was used. I don't think this is the case any more. 
That change is 4 years old touching libcurl-openssl1.0-dev as well.

janvorli: Ok, then I'm fine with it.

janvorli (approved):
LGTM, thank you!

src/ubuntu/18.04/amd64/Dockerfile
```diff
@@ -5,20 +5,20 @@ FROM ubuntu:18.04
 # them in a different layer.
 RUN apt-get update \
```
jkotas, line 6: Ubuntu 18.04 is end of life for almost a year. You should switch to a newer Ubuntu version that is still supported. Otherwise, you will get a security compliance issue filled against you sooner or later.

jkotas: I think it would be best to delete whole Ubuntu 18.04 directory here so that we do not have to maintain and build these docker files.

jkotas: Actually, ignore this. I got confused. Ubuntu 18.04 flavors are end of life, but Ubuntu 18.04 LTS is still supported.

jkotas (approved):

src/ubuntu/18.04/amd64/Dockerfile
wfurt: ok. We have other places where we use ubuntu 18. let me do that in separate steps. 
Unless there is strong objection @jkotas I would merge and try CI to verify that the crash is gone (it was for me running local build) 
I will follow-up with 18 upgrades to 20 or 22. That may take more testing.

BTW do you know if we plan to drop support for Ubuntu18 in 7.0? We seems to claim support even for Ubuntu 16: https://github.com/dotnet/core/blob/main/release-notes/6.0/supported-os.md

jkotas: As I have said, ignore my above comment. Ubuntu 18.04 is still supported. I have looked at a wrong page.

PR #608, opened 2022-04-21
Add Ubuntu 22.04 debpkg images

am11:
This will be used in the runtime repo to build debpkg.

### DIFF
--- a/src/ubuntu/22.04/amd64/Dockerfile
+++ b/src/ubuntu/22.04/amd64/Dockerfile
@@ -1,0 +1,53 @@
+FROM ubuntu:22.04
+
+# Install the base toolchain we need to build anything (clang, cmake, make and the like)
+# this does not include libraries that we need to compile different projects, we'd like
+# them in a different layer.
+RUN apt-get update \
+    && apt-get install -y \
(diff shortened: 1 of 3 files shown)


am11:
@mthalman, 22.04 helix error is:

> Unhandled exception: System.InvalidOperationException: Found published digest 'sha256:023ff689211dd7ccd961d04fd91454b9a8edba4ba306611ada8c13a1e4647157' for tag 'ubuntu:22.04' but could not find a matching digest value from the set of locally pulled digests for this tag: ubuntu@sha256:06b5d30fabc1fc574f2ecab87375692299d45f8f190d9b71f512deb494114e1f. This most likely means that this tag has been updated since it was last pulled.

it was built before 22.04 was officially released (today). Can we clear the cache and rebuild it?

mthalman:
Yep, that would seem to be the issue. Bad timing. 😉 A simple rebuild should take care of it. I've queued a rebuild.

mthalman (approved):

mthalman:
Looks like `clang-9` package doesn't exist.

mthalman (approved):

PR #609, opened 2022-04-26
Add ARM/ARM64 crossbuild PGO support

janvorli:
This change adds new command line argument to the build-rootfs.sh script that adds clang libraries necessary to enable PGO on arm64 and arm. 
It also creates a new image src/ubuntu/18.04/cross/arm, as the PGO libraries are not available for Ubuntu 16.04 and 16.04 is out of support anyways. So after the images are generated, the plan is to switch runtime to build against Ubuntu 18.04 for ARM and ARM64.


janvorli:
cc: @davidwrighton

davidwrighton (approved):

mthalman (approved):

PR #610, opened 2022-04-26
Update centos7 rpmpkg to use ruby 2.5

wtgodbe:
Attempt to update the centos7 image to fix https://github.com/dotnet/aspnetcore/issues/37406.

### DIFF
--- a/src/centos/7/rpmpkg/Dockerfile
+++ b/src/centos/7/rpmpkg/Dockerfile
@@ -1,13 +1,12 @@
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:centos-7
 
 # Install the rpm packaging toolchain required to build rpms
-RUN yum clean all
-RUN yum --enablerepo=centos-sclo-rh -y install rh-ruby25
-RUN yum --enablerepo=centos-sclo-rh -y install rh-ruby25-ruby-devel
-RUN yum --enablerepo=centos-sclo-rh -y install rh-ruby25-rubygems
-RUN yum install -y \
+RUN yum clean all \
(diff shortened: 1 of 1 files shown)


dougbu (approved):
I don't know whether they're better nor whether all remain compatible w/ CentOS 7 but `rh-ruby26*`, `rh-ruby27*`, and `rh-ruby30` packages are available.

wtgodbe:
@mthalman I separated it out to try to find which install command was failing:

> Step 7/9 : RUN gem install --no-document         git:1.7.0         ffi:1.12.2
 ---> Running in 22bbc5163ece
/bin/sh: gem: command not found

does `RUN yum --enablerepo=centos-sclo-rh -y install rh-ruby25-rubygems` not achieve that?

mthalman:
> @mthalman I separated it out to try to find which install command was failing:

If you're still working on debugging the Dockerfile, it would be better to mark this as a draft PR. In addition, it would be better to install Docker Desktop on your local machine and build the Dockerfile locally to diagnose issues.

> > Step 7/9 : RUN gem install --no-document         git:1.7.0         ffi:1.12.2
> > ---> Running in 22bbc5163ece
> > /bin/sh: gem: command not found
> 
> does `RUN yum --enablerepo=centos-sclo-rh -y install rh-ruby25-rubygems` not achieve that?

I don't have much experience in this area. Looking at the other Dockerfiles in this repo that use gem, I see they just install `rubygems`.

wtgodbe:
Ok, should be working now

dougbu (approved):

mthalman (approved):

PR #612, opened 2022-04-29
Creating a 20.04 helix container

LakshanF:
SDK has moved to Ubuntu 20.04 and creating a docker container with some requires pre-requisites to run a SDK test

### DIFF
--- a/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
+++ b/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
@@ -1,8 +1,11 @@
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-helix-amd64
 
+
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Need some additional packages
+
+USER root
 
 RUN apt-get clean && \
     mv /etc/apt/sources.list /etc/apt/sources.list1 && apt-get update && \
@@ -24,3 +27,4 @@ RUN sudo npm install jsvu -g && \
 
 ENV PATH="/home/helixbot/.jsvu:${PATH}"
 
+USER helixbot
--- a/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
+++ b/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
@@ -17,11 +17,9 @@ RUN apt-get clean && \
      \
      && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 
-
 # Install V8 & other engines in the non-root context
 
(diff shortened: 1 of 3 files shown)


mthalman (commented):
Can you refactor https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile so that it is based on this? This will allow for better reusability amongst image layers.

You've already defined a local tag of `ubuntu-20.04-helix-amd64` in the manifest, so the wasm Dockerfile can reference that as its base.

I think the only thing it's doing different than this helix Dockerfile is installing the JS Engine.

src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
```diff
@@ -1,59 +1,20 @@
-FROM ubuntu:20.04
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-helix-amd64
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-# sources tweak is to work around very slow update
+# Need some additional packages
 
 RUN apt-get clean && \
```
mthalman, line 7: You'll need to add `USER root` before you can run package updates. Then at the end of the Dockerfile, you'll need to switch back: `USER helixbot`.

src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
```diff
@@ -1,9 +1,12 @@
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-helix-amd64
 
+
```
mthalman, line 3: Nit: Remove extra line

src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
```diff
@@ -24,3 +27,4 @@ RUN sudo npm install jsvu -g && \
 
 ENV PATH="/home/helixbot/.jsvu:${PATH}"
 
+USER helixbot
```
mthalman, line 30: Nit: include newline at end of file

src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
LakshanF: Thanks!

mthalman (commented):
The build error is due to the use of `sudo` on this line: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/612/files#diff-0e33dbea0c0e14c38703e38fe31be24dc63e4ac8721bfe4d6252185fb9a02f16R24. Using `sudo` is no longer needed there because it's running as root. Prior to this change, that Dockerfile was configured to be running as helixbot at that point.

mthalman (approved):

PR #615, opened 2022-05-09
Add Mariner 2.0 variant

richlander:
- We'd like to use [Mariner](https://github.com/microsoft/CBL-Mariner) to build .NET, likely replacing both CentOS 7 and Ubuntu LTS (to build `linux` artifacts).
- We need this Dockerfile to do that.
- This Dockerfile has been tested and successfully builds dotnet/runtime.

It is straightforward to see that the Mariner Dockerfile is a significant improvement on the [CentOS 7 Dockerfile](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/centos/7/Dockerfile) we are using today (to produce `linux-x64` builds).


MichaelSimons:
@richlander, you will need to add a [manifest entry](https://github.com/dotnet/dotnet-buildtools-prereqs-docker#how-to-modify-or-create-a-new-image) for this new Dockerfile.

richlander:
@MichaelSimons -- I updated the manifest, following the pattern I saw used with Ubuntu. Was that correct?

src/cbl-mariner/manifest.json
```diff
@@ -14,6 +14,21 @@
...
+        },
+        {
+          "platforms": [
+            {
+              "dockerfile": "src/cbl-mariner/2.0/amd64",
+              "os": "linux",
+              "osVersion": "cbl-mariner2.0",
+              "tags": {
+                "cbl-mariner-2.0-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "cbl-mariner-2.0": {
```
MichaelSimons, line 26: The `cbl-mariner-2.0` entry is not needed.  The other parts look correct.

src/cbl-mariner/2.0/amd64/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+        lzma \
+        make \
+        ncurses-devel \
+        numactl-devel \
+        openssl-devel \
+	pkgconf \
+        readline-devel \
+        python3 \
+        python3-devel \
+	sed \
```
mthalman, line 34: Making sure these are alphabetized.

```suggestion
	pkgconf \
        python3 \
        python3-devel \
        readline-devel \
	sed \
```

src/cbl-mariner/2.0/amd64/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM mcr.microsoft.com/cbl-mariner/base/core:2.0
+RUN tdnf install -y \
+        ca-certificates \
+        autoconf \
+        automake \
+        build-essential \
+        clang \
```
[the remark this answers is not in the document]
richlander: Good call. I tried to that, but clearly failed (quite badly!). Thanks for just doing it. Appreciated.

mthalman (approved):

PR #631, opened 2022-07-05
Clean up eng/common directory

mthalman:
Deletes files in the eng/common directory that are no longer needed. The eng/common directory gets automatically populated by automation from the https://github.com/dotnet/docker-tools repo but this only handles added or modified files. File deletions need to be done explicitly in each repo that consumes eng/common content.


MichaelSimons (approved):
Is there an issue tracking the docker-tools infra to remove these when flowing changes?

mthalman:
> Is there an issue tracking the docker-tools infra to remove these when flowing changes?

Yes: https://github.com/dotnet/docker-tools/issues/312

PR #636, opened 2022-07-11
Add docker images from Aspnet/Benchmarks

wtgodbe:
Add docker images from https://github.com/aspnet/Benchmarks to resolve some docker security alerts

CC @sebastienros @JamesNK

### DIFF
--- a/src/debian/8/aspnet-benchmarks/mongo/Dockerfile
+++ b/src/debian/8/aspnet-benchmarks/mongo/Dockerfile
@@ -1,0 +1,4 @@
+FROM mongo:3.4.10
+
+ADD create.js /docker-entrypoint-initdb.d/
+RUN chmod 755 /docker-entrypoint-initdb.d/create.js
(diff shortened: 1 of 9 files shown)


src/debian/8/aspnet-benchmarks/mongo/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mongo:3.4.10
+
+ADD create.js /docker-entrypoint-initdb.d/
```
mthalman, line 3: `COPY` is preferred over `ADD`.
But this file doesn't even exist, resulting in a build error. So what is the intent here? Same question applies to a bunch of the other Dockerfiles that are using `ADD`.

mthalman (commented):
In general, we need to be cleaning up the base images in this repo to be using golden images when available or OSS mirrored images.

Please read https://eng.ms/docs/more/containers-secure-supply-chain/approved-images. For example, there are golden images for golang that should be used instead of the ones from Docker Hub. If a golden image isn't available, you can request mirrored images: https://eng.ms/docs/more/containers-secure-supply-chain/next-step#requesting-a-new-image-for-mcrmirror-aka-azcu-retag

I didn't comment on each of the base images being used but the process I described above applies to all of them.

mthalman:
@wtgodbe - This PR has been around for a while. Will you be making further updates?

wtgodbe:
Yes, will get back to this once 7.0 stuff cools down

PR #643, opened 2022-07-28
Add jq in runtime dependencies

am11:
`jq` is a tiny (< 100 KB) recommended tool to parse JSON in shell scripts.

In dotnet/arcade, we are using jq in a few places with a fallback to 'unrecommended' manual parsing path in bash scripts.

This PR is adding `jq` to base dependency lists, so we use the recommended code path.

### DIFF
--- a/src/ubuntu/18.04/mono/amd64/Dockerfile
+++ b/src/ubuntu/18.04/mono/amd64/Dockerfile
@@ -15,8 +15,8 @@ RUN apt-get update && \
         gdb \
         gettext \
         git \
-        jq \
         iputils-ping \
+        jq \
         libgdiplus \
         libtool \
         locales \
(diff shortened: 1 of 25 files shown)


src/ubuntu/18.04/mono/amd64/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apt-get update && \
         gdb \
         gettext \
         git \
+        jq \
```
mthalman, line 18: Slightly out of order. Needs to be moved to be under `iputils-ping`. Also applies to the other Ubuntu 18.04 Dockerfiles.

mthalman (approved):

src/ubuntu/18.04/mono/amd64/Dockerfile
am11: Thanks, will update once https://github.com/dotnet/arcade/pull/10175.goes in. :)

am11: Fixed.

PR #644, opened 2022-07-28
Add linux-riscv64 image

am11:

PR #646, opened 2022-08-01
Update Android Dockerfile to install Docker

simonrozsival:
We need to be able to build a Docker image during the Android build process so that we can run a gRPC testing server (see https://github.com/dotnet/runtime/pull/73060).

cc @directhex

### DIFF
--- a/src/ubuntu/18.04/android/Dockerfile
+++ b/src/ubuntu/18.04/android/Dockerfile
@@ -13,6 +13,21 @@ RUN wget https://cmake.org/files/v3.23/cmake-3.23.1-linux-x86_64.tar.gz \
     && tar -xf cmake-3.23.1-linux-x86_64.tar.gz --strip 1 -C /usr/local \
     && rm cmake-3.23.1-linux-x86_64.tar.gz
 
+# Install Docker
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        gnupg2 \
+        software-properties-common \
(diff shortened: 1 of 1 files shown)


mthalman:
How are containers from this image intended to be created?

This is a docker-in-docker scenario which requires specific configuration to work properly. It either needs to be run with the Docker socket volume-mounted or with the privileged option. I see that the image is being referenced to be used as a container job: https://github.com/dotnet/runtime/blob/cfe26ecb3b677a506711878c19586ddd06117d0e/eng/pipelines/common/platform-matrix.yml#L209-L211. Does that even support either of those methods?

simonrozsival:
The `/var/run/docker.sock` volume should be mounted correctly (https://dev.azure.com/dnceng/public/_build/results?buildId=1911472&view=logs&j=0657b2d6-71c1-5c98-76b8-e3cb824cf6bd&t=7162708e-c8ea-47a5-bc28-5a1a3986e307&l=77) so docker-in-docker should work. @directhex tested it locally so he can fill in more details.

mthalman:
Ah, thanks for the pointer. Yes, that should work then.

mthalman (approved):

mthalman:
I can merge this but the Ubuntu pipeline is currently broken: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/644#issuecomment-1201351851. So that needs to get resolved before an image for this can be built.

mthalman:
Image is published now. Tag: `mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-android-20220803135138-8fcaabc`

PR #647, opened 2022-08-01
Use Ubuntu 20.04 (or better) repositories on host

am11:
This is a workaround for official machine host, which is still running Ubuntu 18.04: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/644#issuecomment-1201520888.

### DIFF
--- a/src/ubuntu/build-scripts/install-cross-build-prereqs.sh
+++ b/src/ubuntu/build-scripts/install-cross-build-prereqs.sh
@@ -9,8 +9,7 @@ set -u
 # which are available in Ubuntu 20.04+ repositories
 . /etc/os-release
 if [ "$(echo "$VERSION_ID" | tr -d .)" -lt 2004 ]; then
-    cat /etc/apt/sources.list | sed "s/$UBUNTU_CODENAME/focal/g" > /tmp/focal.list
-    sudo mv /tmp/focal.list /etc/apt/sources.list.d/
+    sed "s/$UBUNTU_CODENAME/focal/g" /etc/apt/sources.list | sudo dd of=/etc/apt/sources.list.d/focal.list
 fi
 
 # see (see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/120)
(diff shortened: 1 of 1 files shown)


mthalman:
I did verify that this change allows the build to succeed on the official build.

mthalman (approved):

am11:
Not sure why this is failing on DevOps as there is no error message. In addition to local testing, I have tested the same command (`exec src/ubuntu/build-scripts/install-cross-build-prereqs.sh`) on GitHub Actions using ubuntu-1804 and it worked: https://github.com/am11/dotnet-buildtools-prereqs-docker/actions/runs/2778279962. Can't reproduce.

am11:
Worked on DevOps with a separate list file. :shrug:

akoeplinger (approved):

directhex (changes requested):
I think this is a brittle approach.

Swapping the entire repository out means that whilst individual packages _may_ install cleanly from the 20.04 repo onto 18.04, others may cause a dependency cascade and result in an entirely untested halfway state between both distributions (e.g. if something requires a newer libc, that will be pulled in - and either uninstall or forcibly upgrade half the world in order to reconcile the world)

Whilst it is more manual, I think it is safer to hand-vet which 20.04 packages we try to layer on top of 18.04. That comes with a list of _forseen_ consequences - just swapping the repo to 20.04 comes with tens of thousands of potential _unforseen_ consequences.

am11:
The "layer" where this is done has no effect on what goes inside the image or libc.

am11:
@mthalman, let me know if you want a different strategy here.

From my point of view, we can merge this as-is to unblock official build pipeline. As you know that:

1. this script is used only in the context of Ubuntu host to install three dependencies (which is the subset of entire matrix). 
2. Everything, after this script has executed, builds inside containers so the surface area of this theoretical downside is very small 3. Even if there was some kind of clash, it would have no impact on the images produced by docker running on this host, but rather fail to run builds at places where it uses qemu, which hasn't happened in CI runs.

Also, this is point in time hack; in few months when we switch to Ubuntu 20.04 VMs, this will be a dead condition.

PR #649, opened 2022-08-03
add msquic to Alpine Helix image

wfurt:
This is similar to what we do for fedora/34. There is no package so we build msquic from our clone. 
This change should allow us to run HTTP/3 & Quic tests on Alpine.
We may need to update this when MsQuic has new release (and we update our clone) 

contributes to https://github.com/dotnet/runtime/issues/55639

I was also looking to add it to arm64 but there is no PowerShell so we would either need to compile it from sources or somehow avoid use of pwsh.

### DIFF
--- a/src/alpine/3.14/helix/amd64/Dockerfile
+++ b/src/alpine/3.14/helix/amd64/Dockerfile
@@ -15,6 +15,18 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
     pip install ./helix_scripts-*-py3-none-any.whl
 
+# build MsQuic as we don't have packages
+RUN cd /tmp && \
+    mkdir pwsh && \
+    cd pwsh && \
+    curl -L https://github.com/PowerShell/PowerShell/releases/download/v7.2.5/powershell-7.2.5-linux-alpine-x64.tar.gz | tar xfz - && \
+    cd .. && \
+    git clone --depth 1 --single-branch --recursive https://github.com/dotnet/msquic && \
+    cd msquic/src/msquic  &&  PATH=~/.dotnet/tools:$PATH /tmp/pwsh/pwsh scripts/build.ps1 -Config Release && \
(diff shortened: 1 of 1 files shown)


src/alpine/3.14/helix/amd64/Dockerfile
```diff
@@ -15,6 +15,18 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
...
     pip install ./helix_scripts-*-py3-none-any.whl
 
+# build MsQuic as we don't have packages
+RUN cd /tmp && \
+    mkdir pwsh && \
+    cd pwsh && \
+    curl -L https://github.com/PowerShell/PowerShell/releases/download/v7.2.5/powershell-7.2.5-linux-alpine-x64.tar.gz | tar xfz - && \
+    cd .. && \
+    git clone --depth 1 --single-branch --recursive https://github.com/dotnet/msquic && \
+    cd msquic/src/msquic  &&  PATH=~/.dotnet/tools:$PATH /tmp/pwsh/pwsh scripts/build.ps1 -Config Release && \
```
rzikm, line 25: Should we use `-UseSystemOpenSSLCrypto`?

rzikm (approved):
LGTM, modulo one question

mthalman:
> I was also looking to add it to arm64 but there is no PowerShell so we would either need to compile it from sources or somehow avoid use of pwsh.

For reference, there is an existing issue for Alpine-based Arm64 support at https://github.com/PowerShell/PowerShell/issues/12937.

To work around this, you can use a multi-stage Dockerfile to build msquic in a non-Alpine-based stage and copy the resulting binaries to the final Alpine-based stage.

wfurt:
> Is there any reason not to apply these changes to the Alpine 3.15 Dockerfile instead?

Not really. I look at what runtime uses for testing now and that was 3.14. Maybe I can add it to 3.15 as well but I did not want to change OS version just to add MsQuic. I can do that as follow-up if we get stable CI. (so as arm architectures)

src/alpine/3.14/helix/amd64/Dockerfile
wfurt: We could. I don't think it matters much as this is used only by this one particular image. It would probably produce slightly smaller binary.

mthalman (approved):

PR #652, opened 2022-08-03
retire msquic image and add dependencies to base images

wfurt:
we will need to build arm 32/64 and I did not want to maintain separate images. `gem` is pretty small.


src/ubuntu/18.04/arm32v7/Dockerfile
```diff
@@ -15,4 +15,6 @@ RUN apt-get update && \
         libssl-dev \
         libtool \
         libkrb5-dev \
+        ruby-deb \
```
mthalman, line 18:
```suggestion
        ruby-dev \
```

mthalman (approved):

wfurt:
BTW is there some way how to build arm containers on x64 @mthalman? I should have found it my build but we seems to skip the arm containers...

mthalman:
> BTW is there some way how to build arm containers on x64 @mthalman? I should have found it my build but we seems to skip the arm containers...

Yes, there is a way but not without modification to the build.ps1 script. It's configured to only build the architecture that matches the host. But this is good feedback to improve the script.

PR #668, opened 2022-08-16
Add Debian 11 amd64 dockerfile

lbussell:
Added a Debian 11 dockerfile for amd64. I think it is the same as the arm64v8 one, since all of the packages are available in both architectures.


lbussell:
cc @MichaelSimons @mthalman @crummel since I don't have permissions to add reviewers.

mthalman (approved):

MichaelSimons (approved):

PR #669, opened 2022-08-17
Add haiku-x64 image

am11:
cc @jessicah, the base layer (crossdeps) currently installs cmake 3.23, which has your changes in `Modules/Platform/Haiku.cmake`.

### DIFF
--- a/src/ubuntu/20.04/cross/haiku/Dockerfile
+++ b/src/ubuntu/20.04/cross/haiku/Dockerfile
@@ -1,0 +1,3 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-crossdeps
+
+ADD rootfs.x64.tar crossrootfs
--- a/src/ubuntu/20.04/cross/haiku/hooks/post-build
+++ b/src/ubuntu/20.04/cross/haiku/hooks/post-build
@@ -1,0 +1,1 @@
+../../../../build-scripts/build-rootfs-cleanup.sh
--- a/src/ubuntu/20.04/cross/haiku/hooks/pre-build
+++ b/src/ubuntu/20.04/cross/haiku/hooks/pre-build
@@ -1,0 +1,5 @@
+#!/usr/bin/env sh
+
+SCRIPT=$(readlink -f "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+$SCRIPTPATH/../../../../build-scripts/build-rootfs.sh ubuntu-20.04 haiku x64
--- a/src/ubuntu/20.04/cross/haiku/hooks/pre-build
+++ b/src/ubuntu/20.04/cross/haiku/hooks/pre-build
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../build-scripts/build-rootfs.sh ubuntu-20.04 haiku x64
+$SCRIPTPATH/../../../../build-scripts/build-rootfs.sh ubuntu-20.04 haiku x64 "" /rootfs/x64/boot/system/bin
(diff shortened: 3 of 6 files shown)


am11:
@mthalman, I tried to add new platforms (linux-riscv64 and haiku-x64) in 22.04, but build.ps1 fails during cross rootfs script execution due to seccomp requirements in docker builder. I think all latest OS (Ubuntu, Fedora and Alpine Linux) require these new security features. Are we tracking its progress somewhere?

mthalman:
> @mthalman, I tried to add new platforms (linux-riscv64 and haiku-x64) in 22.04, but build.ps1 fails during cross rootfs script execution due to seccomp requirements in docker builder. I think all latest OS (Ubuntu, Fedora and Alpine Linux) require these new security features. Are we tracking its progress somewhere?

Can you be more specific about what's needed? The docker commands for running the script are at https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/ubuntu/build-scripts/build-rootfs.sh.

am11:
> Can you be more specific about what's needed?

@mthalman, I initially added linux-riscv64 and haiku-x64 first under 22.04 and the build was failing with error indicating that  _some_ security option has changed in 22.04 but I don't have all the details other than "i'ts not working", and I thought you might have some ideas. :slightly_smiling_face:

Switching (e.g. riscv64) from Ubuntu 20.04 to 22.04 breaks like this:

```sh
$ pwsh build.ps1 -DockerfilePath "*ubuntu/22.04/cross/riscv64*"
...
Cloning into '/scripts'.
...
mknod: /rootfs/riscv64/test-dev-null: Operation not permitted
E: Cannot install into target '/rootfs/riscv64' mounted with noexec or nodev
...
```
and the docker version in the executing layer (build-rootfs.sh) is:

```
Client:
 Version:           20.10.16
 API version:       1.41
...
Server: Docker Engine - Community
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
...
```

so something (like -- but not exactly -- this https://github.com/adoptium/containers/issues/215#issuecomment-1142046045) has changed between Ubuntu Focal and Jammy.

am11:
> Can you be more specific about what's needed?

@mthalman, I initially added linux-riscv64 and haiku-x64 under 22.04 and the build was failing with error indicating that  _some_ security option has changed in 22.04 but I don't have all the details other than "i'ts not working", and I thought you might have some ideas. :slightly_smiling_face:

Switching (e.g. riscv64) from Ubuntu 20.04 to 22.04 breaks like this:

```sh
$ pwsh build.ps1 -DockerfilePath "*ubuntu/22.04/cross/riscv64*"
...
Cloning into '/scripts'.
...
mknod: /rootfs/riscv64/test-dev-null: Operation not permitted
E: Cannot install into target '/rootfs/riscv64' mounted with noexec or nodev
...
```
and the docker version in the executing layer (build-rootfs.sh) is:

```
Client:
 Version:           20.10.16
 API version:       1.41
...
Server: Docker Engine - Community
 Engine:
  Version:          20.10.17
  API version:      1.41 (minimum version 1.12)
...
```

so something (like -- but not exactly -- this https://github.com/adoptium/containers/issues/215#issuecomment-1142046045) has changed between Ubuntu Focal and Jammy which is causing the `Operation not permitted`.

mthalman:
@am11 - Why is this thought to be related to Docker seccomp? Have you verified the build-rootfs.sh script works when run, outside the context of a container, in a 22.04 VM or WSL?

am11:
@mthalman, running build-rootfs on 22.04 VM does not give that error. One my virtual machine running 20.04 in managed cloud, I was getting the aforementioned error with 22.04 docker, but 20.04 docker gives no error on the same machine. I have just tested 22.04 container on GithHb Actions\s 20.04 VM, and it succeeded without any error: https://github.com/am11/CrossRepoCITesting/actions/runs/2883652586. So it must be something to do with my cloud VM environment that require some update for 22.04 container, seccomp option was my best guess based on googling..

I have opened https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/673 to test our AzDO. If it works, that is fine and better to add these new platforms in latest LTS. (it will buy us three / four years till EOL)

PR #672, opened 2022-08-17
switch to msquic main to pick up arm fixes

wfurt:
fixes preventing test runs are in main now. (and will flow to msquic 2.1.1 eventually)

### DIFF
--- a/src/debian/10/helix/arm64v8/Dockerfile
+++ b/src/debian/10/helix/arm64v8/Dockerfile
@@ -56,7 +56,7 @@ RUN cd /tmp && \
     echo "709265A0B99232CD8AB6F8A02C01F3AEE94262B959E1A4B0FD871C8789C07396 powershell-7.2.5-linux-arm64.tar.gz" | sha256sum --check - && \
     tar xf powershell-7.2.5-linux-arm64.tar.gz && \
     cd .. && \
-    git clone --depth 1 --single-branch --branch release/7.0 --recursive https://github.com/dotnet/msquic && \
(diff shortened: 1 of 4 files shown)


rzikm (approved):

mthalman:
Builds are failing:

```
cp: cannot stat 'artifacts/bin/linux/arm64_Release_openssl/libmsquic.lttng.so.2.1.0': No such file or directory
```

rzikm (changes requested):
Main is building different version, apparently
> arm64_Release_openssl/libmsquic.so.2.2.0

wfurt:
It seems like the build just retires in loop and then it fails with timeout @mthalman. It would be nice if it just fails with the error at the end...

mthalman:
> It seems like the build just retires in loop and then it fails with timeout @mthalman. It would be nice if it just fails with the error at the end...

Yeah, it's unfortunate. The retry is useful because often times there are connectivity issues with stuff being accessed as part of the build. The retry means you don't have to start from scratch with these really long builds, assuming that a retry fixes the issue. The only way to fix it is to ensure the timeout is long enough to account for retries. But that has a downside as well if something gets blocked in the build, it can cause a waste of resources if the timeout is really long. It's all about striking a balance.

wfurt:
I see it useful for intermittent failures. Maybe limit it to just 1 or 2? But I also learned my lesson. I'm always suspicious of ARM because of timing and performance so I did not dig deeper into the logs.

rzikm (approved):

PR #680, opened 2022-08-31
Add ubuntu-22.04-cross-arm image

am11:
Moved `ubuntu-18.04-cross-arm-16.04` to `ubuntu-22.04-cross-arm`, since we already have `ubuntu-18.04-cross-arm` (xenial variant under bionic is a strange combination and it is not used by anyone). `22.04-cross-arm64` and `18.04-cross-arm64` are already present, so I have simply deleted the unused variant `18.04-cross-arm64-16.04`.

Also removed the unused argumetn `lldb14` from alpine-arm; we only use lldbX argument for Debian or Ubuntu in rootfs, so this is unused and confusing for the reader.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,16 +98,16 @@ The [manifest.json](./manifest.json) contains metadata used by the build infrast
 
 ### Image Dependency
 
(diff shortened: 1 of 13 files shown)


mthalman (approved):

mthalman:
Thanks for cleaning things up.

am11:
FWIW, I was using grep.app to find the usage in dotnet org. e.g. https://grep.app/search?q=arm64-16.0&filter[repo.pattern][0]=dotnet. Their caching is sometimes a few hours / days behind, but it gives an idea on frequency of usage (organization-wide).

PR #683, opened 2022-09-02
Fix codename in riscv64 pre-build hook

am11:
We have started to recognize `jammy` codename in arcade. Before https://github.com/dotnet/arcade/pull/10520, we were just ignoring the codename.

### DIFF
--- a/src/ubuntu/22.04/cross/riscv64/hooks/pre-build
+++ b/src/ubuntu/22.04/cross/riscv64/hooks/pre-build
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../build-scripts/build-rootfs.sh ubuntu-22.04 jammy riscv64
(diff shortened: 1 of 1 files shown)


mthalman:
Build is failing because Arcade's build-rootfs.sh script is doing something with pkgsrc.joyent.com which has an expired cert.

```
--2022-09-06 14:06:09--  https://pkgsrc.joyent.com/packages/SmartOS/trunk/x86_64/All
Resolving pkgsrc.joyent.com (pkgsrc.joyent.com)... 64.30.129.74
Connecting to pkgsrc.joyent.com (pkgsrc.joyent.com)|64.30.129.74|:443... connected.
ERROR: cannot verify pkgsrc.joyent.com's certificate, issued by ‘CN=R3,O=Let's Encrypt,C=US’:
  Issued certificate has expired.
```

mthalman (approved):

am11:
> Build is failing because Arcade's build-rootfs.sh script is doing something with pkgsrc.joyent.com which has an expired cert.

This shouldn't block us. You can always pass `--use-mirror` from the pre hook.

PR #690, opened 2022-09-09
Add msquic to OpenSUSE

rzikm:
This PR installs MsQuic 2.1.1 from official rpm from github (I did not find a way to add packages.microsoft.com feed to zypper). Tests run, except that IPv6 is apparently disabled on the image. That is going to be mitigated by https://github.com/dotnet/runtime/pull/75341.

### DIFF
--- a/src/opensuse/15.2/helix/amd64/Dockerfile
+++ b/src/opensuse/15.2/helix/amd64/Dockerfile
@@ -35,9 +35,13 @@ RUN zypper ref && \
 
 ENV LANG=en_US.utf8
 
-RUN wget https://github.com/microsoft/msquic/releases/download/v2.1.1/libmsquic-2.1.1-1.x86_64.rpm && \
-    rpm -i libmsquic-2.1.1-1.x86_64.rpm && \
-    rm libmsquic-2.1.1-1.x86_64.rpm
+# Install MsQuic from official GitHub releases
+RUN wget https://packages.microsoft.com/keys/microsoft.asc && \
+    echo 2cfd20a306b2fa5e25522d78f2ef50a1f429d35fd30bd983e2ebffc2b80944fa microsoft.asc | sha256sum --check - && \
+    rpm --import microsoft.asc && \
+    rm microsoft.asc && \
+    zypper install -y https://github.com/microsoft/msquic/releases/download/v2.1.1/libmsquic-2.1.1-1.x86_64.rpm && \
+    zypper clean -a
 
 RUN ln -sf /usr/bin/python3 /usr/bin/python && \
     python -m pip install --upgrade pip==20.2 && \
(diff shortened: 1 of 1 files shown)


src/opensuse/15.2/helix/amd64/Dockerfile
```diff
@@ -28,12 +28,17 @@ RUN zypper ref && \
...
         sudo \
+        wget \
         unzip && \
         \
         zypper clean -a
 
 ENV LANG=en_US.utf8
 
+RUN wget https://github.com/microsoft/msquic/releases/download/v2.1.1/libmsquic-2.1.1-1.x86_64.rpm && \
+    rpm -i libmsquic-2.1.1-1.x86_64.rpm && \
```
wfurt, line 39: This seems to bypass signature checking.
```
rpm -i libmsquic-2.1.1-1.x86_64.rpm
warning: libmsquic-2.1.1-1.x86_64.rpm: Header V4 RSA/SHA256 Signature, key ID be1229cf: NOKEY
```
`zypper` can install packages directly via http without wget.
```
zypper install -y https://github.com/microsoft/msquic/releases/download/v2.1.1/libmsquic-2.1.1-1.x86_64.rpm
```

it fails on signature verification ... that is probably correct. My recommendation would be to add signing key

rzikm: I will probably need some hint about the signing key, I have never done that sort of thing before

wfurt:
```
bde607ad2ca8:/ # rpm --import https://packages.microsoft.com/keys/microsoft.asc
bde607ad2ca8:/ # zypper install -y https://github.com/microsoft/msquic/releases/download/v2.1.1/libmsquic-2.1.1-1.x86_64.rpm
Loading repository data...
Warning: Repository 'Main Update Repository' appears to be outdated. Consider using a different mirror or server.
Warning: Repository 'Update Repository (Non-Oss)' appears to be outdated. Consider using a different mirror or server.
Reading installed packages...
Resolving package dependencies...

The following NEW package is going to be installed:
  libmsquic

1 new package to install.
Overall download size: 6.1 MiB. Already cached: 0 B. After the operation, additional 22.0 MiB will be used.
Continue? [y/n/v/...? shows all options] (y): y
Retrieving package libmsquic-2.1.1-1.x86_64                                                                                                                                                                      (1/1),   6.1 MiB ( 22.0 MiB unpacked)

Checking for file conflicts: ...................................................................................................................................................................................................................[done]
(1/1) Installing: libmsquic-2.1.1-1.x86_64 .....................................................................................................................................................................................................[done]
```

I would add same download + verify checksum + import logic we have elsewhere. 
This may be paranoid but prevents cases when somebody sneaks evil bits on the server.

src/opensuse/15.2/helix/amd64/Dockerfile
```diff
@@ -28,12 +28,21 @@ RUN zypper ref && \
         python3-devel \
         python3-pip \
         sudo \
+        wget \
         unzip && \
```
mthalman, line 32:
```suggestion
        unzip \
        wget && \
```

wfurt (approved):
LGTM

PR #697, opened 2022-09-16
Rename build-rootfs to run-arcade-build-rootfs

am11:
When working with cross build, I found it a bit confusing that `build-rootfs.sh` in this repo is basically a wrapper script which invokes the script with same name from dotnet/arcade repository. The main confusion is the difference in their (positional) arguments.

The purpose of this PR is to disambiguate the names by renaming the script in this repo to <del>`invoke-arcade-rootfs.sh R`</del> `run-arcade-build-rootfs.sh`. <del>If this name sounds strange, I am open for suggestions. :)</del>

### DIFF
--- a/src/raspbian/10/helix/arm32v6/hooks/pre-build
+++ b/src/raspbian/10/helix/arm32v6/hooks/pre-build
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../../ubuntu/build-scripts/build-rootfs.sh raspbian-10 raspbian armv6 "" "" true
+$SCRIPTPATH/../../../../../ubuntu/build-scripts/run-arcade-build-rootfs.sh raspbian-10 raspbian armv6 "" "" true
--- a/src/ubuntu/18.04/cross/arm-alpine/hooks/pre-build
+++ b/src/ubuntu/18.04/cross/arm-alpine/hooks/pre-build
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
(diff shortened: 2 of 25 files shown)


src/raspbian/10/helix/arm32v6/hooks/pre-build
```diff
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../../ubuntu/build-scripts/build-rootfs.sh raspbian-10 raspbian armv6 "" "" true
+$SCRIPTPATH/../../../../../ubuntu/build-scripts/invoke-arcade-rootfs.sh raspbian-10 raspbian armv6 "" "" true
```
mthalman, line 5: How about this?
```suggestion
$SCRIPTPATH/../../../../../ubuntu/build-scripts/run-arcade-build-rootfs.sh raspbian-10 raspbian armv6 "" "" true
```

am11: Much better 👌

mthalman (approved):

am11:
It looks like when we change pre-build hook, we don't rebuild the image. In this case, if I had a typo, it went undetected by the CI. 🤔

mthalman:
Yep. I just realized that myself. To workaround this, you can make any kind of edit to one of the Dockerfiles, like adding a comment. That will trigger a cache miss and rebuild the image.

mthalman:
Then you can just revert the edit after verifying the build.

am11:
Yup, in this case, we would need to make change in every corresponding Dockerfile to be 100% sure. 🙈

BTW, I used these commands on macOS for mass replacement:

```sh
$ git grep -l build-rootfs.sh ':/*pre-build' |\
     xargs sed -i '' 's/build-rootfs.sh/run-arcade-build-rootfs.sh/g'
$ mv src/ubuntu/build-scripts/{,run-arcade-}build-rootfs.sh
```

So I will change just one.

am11:
It was all green, so I reverted the dummy change. This is ready.

PR #701, opened 2022-09-19
Use substitute base image

adiaaida:


mthalman:
The pipeline trigger path is not setup to reference the YAML path. So as things are today, it'd require manual queuing of builds. The other option is to add the path to the pipeline in the trigger.

eng/pipelines/steps/set-base-image-override-options.yml
```diff
@@ -0,0 +1,24 @@
...
+  baseOverrideRegistry: ''
+
+steps:
+# Configures Image Builder options to override the base image tags for Ubuntu as defined in the Dockerfiles
+# and instead use the corresponding image from the Ubuntu ACR.
+- powershell: |
+    # For official builds, we configure the regex and substitution
+    # to use a different registry. (e.g. "ubuntu:20.04" becomes "ubuntu.azurecr.io/ubuntu:20.04")".
+    # These tags are multi-arch tags so they'll work across all the architectures we provide.
+    $baseOverrideRegex = "(${{ parameters.dockerfileOs }}:.+)"
```
mthalman, line 14: This will likely require normalizing the references of the base images or making the regex more flexible. For example, there exist tags like the following:

* `alpine:3.13`
* `arm32v7/alpine:3.13`
* `arm64v8/alpine:3.13`
* `mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.13`

The first 3 we want to replace but not the last. This regex doesn't correctly handle the architecture-specific image names. It would replace `arm32v7/alpine:3.13` with `arm32v7/<override-registry>alpine3.13` when it should just be `<override-registry>alpine3.13`.

In order to keep the regex as simple as we can, I think it would be best to update the Dockerfiles to use a standard convention in referencing base image tags for these Linux distros. For example, `arm32v7/alpine:3.13` should just be `alpine:3.13`. The `3.13` tag is multi-arch and the build infrastructure now specifies the `--platform` option in all relevant commands so we are assured in getting the correct architecture. But any refactoring of tags like this should be done in a separate PR.

Also note that for CentOS and Fedora, these are coming from their own respective registries (quay.io and registry.fedoraproject.org), so the regex will need to account for those cases as well.

eng/pipelines/dotnet-buildtools-prereqs-alpine.yml
```diff
@@ -25,3 +25,10 @@ stages:
   parameters:
     internalProjectName: ${{ variables.internalProjectName }}
     publicProjectName: ${{ variables.publicProjectName }}
+    ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+      customCopyBaseImagesInitSteps:
+      - template: ../../../pipelines/steps/set-base-image-override-options.yml
+        parameters:
+          variableName: customCopyBaseImagesArgs
+          dockerfileOs: alpine
+          baseOverrideRegistry: $(overrideRegistry) # Comes from DotNet-Docker-Common variable group
```
mthalman, line 34: Has there been assessment of the override registry to ensure it has available all the image tags that these Dockerfiles require?

eng/pipelines/steps/set-base-image-override-options.yml
adiaaida: My intent here was to basically only match the first case, and ignore the last 3 (or more specifically, definitely the last one). I didn't realize the middle 2 were cases (but they aren't mirrored on the registry anyway). I'm happy to create a separate PR for the refactoring of the tags. If I want to ignore the final case, could I just change the regex to start with `^`? Then it wouldn't match at all?

adiaaida: Will do some thinking about the fedora/centos cases. They are mirrored, but they're mirrored from Dockerhub. Is that correct for us?

eng/pipelines/dotnet-buildtools-prereqs-alpine.yml
adiaaida: I skipped some of the OSes specifically because the override registry doesn't have them, but didn't look specifically into OS versions. I'll do another pass and make sure.

eng/pipelines/steps/set-base-image-override-options.yml
mthalman: Ok, the centos images from quay.io aren't mirrored so only the Docker Hub images are relevant.

For Fedora, I'm not sure if the ones from Docker Hub are ok. Probably? Those were originally added in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/483, so you could follow up with @wfurt on that.

adiaaida: @wfurt, is there anything special about the fedora base images coming from fedoraproject vs dockerhub? If there isn't, I may put out a PR to remove the registry from the fedora images so that we can use the override registry work.

adiaaida:
Passing runs with these changes:
Alpine: https://dev.azure.com/dnceng/internal/_build/results?buildId=1999426&view=results
Centos: https://dev.azure.com/dnceng/internal/_build/results?buildId=2000124&view=results
Ubuntu: https://dev.azure.com/dnceng/internal/_build/results?buildId=2000127&view=results

eng/pipelines/dotnet-buildtools-prereqs-alpine.yml
adiaaida: I have confirmed that all the cases that we would be replacing have the image tags in the registry

dkurepa:
I wonder if it makes sense to make the fact that we're doing some string manipulation more obvious, the way I understand it, in the future if someone was just reading the Dockerfiles, they'd have to know that we're doing this, or they'd have to dive into our yamls to find out

adiaaida:
This was the approach suggested by @mthalman, and is something we do in the dotnet-docker repo already. No matter what we do, it is going to come down to "they have to look at the yaml to realize that we are potentially changing the registry we get our base images from." Even if we were to use an ARG, that just suggests that you *could* change something, not that you are. It is something that we will likely want to internally document, but I think this is the right way to go about this.

adiaaida:
@mthalman do you have any additional feedback for this?

eng/pipelines/steps/set-base-image-override-options.yml
```diff
@@ -0,0 +1,24 @@
...
+  dockerfileOs: ''
+  baseOverrideRegistry: ''
+
+steps:
+# Configures Image Builder options to override the base image tags for Ubuntu as defined in the Dockerfiles
+# and instead use the corresponding image from the Ubuntu ACR.
+- powershell: |
+    # For official builds, we configure the regex and substitution
+    # to use a different registry. (e.g. "ubuntu:20.04" becomes "ubuntu.azurecr.io/ubuntu:20.04")".
+    # These tags are multi-arch tags so they'll work across all the architectures we provide.
```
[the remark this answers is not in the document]
mthalman: This doesn't look resolved. It's still referring to the Ubuntu ACR.

eng/pipelines/steps/set-base-image-override-options.yml
mthalman: What's the plan with this? I would expect that any normalization of base image tag usage in the Dockerfiles would be done before this PR gets merged.

eng/pipelines/steps/set-base-image-override-options.yml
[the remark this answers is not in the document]
adiaaida: Oh, I thought you were just referring to the multi-arch part of the comments (which I removed). The rest of the comment was to just have an example of what this was doing, but I'll change the example

adiaaida:
https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/708 is checked in

eng/pipelines/steps/set-base-image-override-options.yml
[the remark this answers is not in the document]
mthalman: @adiaaida - Were you still planning on updating this comment?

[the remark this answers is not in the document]
adiaaida: Apparently I forgot to push my latest changes. It has been updated

mthalman (approved):

PR #704, opened 2022-09-27
Only use floating tag suffix for staging

mthalman:
Updates the floating tag pattern to only use a suffix for staging tags. So this removes the use of a `-latest` suffix for the production tags.

For example, this means `mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.13-helix-arm64v8-latest` would now be `mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.13-helix-arm64v8`. But `mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.13-helix-arm64v8-staging` remains unchanged.

### DIFF
--- a/eng/pipelines/variables/common.yml
+++ b/eng/pipelines/variables/common.yml
@@ -13,15 +13,12 @@ variables:
   value: false
 - name: publicGitRepoUri
   value: https://github.com/dotnet/dotnet-buildtools-prereqs-docker
-- ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/production') }}:
+- ${{ if eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
   - name: floatingTagSuffix
-    value: latest
-- ${{ elseif eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
-  - name: floatingTagSuffix
-    value: staging
+    value: -staging
 - ${{ else }}:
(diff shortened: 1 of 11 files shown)


eng/pipelines/variables/common.yml
```diff
@@ -13,15 +13,12 @@ variables:
...
   - name: floatingTagSuffix
-    value: latest
-- ${{ elseif eq(variables['Build.SourceBranch'], 'refs/heads/main') }}:
-  - name: floatingTagSuffix
-    value: staging
+    value: -staging
 - ${{ else }}:
   - name: floatingTagSuffix
-    value: $(Build.SourceBranchName)
+    value: ""
```
adiaaida, line 21: Will this not make random other branches possibly produce suffix-less tags? Or do those not *actually* publish

mthalman: Technically that wouldn't happen because builds outside the context of main and production aren't allowed to publish. That's what this enforces:

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/1b3d0bd5290f5d31540f3a180f1afe8546ab8d72/eng/pipelines/variables/common.yml#L3-L5

However, on the off chance of that variable being updated to include another branch, it would be good to avoid having two branches stomp each other's tags. So I'll fix this up.

PR #707, opened 2022-09-28
Change all of the tags that are "isLocal" to use the -local suffix

adiaaida:
We wanted to be able to make images published by production to use a suffix-less tag, but that caused a conflict with the local tags for some dockerfiles, since they are already suffix-less. This change add -local to those local tags, and updates all of the dockerfiles that were using the local images.

### DIFF
--- a/src/ubuntu/20.04/cross/armv6/10/Dockerfile
+++ b/src/ubuntu/20.04/cross/armv6/10/Dockerfile
@@ -1,3 +1,3 @@
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-crossdeps
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-crossdeps-local-local
 
 ADD rootfs.armv6.tar crossrootfs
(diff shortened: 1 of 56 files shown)


src/ubuntu/20.04/cross/armv6/10/Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-crossdeps
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-crossdeps-local-local
```
mthalman, line 1:
```suggestion
FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-20.04-crossdeps-local
```

mthalman (approved):
Just one typo.

mthalman:
The following lines will need to be updated to include `-local` at the end of the tag name: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/facc263cdd5b3b774b2835a8658b350eca58a8fc/src/ubuntu/build-scripts/run-arcade-build-rootfs.sh#L16

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/facc263cdd5b3b774b2835a8658b350eca58a8fc/src/centos/7/mlnet/Dockerfile#L1

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/facc263cdd5b3b774b2835a8658b350eca58a8fc/src/ubuntu/22.04/mlnet/Dockerfile#L1

mthalman (approved):

PR #721, opened 2022-10-11
Remove eol debian dockerfiles and update base images

adiaaida:
This change removed Debian:9 dockerfiles, which was end-of-life on July 1, 2022. It also updates debian:11 images to get the debian:bullseye base image so that the mirror works. Finally, it updates the debian iot dockerfile to directly reference bullseye so that the mirror works.


src/debian/iot/arm32v7/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:latest
+FROM debian:bullseye
```
mthalman, line 1: Since this is now targeting a specific version, I think it would make sense to move the Dockerfile to the version-specific directory: src/Debian/11/iot/arm32v7/Dockerfile. The manifest.json would need to reflect that path change as well.

mthalman (approved):

src/debian/iot/arm32v7/Dockerfile
adiaaida: Done

mthalman (approved):

PR #731, opened 2022-10-31
Add registry override to the -all pipeline.

michellemcdaniel:


mthalman (approved):
Looks good. Nice that it wasn't anything too complicated.

michellemcdaniel:
Testing this here https://dev.azure.com/dnceng/internal/_build/results?buildId=2033461&view=results. Local powershell suggested this would work, but need to make sure

Issue #328, opened 2020-06-10
Public builds have CI enabled


mthalman:
The OS-specific pipelines (e.g. https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/master/eng/pipelines/dotnet-buildtools-prereqs-alpine.yml) are reused for both the internal and public builds.  Because of that, the public build also gets triggered with CI when a commit occurs.  This is completely unnecessary since we've already validated the change in the PR and the internal build is the one that actually publishes the image.   This ends up just wasting build resources for an unnecessary build.

PR #732, opened 2022-10-31
Use all pipeline for production

michellemcdaniel:
For the production branch, we want to build everything, both on CI and PR. This change modifies the pipelines to run the "-all" pipeline for production instead of the split out pipelines.

### DIFF
--- a/eng/pipelines/dotnet-buildtools-prereqs-all-pr.yml
+++ b/eng/pipelines/dotnet-buildtools-prereqs-all-pr.yml
@@ -1,26 +1,25 @@
 trigger: none
-pr:
+pr: 
   branches:
     include:
-    - main
-  paths:
-    include:
-    - src/debian/*
+    - production
 
 resources:
   repositories:
-  - repository: PublicVersionsRepo
+  - repository: InternalVersionsRepo
     type: github
-    endpoint: public
+    endpoint: dotnet
     name: dotnet/versions
 
 variables:
 - template: variables/common.yml
-- name: imageBuilder.pathArgs
-  value: --path 'src/debian/*'
 
 stages:
 - template: ../common/templates/stages/dotnet/build-test-publish-repo.yml
   parameters:
     internalProjectName: ${{ variables.internalProjectName }}
     publicProjectName: ${{ variables.publicProjectName }}
+    linuxAmdBuildJobTimeout: 210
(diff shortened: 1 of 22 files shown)


eng/pipelines/dotnet-buildtools-prereqs-all-pr.yml
```diff
@@ -0,0 +1,25 @@
+trigger: none
```
riarenas, line 1: This file seems to be almost exactly the same as [eng/pipelines/dotnet-buildtools-prereqs-all.yml](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/732/files#diff-b18af6e0e549a23ecd4af3675df1948a2f64ff304c74dcaa9a23d7980e01f068). That pipeline even has conditions for steps that should only run on internal project, which leads me to believe it should be fine to run in the public. Would it make sense to add a PR trigger to that other pipeline and not add a new yaml?

michellemcdaniel: This is the current standard practice in this repo. Just trying to match everything else.

riarenas (approved):

eng/pipelines/dotnet-buildtools-prereqs-all-pr.yml
riarenas: If the duplication is desirable, this change looks good to me I guess. But having a 1:1 duplicated code pipeline doesn't sit all that well with me.

riarenas (approved):

eng/pipelines/dotnet-buildtools-prereqs-all-pr.yml
MichaelSimons: The reason for this is related to the triggers and trying to avoid wasting compute.  See https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/328 for the background.

riarenas: Hmm yeah, duplicating the YAML files seems to be a good workaround for that. The only other alternative I can think of is to change the triggers on the public pipeline through the pipeline UI, but that also has its drawbacks.

mthalman (approved):

PR #737, opened 2022-11-10
Use powershell 7.2.5 on Alpine 3.15 helix container

michellemcdaniel:
We were using "whatever is most recent" but that currently has broken permissions on the executables. We shouldn't float versions of dependencies anyway, so this change downgrades to powershell 7.2.5, which matches other alpine containers.

Addresses https://github.com/dotnet/arcade/issues/11579

### DIFF
--- a/src/alpine/3.15/helix/amd64/Dockerfile
+++ b/src/alpine/3.15/helix/amd64/Dockerfile
@@ -19,8 +19,7 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
 RUN cd /tmp && \
     mkdir pwsh && \
     cd pwsh && \
-    powerShellVersionTag=$(curl --silent https://api.github.com/repos/PowerShell/PowerShell/releases/latest | jq -r .tag_name) && \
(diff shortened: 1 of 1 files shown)


ChadNedzlek (approved):
Wow. They really went out of their way to make this broken, didn't they...

mthalman (approved):
> We shouldn't float versions of dependencies anyway

I understand needing to make this change as a workaround. But as far as a general policy of not using a floating version, that's not consistent. As an example, nearly all Linux package dependencies are installed with whatever is latest, not pinned to a specific version. PowerShell doesn't seem special in that regard. Is there a plan in place to ensure pinned versions are updated periodically?

michellemcdaniel:
Not floating versions on dependencies we install from the internet (like powershell) is actually consistent with our policy in helix. While we do float versions from package managers, everything that we can't get from package managers has to be manually moved forward because we mirror them into our own storage container. We do move versions of dependencies forward, but it is a manual process in helix for these types of dependencies. /cc @MattGal @Chrisboh

MichaelSimons:
> As an example, nearly all Linux package dependencies are installed with whatever is latest, not pinned to a specific version.

While that is true, package managers typically don't update major releases for a specific OS version.

PR #738, opened 2022-11-14
Delete unused debian-iot-arm32v7 Dockerfile

mthalman:
This Dockerfile is failing to build with the error:

```
--2022-11-14 19:56:00--  https://raw.githubusercontent.com/adafruit/Raspberry-Pi-Installer-Scripts/master/libgpiod.sh
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.109.133, 185.199.110.133, 185.199.111.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.109.133|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2022-11-14 19:56:00 ERROR 404: Not Found.
```

This is due to this change: https://github.com/adafruit/Raspberry-Pi-Installer-Scripts/pull/234. The script has moved to a different location.

Rather than simply updating the Dockerfile for the new path, I checked to see whether this image tag for this Dockerfile is being used at all. I can find no usage whatsoever. So I'm deleting the Dockerfile.

@joperezr - You had originally added this Dockerfile so please confirm the deletion is ok.

PR #739, opened 2022-11-15
[Rollout] Production rollout 2022-11-16

michellemcdaniel:


mthalman:
This is going to run into the issue from https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/738. Still waiting for approval on that from @joperezr.

michellemcdaniel:
Thanks for the heads up Matt

riarenas (approved):

mthalman:
It looks like the build for https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/production/src/centos/stream9/Dockerfile has failed:

```
Error: 
 Problem 1: package gcc-toolset-12-gcc-c++-12.1.1-3.2.el9.x86_64 requires gcc-toolset-12-runtime, but none of the providers can be installed
  - package clang-libs-15.0.1-2.el9.x86_64 requires gcc-toolset-12-gcc-c++, but none of the providers can be installed
  - package gcc-toolset-12-runtime-12.0-5.el9.x86_64 requires /usr/sbin/semanage, but none of the providers can be installed
  - package clang-15.0.1-2.el9.x86_64 requires libclang-cpp.so.15()(64bit), but none of the providers can be installed
  - package clang-15.0.1-2.el9.x86_64 requires clang-libs(x86-64) = 15.0.1-2.el9, but none of the providers can be installed
  - package policycoreutils-python-utils-3.4-4.el9.noarch requires python3-policycoreutils = 3.4-4.el9, but none of the providers can be installed
  - cannot install the best candidate for the job
  - nothing provides python3-libsemanage >= 3.4-1 needed by python3-policycoreutils-3.4-4.el9.noarch
 Problem 2: package clang-libs-15.0.1-2.el9.x86_64 requires gcc-toolset-12-gcc-c++, but none of the providers can be installed
  - package gcc-toolset-12-gcc-c++-12.1.1-3.2.el9.x86_64 requires gcc-toolset-12-runtime, but none of the providers can be installed
  - package lldb-15.0.1-2.el9.x86_64 requires libclang-cpp.so.15()(64bit), but none of the providers can be installed
  - package gcc-toolset-12-runtime-12.0-5.el9.x86_64 requires /usr/sbin/semanage, but none of the providers can be installed
  - package lldb-devel-15.0.1-2.el9.x86_64 requires liblldb.so.15()(64bit), but none of the providers can be installed
  - package lldb-devel-15.0.1-2.el9.x86_64 requires liblldbIntelFeatures.so.15()(64bit), but none of the providers can be installed
  - package lldb-devel-15.0.1-2.el9.x86_64 requires lldb(x86-64) = 15.0.1-2.el9, but none of the providers can be installed
  - package policycoreutils-python-utils-3.4-4.el9.noarch requires python3-policycoreutils = 3.4-4.el9, but none of the providers can be installed
  - cannot install the best candidate for the job
  - nothing provides python3-libsemanage >= 3.4-1 needed by python3-policycoreutils-3.4-4.el9.noarch
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

The PR build didn't attempt to build that Dockerfile because the Dockerfile hadn't been changed. And I'm very confused why it's attempting to build it in the internal build as well. The image caching logic that we have should skip it in the same way that the PR build. But for whatever reason, it thinks the commit of the Dockerfile in the branch is different somehow. It thinks the SHA of the Dockerfile is https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/3f90b63cb7c0b31b0d1c0c50793ba954916d7d33/src/centos/stream9/Dockerfile when it's expected to be https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/94fc78a06814446fcf85b34fd4334dfbb1a99a38/src/centos/stream9/Dockerfile. Those SHAs are different and valid but take you to the same commit. I don't understand how that happens. Because they have different SHAs, it causes this Dockerfile to be built. And that's exposed this underlying package installation issue which needs to be investigated.

michellemcdaniel:
Strange, I can't repro the failure locally

PR #741, opened 2022-11-18
Pin cmake version on 18.04 cross containers to 3.24.1

michellemcdaniel:
Runtime requires 3.24.1 on the freeBSD cross container, but the default/latest version provided by kitware was recently upgraded to 3.25.0, which broke runtime's FreeBSD legs here: https://github.com/dotnet/runtime/issues/78522.

This addresses https://github.com/dotnet/arcade/issues/11676.

### DIFF
--- a/src/ubuntu/18.04/crossdeps/Dockerfile
+++ b/src/ubuntu/18.04/crossdeps/Dockerfile
@@ -15,7 +15,8 @@ RUN apt-get update \
         binfmt-support \
         binutils-arm-linux-gnueabihf \
         build-essential \
-        cmake \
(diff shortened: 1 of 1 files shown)


jkotas:
Fix to make dotnet/runtime build with cmake 3.25 was merged this morning. Backport to servicing trees is in progress. I do not think that this change is needed.

We may need better way to deal with these types of breaking updates. I took several people several hours to analyze the build break and trace it down to the silent cmake version update.

michellemcdaniel:
Thanks Jan. Will abandon this PR

PR #752, opened 2022-12-29
Update testrunner Dockerfile to Debian Bullseye

mthalman:
Makes several improvements to the testrunner Dockerfile:
* Upgrade from Debian Buster to Debian Bullseye
* A single Dockerfile that has multi-arch support
* Uses latest version of PowerShell instead of pinning it and getting out-of-date
* Defines a floating multi-arch tag
* ~~Replaces usage of `UniqueId` variable for tag names, which mapped to the build ID, to use the same pattern as all the other Dockerfiles of timestamp and commit SHA~~


MichaelSimons:
The description indicates the `UniqueId` usage was removed but it is still present.  It looks like this was explicitly added back. What's the backstory here?

mthalman:
> The description indicates the `UniqueId` usage was removed but it is still present. It looks like this was explicitly added back. What's the backstory here?

It's because a `DockerfileGitCommitSha` variable cannot be used for a shared tag. A shared tag is directly associated to a single Dockerfile so there's no associated git commit. For consistency, it made sense to go back to `UniqueId` for both the simple tags and shared tags. I've updated the original message to remove the info about getting rid of that variable.

MichaelSimons (approved):

PR #754, opened 2023-01-04
Add arm64 cross-build images targeting Ubuntu 16.04

sbomer:
@janvorli @MichaelSimons PTAL. I can't seem to add reviewers directly.

This is adding an Ubuntu 20.04 image with an arm64 Ubuntu 16.04 rootfs that we can use to build against a lower glibc version than we currently support. The plan is to use this image at least for the .NET 7 part of https://github.com/dotnet/runtime/issues/69361.

Locally, I've used these steps to build a container that can build coreclr with the `--pgoinstrument` argument. Without the profile library build, the build fails with
```
ld.lld: error: cannot open /rootfs/usr/lib/llvm-9/lib/clang/9.0.1/lib/linux/libclang_rt.profile-aarch64.a: No such file or directory
```
Building this library and copying it into the rootfs fixes this.

@janvorli I would appreciate if you could also review the logs here to make sure everything looks ok: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/137325/logs/467. I started with instructions at https://llvm.org/docs/HowToCrossCompileLLVM.html and https://llvm.org/docs/HowToCrossCompileBuiltinsOnArm.html, then had to play around with the arguments and use a mix of CMAKE variables and flags like `--sysroot` (`-DCMAKE_SYSROOT` was not working as I expected). The logs show that `/usr/lib/llvm-9/lib/clang/9.0.1/include` is still being used as an include path - I am not sure if this is a problem, so I'd appreciate your eyes on this.

### DIFF
--- a/src/ubuntu/20.04/cross/arm64/Dockerfile
+++ b/src/ubuntu/20.04/cross/arm64/Dockerfile
@@ -24,7 +24,7 @@ RUN mkdir compiler-rt-build && \
     cd compiler-rt-build && \
     \
     TARGET_TRIPLE=aarch64-linux-gnu \
-    BUILD_CXX_FLAGS="-v --sysroot=$ROOTFS_DIR" && \
+    BUILD_FLAGS="-v --sysroot=$ROOTFS_DIR" && \
     \
     cmake ../compiler-rt-9.0.1.src \
         -DCOMPILER_RT_BUILD_PROFILE=ON \
@@ -41,8 +41,8 @@ RUN mkdir compiler-rt-build && \
         -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON \
         -DLLVM_CONFIG_PATH=/usr/bin/llvm-config-9 \
         \
-        -DCMAKE_C_FLAGS="${BUILD_CXX_FLAGS}" \
-        -DCMAKE_CXX_FLAGS="${BUILD_CXX_FLAGS}"
+        -DCMAKE_C_FLAGS="${BUILD_FLAGS}" \
+        -DCMAKE_CXX_FLAGS="${BUILD_FLAGS}"
 
 RUN cd compiler-rt-build && make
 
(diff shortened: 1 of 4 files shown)


janvorli:
> The logs show that /usr/lib/llvm-9/lib/clang/9.0.1/include is still being used as an include path 

The same happens for .NET crossbuild, so it is ok. It seems like these are compiler internal headers for intrinsics and few other things.

mthalman (approved):

mthalman (approved):

src/ubuntu/20.04/cross/arm64/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+        \
+        -DCMAKE_C_COMPILER=/usr/bin/clang-9 \
+        -DCMAKE_CXX_COMPILER=/usr/bin/clang++-9 \
+        -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" \
+        -DCMAKE_C_COMPILER_TARGET="${TARGET_TRIPLE}" \
+        -DCMAKE_CXX_COMPILER_TARGET="${TARGET_TRIPLE}" \
+        -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON \
+        -DLLVM_CONFIG_PATH=/usr/bin/llvm-config-9 \
+        \
+        -DCMAKE_C_FLAGS="${BUILD_CXX_FLAGS}" \
```
janvorli, line 44: Ah, I am sorry, I have somehow missed the fact that the env var is defined locally and not passed in from the outside. It then doesn't matter what the name of the env var is. Originally I meant using BUILD_C_FLAGS for the CMAKE_C_FLAGS and BUILD_CXX_FLAGS for CMAKE_CXX_FLAGS since in general they sometimes tend to be different.

janvorli (approved):

Issue #755, opened 2023-01-04
Unable to install v8 engine


mthalman:
Builds of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile are failing with this error:

```
Step 9/9 : RUN v8 -e "console.log(version());quit();"
 ---> Running in 22a3c102a8af
/bin/sh: 1: v8: not found
The command '/bin/sh -c v8 -e "console.log(version());quit();"' returned a non-zero code: 127
```

This is really just a symptom of the silent failure to install the v8 engine from an earlier step:

```
Step 7/9 : RUN jsvu --os=linux64 --engines=v8
 ---> Running in 34a88a1d2cf0
📦 jsvu v2.0.1 — the JavaScript engine Version Updater 📦
✔ Read OS from config: linux64
✔ Read engines from config: v8
❯ Finding the latest V8 version…
✔ Found latest V8 version: v11.1.121.
❯ Predicting URL…
✔ URL: https://storage.googleapis.com/chromium-v8/official/canary/v8-linux64-rel-11.1.121.zip
❯ Downloading…


✔ Download completed.
❯ Extracting…
Installing library to ~/.jsvu/engines/v8/icudtl.dat…
Installing library to ~/.jsvu/engines/v8/snapshot_blob.bin…
Installing binary to ~/.jsvu/engines/v8/v8…
Installing wrapper script to ~/.jsvu/bin/v8…
(node:6) UnhandledPromiseRejectionWarning: TypeError: generateScript(...).trimStart(...).replaceAll is not a function
    at Installer.installScript (/usr/local/lib/node_modules/jsvu/shared/installer.js:139:5)
    at Promise (/usr/local/lib/node_modules/jsvu/engines/v8/extract.js:52:14)
(node:6) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:6) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

It's not clear why the JS error is occurring, whether it's a bug in jsvu or not. This was previously working when version 11.1.90 was the latest v8 version. Now it's failing on 11.1.121.

[...]

PR #756, opened 2023-01-05
Update the node and npm so that newer jsvu and v8 work

radekdoulik:
That should fix https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/755 and unblock builds. Next step will be to consider and move to newer Ubuntu 22.04.

### DIFF
--- a/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
+++ b/src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
@@ -22,9 +22,10 @@ RUN apt-get clean && \
 USER helixbot
 
 # update node, which is required for newer npm, jsvu and v8
+ENV NODE_VERSION 18.12.1
 RUN cd ~ && \
-    curl -LO https://nodejs.org/dist/v18.12.1/node-v18.12.1-linux-x64.tar.xz && \
-    tar xf node-v18.12.1-linux-x64.tar.xz
+    curl -LO https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz && \
+    tar xf node-v${NODE_VERSION}-linux-x64.tar.xz
 
 USER root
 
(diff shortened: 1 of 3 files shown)


src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
```diff
@@ -11,18 +11,30 @@ RUN apt-get clean && \
...
      && rm -rf /var/lib/apt/lists/* \
      \
      && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 
 
+USER helixbot
+
+# update node, which is required for newer npm, jsvu and v8
+RUN cd ~ && \
+    curl -LO https://nodejs.org/dist/v18.12.1/node-v18.12.1-linux-x64.tar.xz && \
```
mthalman, line 26: For convenience, it might be better to define an `ENV` variable for the Node version so it can be updated in one location. Currently, there's 3 different spots that need to be updated if the version changes.

mthalman (commented):
I also realized I didn't update the jsvu path here:

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/f65e83005a3bb3ef0261e0f04e24c346caf1ca0d/src/windowsservercore/ltsc2019/webassembly/Dockerfile#L79

It should be updated to `.jsvu\bin`. Could you update that in this PR?

src/ubuntu/20.04/helix/wasm/amd64/Dockerfile
```diff
@@ -11,18 +11,31 @@ RUN apt-get clean && \
...
+
+# update node, which is required for newer npm, jsvu and v8
+ENV NODE_VERSION 18.12.1
+RUN cd ~ && \
+    curl -LO https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz && \
+    tar xf node-v${NODE_VERSION}-linux-x64.tar.xz
+
+USER root
+
+ENV PATH=/home/helixbot/node-v18.12.1-linux-x64/bin:$PATH
```
mthalman, line 32:
```suggestion
ENV PATH=/home/helixbot/node-v${NODE_VERSION}-linux-x64/bin:$PATH
```

mthalman (approved):

PR #762, opened 2023-01-10
[Rollout] Production rollout - 2023-1-11

riarenas:
latest main build: https://dnceng.visualstudio.com/internal/_build/results?buildId=2078446&view=results

### DIFF
--- a/eng/common/Get-ImageBuilder.ps1
+++ b/eng/common/Get-ImageBuilder.ps1
@@ -1,15 +1,13 @@
 #!/usr/bin/env pwsh
 
 # Load common image names
-Get-Content $PSScriptRoot/templates/variables/docker-images.yml |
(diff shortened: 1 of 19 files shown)


am11:
Do the production tags list get published somewhere? I think the README in production branch (https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/production/README.md) should link different `Latest tags` with only production tags, so we know when were they last published.

mthalman:
> Do the production tags list get published somewhere? I think the README in production branch (https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/production/README.md) should link different `Latest tags` with only production tags, so we know when were they last published.

Yes, they get published to https://github.com/dotnet/versions/blob/main/build-info/docker/image-info.dotnet-dotnet-buildtools-prereqs-docker-production.json. I'll update the readme to include links for both staging and production.

PR #766, opened 2023-01-12
Use alpine 3.14 in alpine-arm rootfs

am11:
This should fix https://github.com/dotnet/arcade/issues/12189.

### DIFF
--- a/src/ubuntu/18.04/cross/arm-alpine/hooks/pre-build
+++ b/src/ubuntu/18.04/cross/arm-alpine/hooks/pre-build
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../build-scripts/run-arcade-build-rootfs.sh ubuntu-18.04 alpine arm lldb3.9
+$SCRIPTPATH/../../../../build-scripts/run-arcade-build-rootfs.sh ubuntu-18.04 alpine3.14 arm lldb3.9
--- a/src/ubuntu/22.04/cross/arm-alpine/hooks/pre-build
+++ b/src/ubuntu/22.04/cross/arm-alpine/hooks/pre-build
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../build-scripts/run-arcade-build-rootfs.sh ubuntu-22.04 alpine arm
+$SCRIPTPATH/../../../../build-scripts/run-arcade-build-rootfs.sh ubuntu-22.04 alpine3.14 arm


mthalman (commented):
Shouldn't the fix be made in arcade? Shouldn't the default of `alpine` map to Alpine 3.14? Why would we continue to support Alpine 3.13 in the script when it's EOL?

michellemcdaniel:
Just a heads up that this will mean these changes will not flow to the latest images until next week, assuming we do not take this as a hotfix (ie, everything that rolled out this week is blocked by this change. No new images are flowing).

As a stop gap, I'd also like to see us hard code commit hashes to the arcade clone, so this issue does not happen again.

riarenas:
if we're going to take this as a hotfix there needs to be a compelling business reason why this can't wait until the normal weekly deployment train. I would also like to understand why we want to keep using EOL'd OSes as the default in the arcade script.

am11:
Alpine rootfs is currently based on Alpine 3.13 and for Ubuntu on Xenial (16.04), both are EOL OS.

rootfs is not a complete OS, so EOL of OS which rootfs setup is based on doesn't matter as much as it does for host OS of the docker image. Nothing from rootfs is actually copied to the shipping packages from dotnet/runtime. We only link the binaries against symbols of libs in rootfs (most importantly libc). When user installs the .NET runtime pack on any version of musl-libc or glibc based OS whose version >= rootfs base OS version, it just works (by the virtue of forward-compatibility of libc). The compatibility of binaries, which we ship, is more important than rootfs' base OS EOL'ness from which we don't copy anything in our packages.

cc @jkotas, @janvorli

jkotas:
I agree with @am11 explanation of the problem space.

@riarenas @mthalman Do the build images produced by this change affect .NET 8 main only or are they going to be used for servicing as well?

This is good change to take for .NET 8 main, but it is somewhat risky change to take for servicing.

riarenas:
> I agree with @am11 explanation of the problem space.
> 
> @riarenas @mthalman Do the build images produced by this change affect .NET 8 main only or are they going to be used for servicing as well?
> 
> This is good change to take for .NET 8 main, but it is somewhat risky change to take for servicing.

@michellemcdaniel do you know the answer to this?

michellemcdaniel:
Servicing branches, in most cases, have not been migrated to latest tags, so they would not be getting these changes automatically.

src/ubuntu/18.04/cross/arm-alpine/hooks/pre-build
```diff
@@ -2,4 +2,4 @@
 
 SCRIPT=$(readlink -f "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
-$SCRIPTPATH/../../../../build-scripts/run-arcade-build-rootfs.sh ubuntu-18.04 alpine arm lldb3.9
+$SCRIPTPATH/../../../../build-scripts/run-arcade-build-rootfs.sh ubuntu-18.04 alpine3.14 arm lldb3.9
```
jkotas, line 5: How was `alpine` mapped to actual alpine version before this change?

am11: The default is 3.13: https://github.com/dotnet/arcade/blob/12b42d0d77948889890ef2f35af8c46438d3f2d2/eng/common/cross/build-rootfs.sh#L258-L262.

I think we should also omit `numactl-dev` from alpine 3.13 there (as we omit `libunwind8-dev` from riscv64), so if anyone else is using rootfs with 3.13 their installation don't break.

jkotas: ok

jkotas (approved):
LGTM

@mthalman @michellemcdaniel @riarenas Any concerns?

janvorli (approved):
LGTM, thank you!

michellemcdaniel:
Seems fine to me

mthalman (approved):
Yep, looks good.

PR #757, opened 2023-01-06
Bump v8 version to latest stable - 10.8.168 for wasm ubuntu 18.04 image

radical:

PR #772, opened 2023-01-18
image for wasi 22.04

mkhamoyan:
Contributes to https://github.com/dotnet/runtime/issues/65895

Image with

emscripten
wasi sdk
wasmtime

based on ubuntu 22.04

### DIFF
--- a/src/ubuntu/22.04/webassembly/Dockerfile
+++ b/src/ubuntu/22.04/webassembly/Dockerfile
@@ -40,7 +40,7 @@ RUN mkdir ${EMSCRIPTEN_PATH} \
 # Install V8 Engine
 SHELL ["/bin/bash", "-c"]
 
-ENV V8_VERSION=8.5.183
+ENV V8_VERSION=10.8.168
 RUN curl -sSL https://netcorenativeassets.blob.core.windows.net/resource-packages/external/linux/chromium-v8/v8-linux64-rel-${V8_VERSION}.zip -o ./v8.zip \
     && unzip ./v8.zip -d /usr/local/v8 \
     && echo $'#!/usr/bin/env bash\n\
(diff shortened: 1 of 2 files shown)


src/ubuntu/22.04/webassembly/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    && git clone https://github.com/emscripten-core/emsdk.git ${EMSDK_PATH} \
+    && cd ${EMSDK_PATH} \
+    && ./emsdk install ${EMSCRIPTEN_VERSION}-upstream \
+    && ./emsdk activate ${EMSCRIPTEN_VERSION}-upstream \
+    && chmod -R 777 ${EMSCRIPTEN_PATH}
+
+# Install V8 Engine
+SHELL ["/bin/bash", "-c"]
+
+ENV V8_VERSION=8.5.183
```
pavelsavara, line 43: Latest V8 is 11.1.218 but I don't know if we have specific preference. @radical @maraf

radical: Maybe the latest stable? like https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/757

mkhamoyan: Do I need to update also for 20.04/webassembly/Dockerfile?

mthalman (commented):
Is this meant to replace the following:

https://github.com/dotnet/runtime/blob/4544bddc57c0a602c557b450ed23bd2e550438c0/eng/pipelines/common/templates/pipeline-with-resources.yml#L63-L64

Or where will it be used? If it's going to be a replacement, it would be better to also remove the 20.04 version.

src/ubuntu/22.04/webassembly/Dockerfile
pavelsavara: That would be good. are there yet another images?

mkhamoyan:
Closing this as we will not this for now

PR #777, opened 2023-01-23
Add support for Bookworm

richlander:
These Dockerfiles are (mostly) a copy of the Debian bullseye variants. I've made two changes:

- Updated `FROM` statement to `bookworm`
- Removed `msquic` which is apparently not compatible with OpenSSL 3 (which comes with Bookworm).

The last `RUN` statement fails with the following error. I'll need some help with that. I validated that the Debian 11 variant works and it does.

```bash
 => ERROR [5/5] RUN python -m virtualenv --no-site-packages /home/helixbo  0.6s
------
 > [5/5] RUN python -m virtualenv --no-site-packages /home/helixbot/.vsts-env:
#8 0.574 Traceback (most recent call last):
#8 0.574   File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
#8 0.574     return _run_code(code, main_globals, None,
#8 0.574   File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
#8 0.574     exec(code, run_globals)
#8 0.574   File "/usr/local/lib/python3.10/dist-packages/virtualenv.py", line 2580, in <module>
#8 0.575     main()
#8 0.575   File "/usr/local/lib/python3.10/dist-packages/virtualenv.py", line 821, in main
#8 0.575     create_environment(
#8 0.575   File "/usr/local/lib/python3.10/dist-packages/virtualenv.py", line 1106, in create_environment
#8 0.575     install_python(home_dir, lib_dir, inc_dir, bin_dir, site_packages=site_packages, clear=clear, symlink=symlink)
#8 0.575   File "/usr/local/lib/python3.10/dist-packages/virtualenv.py", line 1666, in install_python
#8 0.575     fix_local_scheme(home_dir, symlink)
#8 0.575   File "/usr/local/lib/python3.10/dist-packages/virtualenv.py", line 1752, in fix_local_scheme
#8 0.575     if sysconfig._get_default_scheme() == "posix_local":
#8 0.575 AttributeError: module 'sysconfig' has no attribute '_get_default_scheme'. Did you mean: 'get_default_scheme'?
#8 0.575 Using base prefix '/usr'
#8 0.575 New python executable in /home/helixbot/.vsts-env/bin/python
------
executor failed running [/bin/sh -c python -m virtualenv --no-site-packages /home/helixbot/.vsts-env]: exit code: 1
```

### DIFF
--- a/src/debian/12/helix/amd64/Dockerfile
+++ b/src/debian/12/helix/amd64/Dockerfile
@@ -1,0 +1,62 @@
+FROM debian:bookworm
+
+# Install Helix Dependencies
+
+RUN apt-get update && \
+    apt-get install -y \
+        autoconf \
+        automake \
+        at \
+        build-essential \
+        curl \
+        gcc \
+        gdb \
+        git \
+        iputils-ping \
+        libcurl4 \
+        libffi-dev \
+        libgdiplus \
+        libicu-dev \
+        libssl-dev \
+        libtool \
+        libunwind8 \
+        locales \
+        locales-all \
+        python3-dev \
+        python3-pip \
+        software-properties-common \
+        sudo \
+        tzdata \
+        unzip \
+    && rm -rf /var/lib/apt/lists/* \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+
+ENV LANG=en_US.utf8
+
+RUN ln -sf /usr/bin/python3 /usr/bin/python && \
+    curl https://bootstrap.pypa.io/get-pip.py -o ./get-pip.py --fail --silent --show-error && \
+    python ./get-pip.py && rm ./get-pip.py && \
+    python -m pip install --upgrade pip==20.2 && \
+    python -m pip install virtualenv==16.6.0 && \
+    pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
+    pip install ./helix_scripts-*-py3-none-any.whl
+
+# Add MsQuic
+RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
+    echo 2cfd20a306b2fa5e25522d78f2ef50a1f429d35fd30bd983e2ebffc2b80944fa microsoft.asc| sha256sum --check - && \
+    apt-key add microsoft.asc && \
+    rm microsoft.asc && \
+    apt-add-repository https://packages.microsoft.com/debian/11/prod && \
+    apt-get update && \
+    apt-get install -y libmsquic && \
+    rm -rf /var/lib/apt/lists/*
+
+# Create helixbot user and give rights to sudo without password
+# additionally, preinstall the virtualenv packages used for VSTS reporting to save time
+RUN /usr/sbin/adduser --disabled-password --gecos '' --uid 1000 --shell /bin/bash --ingroup adm helixbot && \
+    chmod 755 /root && \
+    echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers
+
+USER helixbot
+
+RUN python -m virtualenv --no-site-packages /home/helixbot/.vsts-env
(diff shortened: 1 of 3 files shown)


src/debian/manifest.json
```diff
@@ -175,6 +175,31 @@
...
+              "os": "linux",
+              "osVersion": "bookworm",
+              "tags": {
+                "debian-12-helix-amd64-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "debian-12-helix-amd64$(FloatingTagSuffix)": {}
+              }
+            }
+          ]
+        },
+        {
```
michellemcdaniel, line 193: This should be inside the above platforms array

src/debian/manifest.json
```diff
@@ -175,6 +175,31 @@
...
+          "architecture": "arm64",
+          "dockerfile": "src/debian/11/helix/arm64v8",
```
michellemcdaniel, line 195: this looks like it should be debian/12

src/debian/manifest.json
```diff
@@ -175,6 +175,31 @@
...
+          "os": "linux",
+          "osVersion": "bullseye",
```
michellemcdaniel, line 197: and this looks like it should be bookworm

src/debian/manifest.json
richlander: It wasn't obvious to me how these arrays are intended to work. Can you help with understand that?

michellemcdaniel: I think that @mthalman would do a better job, especially as it seems I still got my suggestion wrong. I think each image definition gets its own outer platform []. but tbh, not sure.

richlander: Got it. I pushed a change for that.

src/debian/12/helix/amd64/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+# Install Helix Dependencies
+
+RUN apt-get update && \
+    apt-get install -y \
+        autoconf \
+        automake \
+        at \
+        build-essential \
+        curl \
```
mthalman, line 11: I don't see cmake listed here. It's listed in the Arm64 version and also in the earlier Bullseye version.

wfurt: I don't think `cmake` its needed for Helix. (nor is gcc etc.. e.g. we generally don't need to build to run tests) 
We may make the images smaller if we don't intend them as build images.

richlander: Are you sure, Matt. I copied the 11 version of the same thing. I think you are looking at the wrong Dockerfile. Make sure the paths match.

mthalman: Ah, sorry. I was looking at this: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/79dc2e51d29d9c755b00c216699160d096726d9c/src/debian/11/amd64/Dockerfile#L11

I didn't notice that your changes are for the Helix Dockerfile. Never mind. Carry on.

mthalman (approved):

src/debian/12/helix/amd64/Dockerfile
richlander: Right. Easy to do.

richlander:
I propose we merge this PR and address the MsQuic topic as a follow-up. Fair?

mthalman:
> I propose we merge this PR and address the MsQuic topic as a follow-up. Fair?

I'll defer to @wfurt. Can progress be made consuming this image without MsQuic?

wtgodbe:
> I'll defer to @wfurt. Can progress be made consuming this image without MsQuic?

Yes, but we'd have to disable a bunch of HTTP3 tests in aspnetcore on this image. I think for now we should move ahead without msquic, and start getting coverage by doing manual runs on the image, with the expectation that HTTP3 tests will fail there. I don't think we'd want to switch from debian11 to 12 _entirely_ until msquic is available, but we could test on both & disable the HTTP3 tests just on 12 in the interim.

PR #780, opened 2023-01-26
[wasm] Bump emscripten to 3.1.30

radekdoulik:


radekdoulik:
@mthalman do we have a way to exclude builds of these images from staging and production after merge? Or would I need to revert my changes afterwards to keep the images with emscripten 3.1.12 in staging and production channels?

context: I need these images to prepare PRs in other repositories to work with the newer emscripten (that usualy takes long time), while keep using the images with our current emscripten of choice (3.1.12) before these PRs are ready.

mthalman:
> do we have a way to exclude builds of these images from staging and production after merge? 

No. If the Dockerfile is referenced by the manifest.json file, it will be built.

> Or would I need to revert my changes afterwards to keep the images with emscripten 3.1.12 in staging and production channels?

Probably the best option is to make copies of these Dockerfiles and reference them in the manifest.json files, giving them new tags that can be referenced for those PRs. You'd only want to define a timestamp tag and not the floating tag though.

PR #781, opened 2023-01-26
Fix rootfs dir

sbomer:
This fixes the rootfs dir for the images added in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/754, to be consistent with the existing images.

I looked for a way to extract this into a single variable in the Dockerfile. The only way I found was to use ENV, but that would persist the environment variable into the created image - and I want to avoid polluting the image's environment.

### DIFF
--- a/src/ubuntu/20.04/cross/arm64/Dockerfile
+++ b/src/ubuntu/20.04/cross/arm64/Dockerfile
@@ -8,7 +8,7 @@ RUN apt-get update \
 
 
 FROM binutils AS rootfsbuild
-ARG ROOTFS_DIR=/crossrootfs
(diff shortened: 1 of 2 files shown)


mthalman:
@sbomer  - If you make an arbitrary change to the base Dockerfile, it will trigger it to get rebuilt. Then you can revert your change. It's based off commit SHA, so it should still rebuild even after you revert your change.

akoeplinger (approved):

mthalman (approved):
If this got you the build results you're looking for, go ahead and revert the crossdeps Dockerfile change.

PR #784, opened 2023-01-31
Disable image caching for production branch builds

### LINKED ISSUE #768: Define CI workflow specifically for production branch
When we run builds from the production branch, we should be setting the `noCache` variable to `true`. This ensures that all Dockerfiles are rebuilt no matter whether the Dockerfile content has changed or its base image has been updated. This ensures we're getting the latest package versions installed as well as dependencies to any changes in the [build-rootfs.sh script](https://github.com/dotnet/arcade/blob/main/eng/common/cross/build-rootfs.sh).

But this `noCache` variable needs to be manually set when queuing the build because the same pipeline used to run builds for the production branch is the same pipeline that is used for the main branch. Due to the frequency of builds that occur in the main branch, we don't want to have `noCache` set because it's quite expensive to rebuild all the images and is often unnecessary.

The same applies to the PR build. We'd want the PR build for the production branch to also set `noCache` in order to catch any issues prepping a production rollout before they get committed. Right now there's inconsistency between the PR build for production and the policy of setting `noCache` for the internal build.

The proposal here is to ensure that `noCache` is set to `true` for both the internal and PR builds of the production branch but left to be unset or set to `false` for the main branch.

mthalman:
Ensures that we don't use image caching for production builds.

Fixes #768

### DIFF
--- a/eng/pipelines/variables/common.yml
+++ b/eng/pipelines/variables/common.yml
@@ -29,6 +29,8 @@ variables:
 - ${{ if or(eq(variables['Build.SourceBranch'], 'refs/heads/production'), eq(variables['System.PullRequest.TargetBranch'], 'refs/heads/production')) }}:
   - name: publicSourceBranch
     value: production
+  - name: noCache
+    value: true
 - ${{ else }}:
   - name: publicSourceBranch
     value: main


michellemcdaniel (approved):
I wonder if we need to increase timeouts for this too

mthalman:
> I wonder if we need to increase timeouts for this too

Shouldn't need to. The existing timeout values already accounted for a full build with no caching. And the values are the same between the internal and PR builds.

mthalman:
I see that caching is still being used in this build for some reason so I need to look at what's going on.

mthalman:
> I see that caching is still being used in this build for some reason so I need to look at what's going on.

Oh, I know why. It's by design. This PR is targeting the main branch so caching is enabled.

michellemcdaniel:
Sounds good. Go forth and commit!

MichaelSimons (approved):

PR #787, opened 2023-02-05
Add FreeBSD 13 x64 image

Thefrank:
This adds a FreeBSD 13 x64 image alongside the existing FreeBSD 13 ARM64 and FreeBSD 12 x64 images.


Thefrank:
After building the required `coredeps` and `crossdeps` images locally on an Ubuntu 18.04 machine:

`sudo pwsh ./build.ps1 -DockerfilePath "*ubuntu/18.04/cross/freebsd/13*"`

```
IMAGES BUILT
------------
mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-cross-freebsd-13-20230205015147-20dc96e
mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-cross-freebsd-13local
```

wfurt (approved):
LGTM
I'm wondering if we should move to newer base image e.g. Ubuntu 20 or 22 but I think it is fine if we do it later.

Thefrank:
> LGTM I'm wondering if we should move to newer base image e.g. Ubuntu 20 or 22 but I think it is fine if we do it later.

It can be done at any point. I was going to wait a bit as 18.04 reaches end of "Hardware and maintenance updates", FreeBSD 12.3 reaches EOL (leaving only FreeBSD 12.4 in support), and FreeBSD 13.2 releases all happen in late March.

mthalman (approved):

PR #789, opened 2023-02-07
Remove base image override for Alpine

mthalman:
Unable to continue with base image overriding for Alpine since new Alpine tags are not being accepted. Falling back to what's specified in the Dockerfile instead. This unblocks the internal build for https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/788.


michellemcdaniel:
we'll need to do this in the all pipeline too: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/eng/pipelines/dotnet-buildtools-prereqs-all.yml#L29

michellemcdaniel (approved):

PR #792, opened 2023-02-09
fix locale setting for Alpine 3.17

wfurt:
related to https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/788#issuecomment-1423473134
I did not run tests but at least when running bash `LANG` is correct.


mthalman (approved):

MattGal (approved):
Seems like a fix

PR #798, opened 2023-02-10
fix locale setting for Alpine 3.16

wtgodbe:
Dupe of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/792

PR #799, opened 2023-02-10
improve Alpine 3.17 helix image

wfurt:
This is follow-up on #792 and #797 

While  #792  improved runtime test runs, some were still failing - particularly parsing foreign numbers. 
It seems like `musl-locales` package adds the missing definition (and locale -a work now just like glibc) 

#797  added base 3.17 image. So I refactor this to use it as base instead of duplicating the package. 
While `helix` may not need all the dev tools this is the structure we used in previous releases. 
Investigation if we can string helix images would be probably separate effort.

### DIFF
--- a/src/alpine/3.17/helix/amd64/Dockerfile
+++ b/src/alpine/3.17/helix/amd64/Dockerfile
@@ -1,46 +1,10 @@
-FROM alpine:3.17
-
-RUN apk update && \
-    apk add --no-cache \
-        autoconf \
-        automake \
-        bash \
-        build-base \
-        cargo \
-        clang \
-        clang-dev \
-        cmake \
-        coreutils \
-        curl \
-        gcc \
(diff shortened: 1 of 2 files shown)


wtgodbe:
Looks like we only have the helix dockerfile for 3.16, is just adding `musl-locales` sufficient there? I can make that change in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/798

mthalman (approved):

wfurt:
> Looks like we only have the helix dockerfile for 3.16, is just adding `musl-locales` sufficient there? I can make that change in #798

yes, I think just adding the package should be sufficient. I'm going to run full `runtime` tests and I can comment on https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/798 when I know the results.

Issue #805, opened 2023-02-16
we should consider removing build tools from Helix images


wfurt:
I somewhat stumbled on this one while running out of space on my local machine. I noticed that most of the helix images are pretty large. For example Ubuntu 18 -> 1.5G 
```
$ docker image ls
mcr.microsoft.com/dotnet-buildtools/prereqs         ubuntu-18.04-helix-amd64-20230216023557-4443d0d   6d7d1241e2c6   7 seconds ago   1.52GB
``` 

It seems like many of the images are pulling build tools directly or indirectly by relaying on the `prereq`  base image.
And in both cases they seems to be geared toward _building_ runtime not executing tests. This may be for historical reasons. 

I did little fiddling (https://github.com/dotnet/dotnet-buildtools-prereqs-docker/compare/ubu18?expand=1) and I can get the image to about 1/3
```
mcr.microsoft.com/dotnet-buildtools/prereqs         ubuntu-18.04-helix-amd64-20230216060058-4443d0d   07e33080f83d   5 seconds ago    593MB
``` 

even less with restricted locale
```
mcr.microsoft.com/dotnet-buildtools/prereqs         ubuntu-18.04-helix-amd64-20230216061525-4443d0d   c2d0fb3b5d9c   52 seconds ago   463MB
```

is the ~ 1G per image saving interesting and worth of the troubles @MattGal @mthalman ? 
I'm reasonably confident that the resulting image would be capable of running `runtime` tests. 
But I don't have really visibility to other repos and any historical reasons. 
If we decouple the dependency on the base local image we may see more duplication on Dockerfile - but probably not too much.

mthalman:
+ @ChadNedzlek 

1 GB of space is significant and would help overall with the overall storage costs associated with the images we produce.

MattGal:
No objection to installing less stuff, other than to point out the obvious that there _will_ be some teething pains in getting the new list correct and some tests will fail.

To my knowledge, most of the "why the test images look like the build images" is just an artifact of folks being new to using docker and there being only that list of dependencies available back when the images were created.  

[...]

PR #806, opened 2023-02-17
refactor Alpine Helix image

wfurt:
contributes to #805
old image was ~ 2.1G 
```
mcr.microsoft.com/dotnet-buildtools/prereqs         alpine-3.17-helix-amd64-20230210201636-609d24f    67b176d4e97d   6 days ago       2.1GB
```

new one is ~ 511M and it _should_ be functionally same AFAIK. 
```
mcr.microsoft.com/dotnet-buildtools/prereqs         alpine-3.17-helix-amd64-20230217065202-7f13c75    355f491a1337   17 seconds ago   511MB
```

there few important changes:
- removed the -dev packages. They should not be needed to run tests. 
- `cargo` & `rust` pulls in `gcc`. It was needed for Helix but I don't think we need them for runtime.  
- removed dev tools from base. Instead each layer that needs needs to add (and remove) them 
- for Helix I added rust, cargo, python3-dev and few more to support building Helix support
- msquic build was refactored to don't need PowerShell.  It does not impact image size but it avoid PWSH download. 

we can save two more layers if we move `helixbot` creation to Helix prep and locale.sh to runtime deps. 
But they are really small so I'm not sure if it is worth it. 

I don't know if there is good way how to run Helix runtime tests on this. If this looks good enough, I can run draft PR on the -staging image and follow-up if needed.  

BTW it is easy to pull in bunch of random dependencies unintentionally. Perhaps PRs for Helix should have at least total size for sanity check. 

Here are the boring details 
```
furt@ubu18:~$ docker image history 355f491a1337
IMAGE          CREATED             CREATED BY                                      SIZE      COMMENT
355f491a1337   12 minutes ago      /bin/sh -c python3 -m virtualenv /home/helix…   25.1MB
183b4c3a60f3   12 minutes ago      /bin/sh -c #(nop)  USER helixbot                0B
4c151dc5296b   12 minutes ago      /bin/sh -c /usr/sbin/adduser -D -g '' -G adm…   4.7kB
97c212c288a4   12 minutes ago      /bin/sh -c echo export LANG=${LANG}  >> /etc…   85B
0ff3449aacc8   12 minutes ago      /bin/sh -c #(nop)  ENV LANG=en-US.UTF-8         0B
a63cc52005ed   12 minutes ago      /bin/sh -c apk update && apk add --no-cache …   9.02MB
9e880fe20eee   About an hour ago   /bin/sh -c apk update && apk add --no-cache …   377MB
402fc0364e2b   About an hour ago   /bin/sh -c apk update && apk add --no-cache …   93.3MB
b2aa39c304c2   6 days ago          /bin/sh -c #(nop)  CMD ["/bin/sh"]              0B
<missing>      6 days ago          /bin/sh -c #(nop) ADD file:40887ab7c06977737…   7.05MB
```

### DIFF
--- a/src/alpine/3.17/helix/amd64/Dockerfile
+++ b/src/alpine/3.17/helix/amd64/Dockerfile
@@ -1,33 +1,52 @@
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.17-local
+FROM alpine:3.17
+# Install .NET and test dependencies
 RUN apk update && apk add --no-cache \
-    cargo \
+    bash \
+    coreutils \
+    curl \
+    icu-data-full \
     iputils \
-    libffi-dev \
+    lttng-ust \
     musl-locales \
-    rust
+    openssl \
+    python3 \
+    sudo \
+    tzdata
 
 # Install Helix Dependencies
-RUN ln -sf /usr/bin/python3 /usr/bin/python && \
+RUN apk update && apk add --no-cache && \
+    apk add cargo linux-headers python3-dev openssl-dev && \
+    apk list --installed > /start2.txt && \
+    ln -sf /usr/bin/python3 /usr/bin/python && \
     curl https://bootstrap.pypa.io/get-pip.py -o ./get-pip.py --fail --silent --show-error && \
-    python ./get-pip.py && rm ./get-pip.py && \
-    python -m pip install --upgrade pip==22.2.2 && \
-    python -m pip install virtualenv==20.16.5 && \
+    python3 ./get-pip.py && rm ./get-pip.py && \
+    python3 -m pip install --upgrade pip==22.2.2 && \
+    python3 -m pip install virtualenv==20.16.5 && \
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
-    pip install ./helix_scripts-*-py3-none-any.whl
+    pip install ./helix_scripts-*-py3-none-any.whl && \
+    apk del cargo linux-headers python3-dev openssl-dev
 
-# build MsQuic as we don't have packages
-RUN cd /tmp && \
-    mkdir pwsh && \
-    cd pwsh && \
-    curl -L https://github.com/PowerShell/PowerShell/releases/download/v7.2.5/powershell-7.2.5-linux-alpine-x64.tar.gz | tar xfz - && \
-    cd .. && \
-    git clone --depth 1 --single-branch --branch main --recursive https://github.com/dotnet/msquic && \
-    cd msquic/src/msquic  &&  PATH=~/.dotnet/tools:$PATH /tmp/pwsh/pwsh scripts/build.ps1 -Config Release -UseSystemOpenSSLCrypto -DisableTools -DisableTest -DisablePerf && \
-    cp artifacts/bin/linux/x64_Release_openssl3/libmsquic.so.2 artifacts/bin/linux/x64_Release_openssl3/libmsquic.lttng.so.2.* /usr/lib && \
+# build MsQuic as we don't have Alpine packages (yet)
+RUN apk update && apk add --no-cache && \
+    apk add cmake g++ gcc git numactl-dev linux-headers lttng-ust-dev make musl-dev openssl-dev perl && \
     cd /tmp && \
-    rm -r pwsh msquic
+    git clone --depth 1 --single-branch --branch main --recursive https://github.com/dotnet/msquic && \
+    cd msquic/src/msquic && \
+    cmake -B build/linux/x64_openssl3 \
+       -DQUIC_OUTPUT_DIR=/tmp/msquic/src/msquic/artifacts/bin/linux/x64_Release_openssl3 \
+       -DCMAKE_BUILD_TYPE=Release \
+       -DQUIC_TLS=openssl3 \
+       -DQUIC_ENABLE_LOGGING=true \
+       -DQUIC_USE_SYSTEM_LIBCRYPTO=true \
+       -DQUIC_BUILD_TOOLS=off \
+       -DQUIC_BUILD_TEST=off \
+       -DQUIC_BUILD_PERF=off && \
+    cmake --build build/linux/x64_openssl3  --config Release && \
+    cp artifacts/bin/linux/x64_Release_openssl3/libmsquic.so.* artifacts/bin/linux/x64_Release_openssl3/libmsquic.lttng.so.* /usr/lib && \
+    rm -rf /tmp/msquic && \
+    apk del cmake g++ gcc git linux-headers lttng-ust-dev make musl-dev openssl-dev perl
 
-# Needed for corefx tests to pass
+# Needed for runtime tests to pass
 ENV LANG=en-US.UTF-8
 RUN echo export LANG=${LANG}  >> /etc/profile.d/locale.sh
 
@@ -39,4 +58,4 @@ RUN /usr/sbin/adduser -D -g '' -G adm -s /bin/bash -u 1000 helixbot && \
 
 USER helixbot
 
-RUN python -m virtualenv /home/helixbot/.vsts-env
+RUN python3 -m virtualenv /home/helixbot/.vsts-env
--- a/src/alpine/3.17/helix/amd64/Dockerfile
+++ b/src/alpine/3.17/helix/amd64/Dockerfile
@@ -16,7 +16,6 @@ RUN apk update && apk add --no-cache \
 # Install Helix Dependencies
 RUN apk update && apk add --no-cache && \
     apk add cargo linux-headers python3-dev openssl-dev && \
-    apk list --installed > /start2.txt && \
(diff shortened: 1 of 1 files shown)


src/alpine/3.17/helix/amd64/Dockerfile
```diff
@@ -1,33 +1,52 @@
...
+    lttng-ust \
     musl-locales \
-    rust
+    openssl \
+    python3 \
+    sudo \
+    tzdata
 
 # Install Helix Dependencies
-RUN ln -sf /usr/bin/python3 /usr/bin/python && \
```
MattGal, line 10: Helix agents expect to just find "python" as the entry point and value for `$HELIX_PYTHONPATH` given the bottom of this file it seems like that doesn't work.  Rest of the helix-y changes seem fine.

MattGal:
> I don't know if there is good way how to run Helix runtime tests on this. If this looks good enough, I can run draft PR on the -staging image and follow-up if needed.

Without merging to staging, you can't easily use the images this repo's pipelines will make.   If you can, however, publish your images to a test public Azure Container registry (I can't provide this but if you have access to a subscription it's quite easy), it should work just like mcr (although it _might_ trigger some Azure Security pack scanning stuff :) )

src/alpine/3.17/helix/amd64/Dockerfile
wfurt: I was no intention, I'll put it back. I updated the rest so we do not rely on the link in case we fix/change the HELIX_PYTHONPATH one day.

wfurt: hmm. actually, it is just moved to line 20. so everything should work as it did before. I put the `apk add` first to preserve the sequence but I guess in this case it does not matter.

wfurt:
BTW since the `apk add` and `apk del` needs same list of packages, I was thinking about using ENV to hold them. But I did not like the fact that it creates new layer. https://docs.docker.com/develop/develop-images/dockerfile_best-practices/ contradicts that. 

Would anybody here have opinion on this?

ChadNedzlek:
Are there any functional changes here? Almost all of this is going to get lost in the new method of making images because we are using artifacts that are shared everywhere to install things (like python, helix, pip, and cmake), so most of these won't work.

src/alpine/3.17/helix/amd64/Dockerfile
```diff
@@ -1,33 +1,51 @@
...
-    python ./get-pip.py && rm ./get-pip.py && \
-    python -m pip install --upgrade pip==22.2.2 && \
-    python -m pip install virtualenv==20.16.5 && \
+    python3 ./get-pip.py && rm ./get-pip.py && \
+    python3 -m pip install --upgrade pip==22.2.2 && \
+    python3 -m pip install virtualenv==20.16.5 && \
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
-    pip install ./helix_scripts-*-py3-none-any.whl
+    pip install ./helix_scripts-*-py3-none-any.whl && \
+    apk del cargo linux-headers python3-dev openssl-dev
```
ChadNedzlek, line 26: Helix does a bunch of wheel management in it's start up scripts. I'm worried removing these packages might break something in some weird corner of some servicing branch.

wfurt: Right. This is one reason why I chose Alpine 3.17 - it is brand new and even runtime did not use it until very recently.

wfurt:
> Are there any functional changes here? Almost all of this is going to get lost in the new method of making images because we are using artifacts that are shared everywhere to install things (like python, helix, pip, and cmake), so most of these won't work.

I found (and fix) missing `lldb`. We use it to dump info when core file is created. Adds ~ 300M -> 1/3 of the image.
The rest should be functionally identical. I will not know for sure until I can do Helix test run. 

As far as the upcoming changes - I certainly cannot speak to that. The lesson I learn is that each layer should only bring what is needed. If `pip` just brings pip and dependencies it should not matter if that is global. 

The other past is not to bring packages that are not needed. Like the `cmake`. It is tool to build, not to run tests. e.g. it should be in build images but not test images. and same for the gcc/clang and bunch of `-dev` packages we currently suck in.

src/alpine/3.17/helix/amd64/Dockerfile
michellemcdaniel: I think the concern here is about helix itself, not the branches that are running. If *helix* needs these packages installed to do something in some niche way, we'll need them to stick around

wfurt: Is there some way how to determine that @michellemcdaniel? As I mentioned I'm planning to run tests but the questions if that would show up. AFAIK this is currently only used by runtime in `extra-platforms` pipeline that runs rarely and probably can take some risk. 

This does not block us in any way, so I can just leave it as is. I was trying to check if we can come up with smaller images - but it should OK if we cannot it. Alpine in particular is focus on small footprint so seeing 2G made me wonder.  

We _could_ also leave `rust` or whatever Helix needs there for now. We would still end up with much smaller size.

ChadNedzlek:
If it's only used in a narrow case that it sounds like we don't care much about... why make this change, especially given that it's only temporary?

wfurt:
> If it's only used in a narrow case that it sounds like we don't care much about... why make this change, especially given that it's only temporary?

I have no visibility to what is coming. I put it up as improvement - and possibly something we can compare to. 
If the consensus is that we don't care for now, I can close it.

src/alpine/3.17/helix/amd64/Dockerfile
michellemcdaniel: The only way I can think of for confirming that is to test it. Like publish a version of it and then run it through a pipeline. What if instead of modifying this dockerfile, we created a scouting file/version with these changes so that we won't break anyone who might be using this and then we can test this version out in various pipelines so we can see what can and can't be removed? And then once we've done all the scouting, we can make changes to the main version of the file and remove the scouting version. I don't know how @mthalman would feel about that though.

wfurt: we could. AFAIK this is only consumed by https://github.com/dotnet/runtime/pull/81841 - the `extra-platform` only runs on demand and AFAIK it had 0% success rate in last few month. 
I was also thinking about leaving the `cargo` in but leave it in the Helix layer to make it clear it is Helix dependency.
At this point I'm trying to stabilize the pattern and duplicated for other architectures.

src/alpine/3.17/helix/amd64/Dockerfile
```diff
@@ -1,33 +1,52 @@
...
-    rust
+    openssl \
+    python3 \
+    sudo \
+    tzdata
 
 # Install Helix Dependencies
-RUN ln -sf /usr/bin/python3 /usr/bin/python && \
+RUN apk update && apk add --no-cache && \
+    apk add cargo linux-headers python3-dev openssl-dev && \
```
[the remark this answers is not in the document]
wfurt: updated.

src/alpine/3.17/helix/amd64/Dockerfile
wfurt: BTW we could simply not promote it to production, right? e.g. leave it in staging until proven good.

michellemcdaniel: Not really. That would block every change that is made to this repo from rolling out. The best we can do is check in on a Tuesday after the rollout PR is made (assuming nothing else is necessary, so really, Wednesday is best), and then that would give you a week to test it and iterate (hopefully quickly) before the next rollout would kind of require it to be reverted, assuming that there is something else that actually needs to be rolled out.

mthalman: If this is just a one-time thing, and not going to be a regular occurrence, I can build this Dockerfile and push it to our dev ACR. You can then pull it from there. If this is going to happen more often, my suggestion would be for the runtime team to have its own ACR for these testing purposes.

ChadNedzlek: > duplicated for other architectures

Are you planning on making more sweeping changes? Remember, most of this repository is going to be discarded in a month or two. _Please_ coordinate with @michellemcdaniel and I, otherwise all of this work is going to get discarded along with it.  Now is not a good time for large changes to this repository.  Since we are in the middle of attempting to migrate it, it's very likely any changes will be lost.

wfurt: I don't see this as sweeping change @ChadNedzlek. This is minor cleanup to avoid pulling in unneeded packages. And no, I don't plan more changes like this. We will need to add MsQuic build to more images but that does not really change the infrastructure.

mthalman (approved):

mthalman:
Given that the work to internally produce these images is basically suspended, there's no reason to not move forward with these changes.

wfurt:
I will make test runs in runtime and I'll monitor it for a while. Just as I did more testing for other issues I found out our Centos 7 image is 4.2GB. That seems excessive and perhaps deserves some look regardless how we construct the image.

PR #808, opened 2023-02-21
update msquic for Ubuntu 22

wfurt:
This brings current msquic to both x64 & arm64. 
We use OpenSSL 3 using `UseSystemOpenSSLCrypto`

Aside from that there are two more changes:
- drop PowerShell and build directly using cmake
- build and install package instead of file copy. That will hopefully make it closer to final production.

### DIFF
--- a/src/ubuntu/22.04/helix/amd64/Dockerfile
+++ b/src/ubuntu/22.04/helix/amd64/Dockerfile
@@ -43,6 +43,33 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
     pip install ./helix_scripts-*-py3-none-any.whl
 
+# build MsQuic as we don't have binary package.
+# Ubuntu 22  does not have OpenSSL 1.1.1 package and Quic releases currently do not work with OpenSsl3
+# support for OpenSSL3 is early progress in msquic/main
+RUN apt-get update && \
+    apt-get install -qq -y libnuma-dev liblttng-ust-dev ruby-dev && \
+    gem install fpm && \
+    cd /tmp && \
+    git clone --depth 1 --single-branch --branch main --recursive https://github.com/dotnet/msquic && \
+    cd msquic/src/msquic && \
+    cmake -B build/linux/x64_openssl3 \
+        -DQUIC_OUTPUT_DIR=/tmp/msquic/src/msquic/artifacts/bin/linux/x64_Release_openssl3 \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DQUIC_TLS=openssl3 \
+        -DQUIC_ENABLE_LOGGING=true \
+        -DQUIC_USE_SYSTEM_LIBCRYPTO=true \
+        -DQUIC_BUILD_TOOLS=off \
+        -DQUIC_BUILD_TEST=off \
+        -DQUIC_BUILD_PERF=off && \
+     cmake --build build/linux/x64_openssl3  --config Release && \
+     ./scripts/make-packages.sh --tls openssl3 --config Release && \
+     dpkg -i artifacts/packages/linux/x64_Release_openssl3/libmsquic_*.deb && \
+     rm -rf /tmp/msquic && \
+     apt-get remove -y libnuma-dev liblttng-ust-dev ruby-dev && \
+     apt autoremove -y && \
+     rm -rf  /tmp/msquic && \
+     rm -rf /var/lib/apt/lists/*
+
 # create helixbot user and give rights to sudo without password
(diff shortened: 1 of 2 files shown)


wfurt:
BTW

```
helixbot@38871c2f2e04:/$ dpkg -L libmsquic
/.
/usr
/usr/share
/usr/share/doc
/usr/share/doc/libmsquic
/usr/share/doc/libmsquic/changelog.gz
/usr/lib
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/libmsquic.lttng.so.2.2.0
/usr/lib/x86_64-linux-gnu/libmsquic.so.2.2.0
/usr/lib/x86_64-linux-gnu/libmsquic.so.2

helixbot@38871c2f2e04:/$ ldd /usr/lib/x86_64-linux-gnu/libmsquic.so.2.2.0
	linux-vdso.so.1 (0x00007ffe27bdb000)
	libcrypto.so.3 => /lib/x86_64-linux-gnu/libcrypto.so.3 (0x00007fdbf6a8b000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fdbf6863000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fdbf7035000)
```

src/ubuntu/22.04/helix/amd64/Dockerfile
```diff
@@ -43,6 +43,33 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
     pip install ./helix_scripts-*-py3-none-any.whl
 
+# build MsQuic as we don't have binary package.
+# Ubuntu 22  does not have OpenSSL 1.1.1 package and Quic releases currently do not work with OpenSsl3
+# support for OpenSSL3 is early progress in msquic/main
```
ManickaP, line 48:
```suggestion
# Build MsQuic as we don't have binary package.
# Ubuntu 22  does not have OpenSSL 1.1.1 package and MsQuic releases currently do not work with OpenSSL 3.
# Support for OpenSSL 3 is early progress in msquic/main.
```

src/ubuntu/22.04/helix/amd64/Dockerfile
```diff
@@ -43,6 +43,33 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
...
+RUN apt-get update && \
+    apt-get install -qq -y libnuma-dev liblttng-ust-dev ruby-dev && \
+    gem install fpm && \
+    cd /tmp && \
+    git clone --depth 1 --single-branch --branch main --recursive https://github.com/dotnet/msquic && \
+    cd msquic/src/msquic && \
+    cmake -B build/linux/x64_openssl3 \
+        -DQUIC_OUTPUT_DIR=/tmp/msquic/src/msquic/artifacts/bin/linux/x64_Release_openssl3 \
```
ManickaP, line 56: Does this need to be a full path? Do we even need this? I.e. how (un)predictable the output path is?

src/ubuntu/22.04/helix/amd64/Dockerfile
```diff
@@ -43,6 +43,33 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
...
+        -DQUIC_BUILD_TOOLS=off \
+        -DQUIC_BUILD_TEST=off \
+        -DQUIC_BUILD_PERF=off && \
+     cmake --build build/linux/x64_openssl3  --config Release && \
+     ./scripts/make-packages.sh --tls openssl3 --config Release && \
+     dpkg -i artifacts/packages/linux/x64_Release_openssl3/libmsquic_*.deb && \
+     rm -rf /tmp/msquic && \
+     apt-get remove -y libnuma-dev liblttng-ust-dev ruby-dev && \
+     apt autoremove -y && \
+     rm -rf  /tmp/msquic && \
```
ManickaP, line 70: You have `rm -rf /tmp/msquic` twice, here and on line 67.

ManickaP (approved):
Some small comments, otherwise LGTM. Thanks!

src/ubuntu/22.04/helix/amd64/Dockerfile
wfurt: it seems to. I started with the relative path from `-B` but then the output was created under the ` build/linux/x64_openssl3` e.g. `build/linux/x64_openssl3/artifacts/bin/linux/x64_Release_openssl3`.

src/ubuntu/22.04/helix/amd64/Dockerfile
wfurt: removed

mthalman (approved):

wfurt:
it seems like this was built as `-staging` but never made it to "production" @mthalman 
Do we just need to rebuild the queue to get new tags? 
Attempt to pull it to runtime looks good https://github.com/dotnet/runtime/pull/82460

wfurt:
mariner image seems to have the same issue.

PR #813, opened 2023-02-25
Add linux-musl-riscv64 image

am11:


am11:
cc @janvorli, from `ubuntu-22.04-coredeps-graph` logs https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/189426/logs/485:

> 2023-03-01T14:37:27.8080607Z fetch http://dl-cdn.alpinelinux.org/alpine/edge/community/riscv64/APKINDEX.tar.gz
> 2023-03-01T14:37:28.0872970Z fetch http://dl-cdn.alpinelinux.org/alpine/edge/main/riscv64/APKINDEX.tar.gz
> ...
> 2023-03-01T14:37:52.9116496Z (107/121) Installing llvm15-libs (15.0.7-r0)

which means with `edge`, it used the new lookup mechanism to find `llvm15-libs` package and installed it. Alpine riscv64 is only available in edge. Once Alpine riscv64 becomes part of the release, we will distill the versions in build-rootfs (like we have for other architectures).

akoeplinger (approved):

PR #815, opened 2023-02-27
Use same rootfs in Ubuntu 22 cross arm64 as 18

am11:
We need to keep this to lowest for portability.


mthalman (approved):

am11:
Second commit: Just remembered that, while coreclr is happy with just the "modern llvm toolchain", there is one place in mono which is forcing us to install binutils https://github.com/dotnet/runtime/blob/b4b3105e944409ad0a20eb2c1948cf98ef2833f1/src/mono/mono.proj#L644. cc @lambdageek, @akoeplinger another reason why making mono use `eng/native/configuretools.cmake` for objcopy will be an improvement. :)

akoeplinger (approved):

akoeplinger:
@am11 if that is the only place then we could probably detect whether we have `llvm-objcopy` instead there.

PR #817, opened 2023-02-28
add MsQuic to Alpine 3.17 arm images

wfurt:
contributes to https://github.com/dotnet/runtime/issues/81801


src/alpine/manifest.json
```diff
@@ -244,7 +244,8 @@
                 "alpine-3.17-local": {
                   "isLocal": true
                 }
-              }
+              },
+              "variant": "v7"
```
mthalman, line 248: Looks like this got added in the wrong spot. Should be with the arm32 Dockerfile you added.

mthalman (approved):

PR #820, opened 2023-03-03
Add krb5-libs to AMD64 Alpine 3.15 image

vcsjones:
The AMD64 Alpine 3.15 image was missing `krb5-libs`. I confirmed that the ARM32 and ARM64  images had the package.

Fixes https://github.com/dotnet/runtime/issues/82945

/cc @wfurt


wfurt (approved):
thanks

akoeplinger (approved):

PR #829, opened 2023-03-23
Add signing keys for raspbian and debian sid rootfs build

sbomer:
This should keep existing rootfs builds working with the change in https://github.com/dotnet/arcade/pull/12900 that forces package signature checks in debootstrap by default.


src/raspbian/10/helix/arm32v6/Dockerfile
```diff
@@ -7,6 +7,7 @@ FROM scratch
 # But we can use a crossrootfs from our image generation scripts, and
 # extract into the image root
 
+RUN echo no-op
```
am11, line 10: `FROM scratch` containers (line 4) is literally empty / zero-byte. The failures that `/bin/sh` not found is because docker's `RUN` command is trying to invoke `/bin/sh echo no-op` here and failing to find the executable.

am11, line 10: `FROM scratch` containers (line 4) areliterally empty / zero-byte. The failures that `/bin/sh` not found is because docker's `RUN` command is trying to invoke `/bin/sh echo no-op` here and failing to find the executable.

am11, line 10: `FROM scratch` containers (line 4) are literally empty / zero-bytes. The failure that `/bin/sh` was not found is because docker's `RUN` command is trying to invoke `/bin/sh echo no-op` here and failing to find the executable.

You can move it before `FROM scratch` (line 3).

mthalman (approved):

PR #832, opened 2023-03-27
Add mariner build images

### LINKED ISSUE #833: Update image name for Mariner 1.0
The [CBL-Mariner 1.0 Helix Dockerfile](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/4404b5c3d9fdb70a765a8cc8d7df6423f15e6033/src/cbl-mariner/1.0/helix/amd64/Dockerfile#L1) is referencing an obsolete image name that no longer works.

It leads to this build failure:

```
PULLING LATEST BASE IMAGES
--------------------------
-- EXECUTING: docker pull --platform linux/amd64 cblmariner.azurecr.io/base/core:1.0
Error response from daemon: Head "[https://cblmariner.azurecr.io/v2/base/core/manifests/1.0"](https://cblmariner.azurecr.io/v2/base/core/manifests/1.0%22): unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information.
```

The image name should be changed to `mcr.microsoft.com/cbl-mariner/base/core:1.0` instead.

This is blocking the build of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/832.

sbomer:
This adds build images for CBL-mariner which can be used to build dotnet/runtime (using the changes in https://github.com/dotnet/runtime/compare/main...sbomer:runtime:marinerBuild). All of these images (including x64) contain a rootfs, to ensure consistent glibc and musl targeting. The rootfs is built using the changes in https://github.com/dotnet/arcade/pull/12900, which expect debootstrap to be available on the host (for the ubuntu rootfs). We obtain debootstrap from the mariner extended repos for now.

The rootfs distros are:
- Ubuntu 16.04 (Xenial), with glibc 2.23
- Alpine 3.13, with musl libc 1.2.2

The images are all using LLVM 12, matching the version available in the mariner repositories. They don't contain binutils on the host, and expect the build to use LLVM-based tools. We plan to upgrade to LLVM 16 in a future change.

For x64 and x86, we use LLVM from the mariner repos. For arm/arm64, we build LLVM from source. We also build the compiler-rt.profile library (used for PGO instrumentation) for the target architecture where necessary.

Arm images will be added in a future change (I still need to investigate the libnuma dependency which isn't available on xenial or alpine 3.13).

Contributes to https://github.com/dotnet/runtime/issues/83428.

Fixes #833

### DIFF
--- a/src/cbl-mariner/manifest.json
+++ b/src/cbl-mariner/manifest.json
@@ -49,6 +49,8 @@
               "os": "linux",
               "osVersion": "cbl-mariner2.0",
               "tags": {
+                "cbl-mariner-2.0-crossdeps-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "cbl-mariner-2.0-crossdeps$(FloatingTagSuffix)": {},
                 "cbl-mariner-2.0-crossdeps-local": {
                   "isLocal": true
                 }
@@ -63,6 +65,8 @@
               "os": "linux",
               "osVersion": "cbl-mariner2.0",
               "tags": {
+                "cbl-mariner-2.0-crossdeps-builder-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "cbl-mariner-2.0-crossdeps-builder$(FloatingTagSuffix)": {},
                 "cbl-mariner-2.0-crossdeps-builder-local": {
                   "isLocal": true
                 }
@@ -77,6 +81,8 @@
               "os": "linux",
               "osVersion": "cbl-mariner2.0",
               "tags": {
+                "cbl-mariner-2.0-crossdeps-amd64-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "cbl-mariner-2.0-crossdeps-amd64$(FloatingTagSuffix)": {},
                 "cbl-mariner-2.0-crossdeps-amd64-local": {
                   "isLocal": true
                 }


src/cbl-mariner/2.0/cross/arm64/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-arm-local
+ARG ROOTFS_DIR=/crossrootfs/arm64
+
+RUN tdnf install -y debootstrap
+
+RUN /scripts/eng/common/cross/build-rootfs.sh arm64 xenial --skipunmount
+
+# Build compiler-rt profile library for PGO instrumentation
+COPY --from=mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-artifacts-local \
```
mthalman, line 9: You can't reference a local tag in the `--from` option. This only works if the tag exists in a registry. Instead, you'll need to define a stage from this image name. Something like this:

```Dockerfile
FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-artifacts-local AS artifacts


FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-arm-local
ARG ROOTFS_DIR=/crossrootfs/arm64

RUN tdnf install -y debootstrap

RUN /scripts/eng/common/cross/build-rootfs.sh arm64 xenial --skipunmount

# Build compiler-rt profile library for PGO instrumentation
COPY --from=artifacts /llvm-project.src.tar.xz .
```

I'm not sure our infrastructure can support a Dockerfile like this. Give it a shot though and see.

src/cbl-mariner/manifest.json
```diff
@@ -42,6 +42,127 @@
...
+              "os": "linux",
+              "osVersion": "cbl-mariner2.0",
+              "tags": {
+                "cbl-mariner-2.0-crossdeps-artifacts-local": {
+                  "isLocal": true
+                }
+              }
+            }
+          ]
+        },
```
mthalman, line 100: The Dockerfiles in the manifest should be listed in dependency order. So this one should be listed first in your changes.

mthalman:
Looks like the llvm stuff is causing the build to take much longer. Please update the timeout of the pipeline. You can see an example of that here: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/4404b5c3d9fdb70a765a8cc8d7df6423f15e6033/eng/pipelines/dotnet-buildtools-prereqs-ubuntu.yml#L27

I would suggest a value of 150. This needs to be applied to two pipelines:
* https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/eng/pipelines/dotnet-buildtools-prereqs-mariner-pr.yml
* https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/eng/pipelines/dotnet-buildtools-prereqs-mariner.yml

mthalman (commented):
The build failure is because the infrastructure doesn't support local-only Dockerfiles. There must be at least one tag associated with a Dockerfile that does not have `isLocal` set. So you'd need to add a unique tag for each of these crossdeps Dockerfiles.

src/cbl-mariner/2.0/cross/arm64/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+        -DCMAKE_C_FLAGS="${BUILD_FLAGS}" \
+        -DCMAKE_CXX_FLAGS="${BUILD_FLAGS}" && \
+    make -j $(getconf _NPROCESSORS_ONLN) && \
+    make install
+
+RUN LLVM_VERSION=12.0.1 LLVM_VERSION_MAJOR="${LLVM_VERSION%%.*}" && \
+    mkdir -p $ROOTFS_DIR/usr/lib/llvm-${LLVM_VERSION_MAJOR}/lib/clang/${LLVM_VERSION}/lib/linux/ && \
+    cp compiler-rt_build/lib/linux/libclang_rt.profile-aarch64.a $ROOTFS_DIR/usr/lib/llvm-${LLVM_VERSION_MAJOR}/lib/clang/${LLVM_VERSION}/lib/linux/
+
+RUN rm -r compiler-rt_build llvm-project.src
```
janvorli, line 43: If I remember how docker works correctly, although this "rm" removes the directories, they will stay hidden in the image, making it unnecessarily larger. In other words, every "RUN" command adds a layer to the docker image. Consider everything that exists at the end of the "RUN" as "committed" into the image forever. So the "RUN" with rm basically just creates a "commit" that removes the visibility of the files. Similar to what removing a file in a git commit does. So if you really want to remove some build artifacts, it should be done in the same "RUN" command that created the files.
The multistage builds were invented to remedy this problem in a cleaner way. Each stage can copy selected files from the previous stage and things that were not copied are thrown away. See https://docs.docker.com/build/building/multi-stage/ for more details.

wfurt: yes, they do. This is what many of the commands are often linked together with `&& \`. 
(and when packages are added we clean each layer separately)

src/cbl-mariner/2.0/crossdeps-builder/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# Build LLVM cross-toolchain (with support for targeting arm architectures)
+RUN mkdir llvm-project.src && \
+    tar -xf llvm-project.src.tar.xz --directory llvm-project.src --strip-components=1 && \
+    rm llvm-project.src.tar.xz && \
+    mkdir build && cd build && \
+    cmake ../llvm-project.src/llvm \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DLLVM_TARGETS_TO_BUILD="host;AArch64;ARM" \
+        -Wno-dev \
+        -DLLVM_ENABLE_PROJECTS="clang;lld" && \
```
am11, line 58:
```suggestion
        -DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra;lld;lldb" && \
```
This is to make sure we have `lld-12`, `lldb-12`, `llvm-objcopy-12`, `llvm-dwarfdump-12` and `llvm-symbolizer-12` available in PATH. Please double check that these prerequisites, used by various runtime CI legs, are available in the final image's host layer (crossrootfs only needs llvm-libs which is handled by build-rootfs.sh).

src/cbl-mariner/2.0/crossdeps-builder/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+    LLVM_VERSION=12.0.1 && \
+    wget -O llvm-project.src.tar.xz.sig https://github.com/llvm/llvm-project/releases/download/llvmorg-${LLVM_VERSION}/llvm-project-${LLVM_VERSION}.src.tar.xz.sig && \
+    echo "6cc956d622a7d3d746de0d71d8ca616a6c291e2c561703ac7a9535f38b999955 llvm-project.src.tar.xz.sig" | sha256sum -c && \
+    wget -O llvm-project.src.tar.xz https://github.com/llvm/llvm-project/releases/download/llvmorg-${LLVM_VERSION}/llvm-project-${LLVM_VERSION}.src.tar.xz && \
+    echo "129cb25cd13677aad951ce5c2deb0fe4afc1e9d98950f53b51bdcfb5a73afa0e llvm-project.src.tar.xz" | sha256sum -c && \
+    gpg --keyring /release-keys.gpg --verify llvm-project.src.tar.xz.sig && \
+    rm llvm-project.src.tar.xz.sig
+
+# Build LLVM cross-toolchain (with support for targeting arm architectures)
+RUN mkdir llvm-project.src && \
```
am11, line 50: nit: I think if we `make` everything under `/tmp` then the intermediate artifacts will be cleaned up (and we will only have `make install`'d binaries/files under `/usr` in the final image; which is what we need).

src/cbl-mariner/2.0/crossdeps-amd64/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-local
+
+RUN tdnf install -y \
+    clang \
+    lld \
+    compiler-rt
```
mthalman, line 6: Nit: please indent these

src/cbl-mariner/2.0/crossdeps-builder/Dockerfile
am11: Thanks for checking.

`lldb` is required in diagnostics repo's CI (since forever) and in runtime repo (since  https://github.com/dotnet/runtime/pull/82867) to troubleshoot stuff in the very image reporting a crash.

`clang-tools-extra` is not _required_ but would be nice to put in the image while we are at it, so we don't have to rely on random (unmaintained) blob for `clang-format` and `clang-tidy`: https://github.com/dotnet/jitutils/blob/main/bootstrap.sh#L90 which CI in runtime end-up using in jit-diff legs.

`llvm-ranlib` seems to be missing in the list. From alpine-arm64 prereq image:

```
$ docker run -it mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-22.04-cross-arm64 bash -c 'find /usr -name llvm* | grep ranlib'

/usr/lib/llvm-15/bin/llvm-ranlib
/usr/bin/llvm-ranlib-15
```

src/cbl-mariner/2.0/crossdeps-builder/Dockerfile
am11: Cool. If this layer is always going to be consumed in multistage, then no need to switch to /tmp. BTW, all builttools images have empty /tmp, so anything we are putting there won't be persisted.

src/cbl-mariner/2.0/cross/amd64/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN /scripts/eng/common/cross/build-rootfs.sh x64 xenial --skipunmount
+
+
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-amd64-local
+ARG ROOTFS_DIR=/crossrootfs/x64
+
+COPY --from=builder $ROOTFS_DIR $ROOTFS_DIR
+
+RUN LLVM_VERSION=12.0.1 LLVM_VERSION_MAJOR="${LLVM_VERSION%%.*}" && \
```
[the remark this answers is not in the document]
am11: Curious, why v12.0.1 specifically even as a temporary solution?  If it was v9.0.x (our go-to llvm version in official legs) or v15.0.x (modern version used in latest linux-musl legs) that would make more sense for a temporary solution. AFAIK, we aren't using 12.0.x explicitly for anything.

src/cbl-mariner/2.0/crossdeps-builder/Dockerfile
mthalman: There's nothing special about the /tmp directory. Anything placed there will be committed to the image layer after executing a `RUN` instruction.

am11: You are right, we do clear it manually. e.g. https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/e1bc4b7f52f903020b147ee26de6e3e918f67915/src/ubuntu/22.04/cross/x86-alpine/Dockerfile#L3-L20

I was reading a post in docker repo about `/var/tmp` which wasn't get cleaned up as it was supposed to (that bug was fixed in some version of docker).

mthalman:
> > I believe the failure earlier was due to missing `qemu-user-static` on the host machine. It seems like the local-only images were working. Is it ok to change back to local-only images for the intermediate images that don't need to be consumed in dotnet/runtime?
> 
> Ok, I thought I was seeing something else. Go ahead and try switching back. Hopefully that works then.

The current build failure is this:

```
Unhandled exception: System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.ThrowHelper.ThrowNoMatchException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.PublishImageInfoAsync() in /image-builder/src/Commands/BuildCommand.cs:line 155
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.<ExecuteAsync>b__14_0() in /image-builder/src/Commands/BuildCommand.cs:line 77
   at Microsoft.DotNet.ImageBuilder.DockerHelper.ExecuteWithUserAsync(Func`1 action, String username, String password, String server, Boolean isDryRun) in /image-builder/src/DockerHelper.cs:line 49
   at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.ExecuteAsync() in /image-builder/src/Commands/BuildCommand.cs:line 66
   at System.CommandLine.Invocation.CommandHandler.GetResultCodeAsync(Object value, InvocationContext context)
   at System.CommandLine.Invocation.ModelBindingCommandHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
```

This is what I saw earlier that is caused by having local-only tags. So you'll need to add those non-local tags back.

mthalman (approved):

janvorli (approved):
LGTM, thank you!

PR #836, opened 2023-03-31
Add arm mariner build images

sbomer:
This adds the arm images which were left out of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/832. ~They include logic to download the numa headers, working around packages that didn't exist for arm in Ubuntu 16.04 or Alpine 3.13.~

See https://github.com/dotnet/arcade/pull/12144 for context on why we need the numa headers installed.

Contributes to https://github.com/dotnet/runtime/issues/83428

@janvorli @am11 @mthalman PTAL

### DIFF
--- a/src/cbl-mariner/2.0/cross/arm-alpine/Dockerfile
+++ b/src/cbl-mariner/2.0/cross/arm-alpine/Dockerfile
@@ -1,0 +1,22 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-builder-local AS builder
+ARG ROOTFS_DIR=/crossrootfs/arm
+
+RUN /scripts/eng/common/cross/build-rootfs.sh arm alpine3.13 --skipunmount
+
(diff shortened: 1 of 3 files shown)


mthalman (approved):

am11:
@sbomer, thanks. I was thinking about NativeAOT [static executable](https://gist.github.com/am11/7cca4a54745eb4ce37d5789301b9f735) use-case, where we have option to link libnuma statically in the binary. The license of libnuma is LGPL 2.1, which restricts linking code with other license. Even using its headers could be problematic as [wordings in license](https://github.com/numactl/numactl/blob/693fee1ed3d9524d2696f6ba021431b3176b6f57/LICENSE.LGPL2.1#L252-L257) look a bit vague. Above all, our usage was pretty simple (like cgroup, for which we don't use any lib either), so I ended up simplifying the support instead.

FYI: @ayakael, .NET 8+ build machines won't be needing libnuma anymore (after https://github.com/dotnet/runtime/pull/84207).

janvorli (approved):
LGTM, thank you!

PR #840, opened 2023-04-07
Upgrade mariner images to LLVM 16

sbomer:
As per title, upgrades LLVM from 12.0.1 to 16.0.0. Because this version isn't available in the mariner repos, we build it from source on all platforms (whereas we used to get 12.0.1 from the repos for the amd64 image).

LLVM 16 is the first release that has started to use C++17 (previous releases were using C++14). At first this was causing the compiler-rt builds to fail, because not all C++17 features were available in the libstdc++ version that is installed in the rootfs (and no version supporting these features is available on xenial without using pinned packages from later releases). However, as long as we are careful to build only the compiler-rt library we need (profile), no C++ is required.

Contributes to https://github.com/dotnet/runtime/issues/83428.


src/cbl-mariner/2.0/crossdeps-llvm/Dockerfile
am11, line 1: Is there any benefit of keeping crossdeps, crossdeps-llvm and crossdeps-builder separate? We usually split the layers when it's used by multiple images. If each part of these three layers are reused separately then it's fine. Otherwise, IMO, combine them to avoid 'too many moving parts'.

am11 (approved):

mthalman (commented):
Looks like src/cbl-mariner/2.0/cross/arm-alpine/Dockerfile is still referencing `cbl-mariner-2.0-crossdeps-arm-local` which should be updated to `cbl-mariner-2.0-crossdeps-llvm-local`

mthalman (approved):

janvorli (approved):
LGTM, thank you

PR #841, opened 2023-04-08
Install diffutils in mariner images

sbomer:
This makes `diff` available on the mariner build images. We would rather require `diff` as a dependency in our runtime build than `git`, since it sounds like we support source-building runtime without `git` installed per https://github.com/dotnet/runtime/pull/84148#discussion_r1161028942.

@janvorli @am11 PTAL


am11 (approved):
LGTM, thank you!

mthalman (approved):

PR #842, opened 2023-04-11
Build llvm using clang on mariner

sbomer:
Previously this was building with gcc 11.2.0 (which is installed as an indirect dependency of `git`, through `perl`). Building with gcc 11.2.0 hit this bug: https://github.com/dotnet/runtime/issues/84503.

Using clang to compile llvm avoids the issue described there. I also couldn't reproduce the crash with llvm 16 regardless of compiler used to build llvm. However, I figure it's best to use clang here to be safe.

@janvorli @mthalman PTAL


mthalman (approved):

janvorli (approved):
LGTM, thank you!

PR #849, opened 2023-04-14
Improve caching of mariner image builds

sbomer:
This changes the base of crossdeps-builder from crossdeps to the core mariner image. This way, adding dependencies to the crossdeps image doesn't require re-building LLVM. For example this should make changes like https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/847 much faster in the future.

I am hoping our infrastructure is also smart enough not to re-run the rootfs builds when `crossdeps` changes. The rootfs builds happen in a Dockerfile whose final output image is based on `crossdeps`, but the actual rootfs build step happens in a separate stage that doesn't depend on `crossdeps`. Even if that's not the case, though, this should at least avoid some LLVM re-builds.

### DIFF
--- a/src/cbl-mariner/2.0/crossdeps-builder/Dockerfile
+++ b/src/cbl-mariner/2.0/crossdeps-builder/Dockerfile
@@ -1,16 +1,24 @@
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-crossdeps-local
+FROM mcr.microsoft.com/cbl-mariner/base/core:2.0
 
 # TODO: remove this once debootstrap is available in the base repos.
 COPY mariner-extended.repo /etc/yum.repos.d/
 
 RUN tdnf install -y \
+        # Common utilities
+        ca-certificates \
+        git \
+        tar \
+        wget \
         # LLVM build dependencies
+        binutils \
         clang \
-        # Rootfs build dependencies
-        debootstrap \
-        # LLVM build dependencies
+        cmake \
+        diffutils \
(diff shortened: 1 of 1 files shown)


mthalman:
> I am hoping our infrastructure is also smart enough not to re-run the rootfs builds when `crossdeps` changes. The rootfs builds happen in a Dockerfile whose final output image is based on `crossdeps`, but the actual rootfs build step happens in a separate stage that doesn't depend on `crossdeps`.

The fact that the Dockerfiles dependent on crossdeps-builder don't technically need to be rebuilt is kind of unrelated to what is happening. It's because `crossdeps-builder` needs to be published and it's base has changed. So regardless of whether it had any dependents, it would always get rebuilt if it's base image changed. And yes, unfortunately the infrastructure doesn't support an image being local-only and not published. If we did, then the caching logic could be smart enough to skip building it.

mthalman:
> Right - this is what I am trying to address in this change, by making the base independent of crossdeps, so that future changes to crossdeps shouldn't trigger a re-build of crossdeps-builder. If I understand correctly this will work with our infrastructure.

Correct

> My second point was specifically about the downstream things that then depend on `crossdeps-builder` as one of the stages in a multi-stage build - I'm not sure what our infra will do there if `crossdeps-builder` doesn't change, but the base image of a different stage in the same Dockerfile does change.

With your changes, if crossdeps-builder hasn't changed then it will not rebuild it, even if crossdeps has changed.

mthalman (approved):

jkoritzinsky (approved):

PR #850, opened 2023-04-16
Install newer cmake to debian/11/gcc12 image

am11:
This installs cmake 3.25 in gcc-12 leg.


src/debian/11/gcc12/Dockerfile
```diff
@@ -25,6 +24,11 @@ RUN apt-get update && \
 
 ENV LANG=en_US.utf8
 
+RUN echo 'deb http://deb.debian.org/debian bullseye-backports main' >> /etc/apt/sources.list && \
```
am11, line 27:
```suggestion
// we want a higher version of cmake (3.20 or above) which is not available in the stable repo,
// so switching to stable-bpo repo
RUN echo 'deb http://deb.debian.org/debian bullseye-backports main' >> /etc/apt/sources.list && \
```

src/debian/11/gcc12/Dockerfile
```diff
@@ -25,6 +24,13 @@ RUN apt-get update && \
 
 ENV LANG=en_US.utf8
 
+// we want a higher version of cmake (3.20 or above) which is not available in the stable repo,
+// so switching to stable-bpo repo
```
am11, line 28:
```suggestion
# we want a higher version of cmake (3.20 or above) which is not available in the stable repo,
# so switching to stable-bpo repo
```

jkoritzinsky (approved):
Thank you so much!

PR #866, opened 2023-05-04
update msquic on Ubuntu 22

wfurt:
msquic 2.2 now support OpenSSL 3. Use published packages instead of building our own.


CarnaViire (approved):
LGTM, thanks!

mthalman (approved):

PR #868, opened 2023-05-05
Add an AlmaLinux-based image for us to use for our source-build legs

jkoritzinsky:
AlmaLinux is a better base image to use for our source-build legs as it is a rebuild of RHEL instead of a lead-ahead distro like CentOS Stream.

### DIFF
--- a/src/almalinux/8/source-build/Dockerfile
+++ b/src/almalinux/8/source-build/Dockerfile
@@ -1,0 +1,41 @@
+FROM almalinux:8
+
+# Install dependencies
+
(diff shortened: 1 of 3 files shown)


jkoritzinsky:
I started from the CentOS 7 source-build image.

Anything scl-related is not needed since we're using the in-box packages.

AlmaLinux only has the C locales in box, so we need to add the English locales.

Lttng has a different package name and needs the extra DNF feed (which needs the DNF command package).

Some of the other packages didn't have the same name, so I removed them. We can probably add back some of them (like the libunwind dev package) since I added the PowerTools feed.

PR #872, opened 2023-05-19
Add Mariner-based optimization images

sbomer:
For use in `dotnet-optimization` repo. `llvm-profdata` 16 is required to consume PGO data produced by llvm 16 instrumented binaries. Since we don't need a toolchain capable of cross-compiling, we can use the new packages available in Mariner.


jkoritzinsky (approved):

janvorli (approved):
LGTM, thank you!

PR #874, opened 2023-05-25
Add a Mariner-based Android build image

jkoritzinsky:
Validated with a local build that we can build the Android target in dotnet/runtime with this image.

This change combined with the WASM image addition will enable us to move all platforms we officially build for to build from a Mariner-based image.

### DIFF
--- a/src/cbl-mariner/2.0/android/Dockerfile
+++ b/src/cbl-mariner/2.0/android/Dockerfile
@@ -23,6 +23,7 @@ RUN tdnf update -y \
 
 # Install Android NDK
 RUN wget https://dl.google.com/android/repository/android-ndk-r23c-linux.zip \
+    && echo "e5053c126a47e84726d9f7173a04686a71f9a67a android-ndk-r23c-linux.zip" | sha1sum -c \
     && unzip android-ndk-r23c-linux.zip -d /usr/local \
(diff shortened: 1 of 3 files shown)


jkoritzinsky:
@akoeplinger this PR will downgrade the CMake version we use to 3.21 from 3.23. From what I was able to find, it looks like you updated the version to 3.23 to get a fix that came after 3.19. I think 3.21 should be fine, but I wanted to raise the visibility on that change to make sure there weren't any concerns.

jkoritzinsky:
This PR does not yet include a Mariner image that enables docker-in-docker (which the Grpc tests for Android needs).

We can keep using the old image for that leg as it isn't an official build scenario, but I'll first try to come up with a docker image that supports docker-in-docker for this scenario.

jkoritzinsky:
I've added checksum validation for the Android downloads. It doesn't look like Google provides signatures to verify, but I was able to find checksums for each component we're downloading.

akoeplinger:
> @akoeplinger this PR will downgrade the CMake version we use to 3.21 from 3.23. From what I was able to find, it looks like you updated the version to 3.23 to get a fix that came after 3.19. I think 3.21 should be fine, but I wanted to raise the visibility on that change to make sure there weren't any concerns.

Thanks, I'll keep an eye on any issues when this rolls out.

PR #890, opened 2023-06-09
add missing libnuma to Debian 11 packages 

wfurt:
fixes https://github.com/dotnet/runtime/issues/87275, workaround for https://github.com/microsoft/msquic/issues/3421
related to https://github.com/dotnet/runtime/issues/87181

### DIFF
--- a/src/debian/11/helix/amd64/Dockerfile
+++ b/src/debian/11/helix/amd64/Dockerfile
@@ -48,7 +48,7 @@ RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
     rm microsoft.asc && \
     apt-add-repository https://packages.microsoft.com/debian/11/prod && \
     apt-get update && \
-    apt-get install -y libmsquic && \
+    apt-get install -y libnuma1 libmsquic && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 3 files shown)


src/debian/11/helix/amd64/Dockerfile
```diff
@@ -48,7 +48,7 @@ RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
     rm microsoft.asc && \
     apt-add-repository https://packages.microsoft.com/debian/11/prod && \
     apt-get update && \
-    apt-get install -y libmsquic && \
+    apt-get install -y libnuma1 libmsquic && \
```
mthalman, line 51: Nit: I understand only two packages are here but might as well follow best practices right off the bat and have them listed alphabetically. My preference is to have them listed on separate lines as well. Same applies to other Dockerfiles.

```suggestion
    apt-get install -y \
      libmsquic \
      libnuma1 && \
```

wfurt: I can do that. this will hopefully be reversed soon as msquic fixes their packages e.g. this should not really be needed.

ManickaP (approved):

wfurt:
closing in favor of 2.2.2 upgrade.

PR #898, opened 2023-06-16
Add libatomic to Centos Stream 8 image

directhex:
The CentOS 8 image is used to run Mono LLVM AOT tests, in the interim, until Mono's AOT compiler is updated to use `llvm-as` instead of GNU Binutils `as`. However, LLVM 16 introduces a dependency on libatomic.so.1 which is satisfied on Mariner, but is absent on our CentOS Stream 8 image.

Fixing Mono's aot-compiler.c is the better long-term solution, but this unblocks test failures on the LLVM 16 integration work.

### DIFF
--- a/src/centos/stream8/Dockerfile
+++ b/src/centos/stream8/Dockerfile
@@ -26,6 +26,7 @@ RUN dnf install --setopt tsflags=nodocs --refresh -y \
         hostname \
         jq \
         krb5-devel \
+        # new dependency required when linking against libc++ instead of libstdc++ \
         libatomic \
         libcurl-devel \
         libedit-devel \
--- a/src/centos/stream8/Dockerfile
+++ b/src/centos/stream8/Dockerfile
@@ -26,6 +26,7 @@ RUN dnf install --setopt tsflags=nodocs --refresh -y \
         hostname \
         jq \
         krb5-devel \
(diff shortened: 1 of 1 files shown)


src/centos/stream8/Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN dnf install --setopt tsflags=nodocs --refresh -y \
         hostname \
         jq \
         krb5-devel \
+        libatomic \
```
jkoritzinsky, line 29: Can we add a comment about the dependency?

jkoritzinsky (approved):

src/centos/stream8/Dockerfile
directhex: Maybe! Let's see

directhex:
Is CI being weird? I don't see the mandatory `-nanoserver` job in the checks tab

directhex:
dotnet-buildtools-prereqs-docker-nanoserver hasn't run since September 2022. Maybe it's no longer appropriate as a required check on the repo?

mthalman:
I have no idea why GitHub thinks `dotnet-buildtools-prereqs-docker-nanoserver` should be run. Its trigger path is not applicable to this PR. I'll bypass the check.

PR #899, opened 2023-06-20
add ubuntu-20.04-arm64 image

oleksandr-didyk:
Contributes to https://github.com/dotnet/source-build/issues/2993

Adding an image for Ubuntu 20.04 ARM64 to be used for source-build CI builds


src/ubuntu/manifest.json
```diff
@@ -638,7 +638,7 @@
         {
           "platforms": [
             {
-              "dockerfile": "src/ubuntu/22.04/amd64",
+              "dockerfile": "src/ubuntu/22.04/",
```
mthalman, line 641: Nit: just for consistency

```suggestion
              "dockerfile": "src/ubuntu/22.04",
```

src/ubuntu/manifest.json
```diff
@@ -651,6 +651,23 @@
             }
           ]
         },
+        {
+          "platforms": [
+            {
+              "architecture": "arm64",
+              "dockerfile": "src/ubuntu/22.04/",
```
mthalman, line 658: Nit: just for consistency

```suggestion
              "dockerfile": "src/ubuntu/22.04",
```

src/ubuntu/manifest.json
```diff
@@ -651,6 +651,23 @@
...
+              "os": "linux",
+              "osVersion": "jammy",
+              "tags": {
+                "ubuntu-22.04-arm64-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "ubuntu-22.04-arm64$(FloatingTagSuffix)": {},
+                "ubuntu-22.04-arm64-local": {
+                  "isLocal": true
+                }
+              }
```
mthalman, line 667: The `local` tag can be removed. This is only needed if another Dockerfile in this repo needs to reference this as a base image.

Also adding the `variant` field.
```suggestion
              "tags": {
                "ubuntu-22.04-arm64-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
                "ubuntu-22.04-arm64$(FloatingTagSuffix)": {}
              },
              "variant": "v8"
```

src/ubuntu/manifest.json
oleksandr-didyk: Sure thing, thanks for noticing, fixed

src/ubuntu/manifest.json
oleksandr-didyk: Thanks for explaining! Removed the `local` tag

CI also failed on missing `variant`, added it to the declaration

oleksandr-didyk:
> Looks like the build is failing because the Azure CLI install helper script is getting an amd64-only DEB package
> This will need to be changed to not use the helper script and instead use the other installation method

Thanks, changed it up and tested locally, hopefully the CI won't fail this time

mthalman (approved):

PR #901, opened 2023-07-03
Add docker image w/ both Android SDK and an x64 crossroot

directhex:
(for AllSubsets jobs)


directhex:
Sigh. How do I get the intra-dependency to show up?

jkoritzinsky (approved):

jkoritzinsky:
You need to depend on the -local versions of the base images and add local tags for the base images in the manifest if they aren't there already.

directhex:
`write /var/lib/docker/tmp/GetImageBlob1305137427: no space left on device` I think it's time to hit the tequila

directhex:
This seems OK now? (thanks @lbussell)

Can I get a re-review @jkoritzinsky?

jkoritzinsky (approved):

PR #907, opened 2023-07-27
Install sudo in CBL crossdeps

am11:
Can't use `sudo` at the moment in these images.

cc @akoeplinger

### DIFF
--- a/src/cbl-mariner/2.0/crossdeps/Dockerfile
+++ b/src/cbl-mariner/2.0/crossdeps/Dockerfile
@@ -6,9 +6,9 @@ RUN tdnf update -y && \
         # Provides 'su', required by Azure DevOps
         ca-certificates \
         git \
+        sudo \
         util-linux \
         wget \
-        sudo \
 
         # Common runtime build dependencies
         awk \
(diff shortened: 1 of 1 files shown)


src/cbl-mariner/2.0/crossdeps/Dockerfile
```diff
@@ -6,6 +6,7 @@ RUN tdnf update -y && \
         # Provides 'su', required by Azure DevOps
         ca-certificates \
         git \
+        sudo \
```
akoeplinger, line 9: based on the comment above we might be able to just use `su` instead?

am11: After using `su -c ".." root`, CI is not outputting anything https://dev.azure.com/dnceng-public/public/_build/results?buildId=353710&view=logs&j=72449f43-ea21-55f2-2199-f125e9939b75&t=0040a3e2-424f-53a6-9149-dfb50282a3b1 (either it is running and not outputting anything, or `su` command is stuck)

We do install sudo in other images in this repo, though.

am11: su is a no go. Build is hanging for hours going nowhere. CBL-mariner is based on RedHat, so here are some details: https://access.redhat.com/solutions/432583. We don't have the HOME dir. Working around that requires elevated permission, so we are back to square one.

Lets install sudo, like we do elsewhere. (you search sudo in this repo or `runtime/eng/pipelines` to see the existing patterns)

am11:
cc @mthalman, this is needed to unblock https://github.com/dotnet/llvm-project/pull/456.

PR #911, opened 2023-08-17
Add new image for Ubuntu 22.04/helix/webassembly

radical:


radical:
@akoeplinger @mthalman is there any way of using the image generated in the CI build, on a test PR in `dotnet/runtime` so I can confirm that everything works? Or does the PR need to be merged first?

mthalman:
> @akoeplinger @mthalman is there any way of using the image generated in the CI build, on a test PR in `dotnet/runtime` so I can confirm that everything works? Or does the PR need to be merged first?

No, the PR build from this repo doesn't cause any images to be stored that are referenceable externally. You would need to build it yourself, push it to a registry that you have access to, and update the PR to reference the registry location.

radical:
I was able to test this by creating a container registry on azure, and using that to test the image on a PR in `dotnet/runtime`.

radical:
Does `    curl -fsSL https://deb.nodesource.com/setup_18.x | bash -` look acceptable for installing node, or should I download a tarball and use that? (and add any needed dependencies manually)

radical:
> > Does ` curl -fsSL https://deb.nodesource.com/setup_18.x | bash -` look acceptable for installing node, or should I download a tarball and use that? (and add any needed dependencies manually)
> 
> What does that do that installing `nodejs` doesn't do? Looking at other Dockerfiles in this repo, I see they just install `nodejs`.

This is the recommended way on nodejs.org (https://nodejs.org/en/download/package-manager#debian-and-ubuntu-based-linux-distributions). It adds apt sources, and then installs the package for that specific version, IIUC.

directhex:
`nodejs` in Ubuntu 22.04 is 12.22.9

We need a third party repo, or a from-source build, or I guess dogfooding our NuGet, for newer.

Ankit has followed the upstream instructions for installing newer. If we wanted we could extract the detail of what that's actually doing, to have a little less "pipe random downloads into root bash". It's a 400 line bash script which adds a relevant sources.list entry and GPG key to the system.

src/ubuntu/22.04/helix/webassembly/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+# WebAssembly build needs working UTF-8 locale
+RUN locale-gen en_US.UTF-8
+
+# install node, which is required for newer npm, jsvu and v8
+RUN cd ~ && \
+    curl -fsSL https://deb.nodesource.com/setup_18.x | bash - && \
+    apt-get -y install nodejs
+
+USER root
```
mthalman, line 43: This seems unnecessary since the user is already root at this point.

radical:
>  It's a 400 line bash script which adds a relevant sources.list entry and GPG key to the system.

I have extracted the relevant bits, and kept them in a separate file just so it is easier to update in future.

mthalman (approved):

PR #912, opened 2023-08-18
Default job timeouts to an integer value

mthalman:
When attempting to run a build for a pipeline that doesn't set the job timeout, it gets the following error:

```
/eng/common/templates/jobs/build-images.yml (Line: 24, Col: 21): Expected an integer value. Actual value: 'null'
```

This is a regression caused by the changes in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/901. The [default values for the timeouts are set to `null`](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/213a54af419383d9a8f1f7b3dc4b474c3af64e39/eng/pipelines/stages/build-test-publish-repo.yml#L5-L6) which is not a valid value.

Fixed by setting them to an integer value.

### DIFF
--- a/eng/pipelines/dotnet-buildtools-prereqs-alpine-pr.yml
+++ b/eng/pipelines/dotnet-buildtools-prereqs-alpine-pr.yml
@@ -24,3 +24,4 @@ stages:
   parameters:
     internalProjectName: ${{ variables.internalProjectName }}
     publicProjectName: ${{ variables.publicProjectName }}
+    linuxArmBuildJobTimeout: 120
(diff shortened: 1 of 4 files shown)


lbussell (approved):

lbussell:
`Linux_arm32 src-alpine-3.15-helix-arm32v7` ran longer than 60 minutes. Maybe it should be 90.

wfurt (approved):
LGTM.
It seems like the almalinux is failing because of missing `-y` for `yum update`

PR #913, opened 2023-08-18
Add images with both "host" and "target" crossroots in one image.

directhex:
This is required in some cases - e.g. compiling node.js for ARM64/Alpine, and the ever-broken ARM64 runtime perf lane (where we need headers for ARM64 available alongside linkable libraries for AMD64, depending on the build phase).

As is, we've achieved the same result largely by accident in the pre-Mariner build images, where we have enough x64 in the host OS to link successfully, and the ARM64 files in the crossroot.

The precedent on this is https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/901 which fixed our Android cross-compiler builds, by offering both Android and $HOST in the same system

If at some point in the future we move from building on x64 to building on ARM64, we will need the equivalent images for that.

### DIFF
--- a/src/cbl-mariner/2.0/cross/biarch/amd64-alpine/Dockerfile
+++ b/src/cbl-mariner/2.0/cross/biarch/amd64-alpine/Dockerfile
@@ -1,0 +1,7 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64-alpine-local AS crossrootbiarch
+
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64
+
+# Copy crossrootfs from Alpine build image, so we can build and cross-compile
+# cross compilers within the same build system (e.g. required for Node.js)
+COPY --from=crossrootbiarch /crossrootfs/x64 /crossrootfs/x64-alpine
--- a/src/cbl-mariner/2.0/cross/biarch/amd64-alpine/Dockerfile
+++ b/src/cbl-mariner/2.0/cross/biarch/amd64-alpine/Dockerfile
@@ -1,6 +1,6 @@
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64-alpine-local AS crossrootbiarch
 
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64-local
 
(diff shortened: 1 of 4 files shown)


src/cbl-mariner/2.0/cross/biarch/amd64-alpine/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64-alpine-local AS crossrootbiarch
+
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64
```
mthalman, line 3: You'll wanna use `local` here, and for the others:

```suggestion
FROM mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-amd64-local
```

directhex:
I've validated these images locally in https://github.com/dotnet/node/pull/36 - I was able to cross-compile for Musl for the first time, and arm64 Ubuntu on Mariner without gross accidental functioning via the old Ubuntu images.

src/cbl-mariner/manifest.json
```diff
@@ -318,6 +327,45 @@
...
+          ]
+        },
+        {
+          "platforms": [
+            {
+              "dockerfile": "src/cbl-mariner/2.0/cross/biarch/amd64-alpine",
+              "os": "linux",
+              "osVersion": "cbl-mariner2.0",
+              "tags": {
+                "cbl-mariner-2.0-cross-biarch-amd64-amd64-alpine$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
```
mthalman, line 363:
```suggestion
                "cbl-mariner-2.0-cross-biarch-amd64-amd64-alpine-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
```

mthalman (approved):

src/cbl-mariner/manifest.json
```diff
@@ -318,6 +327,45 @@
...
+        },
+        {
+          "platforms": [
+            {
+              "dockerfile": "src/cbl-mariner/2.0/cross/biarch/amd64-alpine",
+              "os": "linux",
+              "osVersion": "cbl-mariner2.0",
+              "tags": {
+                "cbl-mariner-2.0-cross-biarch-amd64-amd64-alpine-$(System:TimeStamp)-$(System:DockerfileGitCommitSha)": {},
+                "cbl-mariner-2.0-cross-biarch-amd64-amd64-alpine$(FloatingTagSuffix)": {}
```
akoeplinger, line 364: is the `amd64-amd64` a typo?

directhex:
No. amd64 primary crossrootfs, amd64-alpine secondary

Issue #903, opened 2023-07-20
Debian and Raspbian Helix images fail to build, asks for Rust to be installed


lbussell:
In https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/897, unrelated to the PR's changes, the Raspbian and Debian Helix stages failed to build:
https://dev.azure.com/dnceng-public/public/_build/results?buildId=343439&view=logs&j=3269ed96-e1ee-5791-373d-a55e5a2175a9&t=6d9bf06a-e7bf-5504-6e26-c5b767fa7b88

The specific error is:
```
> [stage-1 3/5] RUN ln -sf /usr/bin/python3 /usr/bin/python &&     curl https://bootstrap.pypa.io/get-pip.py -o ./get-pip.py --fail --silent --show-error &&     python ./get-pip.py && rm ./get-pip.py &&     python -m pip install --upgrade pip==20.2 &&     python -m pip install virtualenv==16.6.0 &&     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple &&     export CRYPTOGRAPHY_DONT_BUILD_RUST=1 &&     pip install ./helix_scripts-*-py3-none-any.whl:
#6 98.52   If you did intend to build this package from source, try installing a Rust compiler from your system package manager and ensure it is on the PATH during installation. Alternatively, rustup (available at https://rustup.rs/) is the recommended way to download and update the Rust compiler toolchain.
#6 98.52   
#6 98.52   This package requires Rust >=1.56.0.
#6 98.52   ----------------------------------------
#6 98.52   ERROR: Failed building wheel for cryptography
#6 98.52 Successfully built psutil pycparser cffi
#6 98.52 Failed to build cryptography
#6 98.52 ERROR: Could not build wheels for cryptography which use PEP 517 and cannot be installed directly
#6 98.57 WARNING: You are using pip version 20.2; however, version 23.2 is available.
#6 98.57 You should consider upgrading via the '/usr/bin/python -m pip install --upgrade pip' command.
```

It seems that the Helix install script doesn't respect the `CRYPTOGRAPHY_DONT_BUILD_RUST` environment variable that has been set:

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/af784859c2a9daa4322f2d6d0750f644cedd08d0/src/debian/11/helix/arm32v7/Dockerfile#L50-L57

[...]

PR #921, opened 2023-10-16
Update MsQuic on Ubuntu 18.04

CarnaViire:
cc @wfurt @ManickaP @mthalman

### DIFF
--- a/src/raspbian/10/helix/arm32v6/Dockerfile
+++ b/src/raspbian/10/helix/arm32v6/Dockerfile
@@ -49,7 +49,7 @@ ENV LANG=en_US.utf8
 RUN ln -sf /usr/bin/python3 /usr/bin/python && \
     curl https://bootstrap.pypa.io/get-pip.py -o ./get-pip.py --fail --silent --show-error && \
     python ./get-pip.py && rm ./get-pip.py && \
-    python -m pip install --upgrade pip==21.3.1 && \
+    python -m pip install --upgrade pip==20.2 && \
     python -m pip install virtualenv==16.6.0 && \
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
     export CRYPTOGRAPHY_DONT_BUILD_RUST=1 && \
--- a/src/raspbian/10/helix/arm32v6/Dockerfile
+++ b/src/raspbian/10/helix/arm32v6/Dockerfile
@@ -36,6 +36,8 @@ RUN apt-get update && \
         locales-all \
         python3-dev \
         python3-pip \
+        # rustc is needed to build cryptography wheel
+        rustc \
         sudo \
         tzdata \
         unzip \
--- a/src/raspbian/10/helix/arm32v6/Dockerfile
+++ b/src/raspbian/10/helix/arm32v6/Dockerfile
@@ -36,8 +36,6 @@ RUN apt-get update && \
         locales-all \
         python3-dev \
         python3-pip \
-        # rustc is needed to build cryptography wheel
-        rustc \
         sudo \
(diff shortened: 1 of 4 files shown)


ManickaP (approved):

mthalman (approved):

mthalman:
@lbussell - It appears the Ubuntu builds are broken unrelated to these changes.

CarnaViire:
I assume this is the same problem as https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/903 😢

lbussell:
Taking a look at the failure.

wfurt:
I think this needs never node @lbussell. 
```
#7 [4/7] RUN npm install -g npm
#7 0.775 npm ERR! code EBADENGINE
#7 0.777 npm ERR! engine Unsupported engine
#7 0.777 npm ERR! engine Not compatible with your version of node/npm: npm@10.2.0
#7 0.777 npm ERR! notsup Not compatible with your version of node/npm: npm@10.2.0
#7 0.777 npm ERR! notsup Required: {"node":"^18.17.0 || >=20.5.0"}
#7 0.777 npm ERR! notsup Actual:   {"npm":"8.19.2","node":"v18.12.1"}
#7 0.779 
```

lbussell:
> @wfurt: I think this needs newer node

Yeah, that was it, got it building locally so let's hope that's the only issue.

lbussell:
> The Raspbian may be different issue. I was trying to figure out of I can build arm32 on arm64.... and in mean time I'm refreshing my Raspberry so I can possibly build arm32 natively.

I didn't see a Raspbian issue, don't see it in the [public pipeline runs](https://dev.azure.com/dnceng-public/public/_build?definitionId=57) either. You should be able to build arm32 on arm64 by using the `--platform linux/arm32/v7` argument on the build command.

lbussell:
OK, trying to fix that issue by adding `rustc`. From my research, the `CRYPTOGRAPHY_DONT_BUILD_RUST` env variable might have been a temporary workaround. If we run into compilation issues (which is likely), we should try updating the pip version, which is the official recommendation from [cryptography](https://github.com/pyca/cryptography). I'm trying this first because I don't know if helix relies on a specific pip version.

wfurt:
I have access to arm64 build machine so I can try more if needed @lbussell. 
Some images use `pip==22.0.4` and `virtualenv==20.14.0` so I would expect that should be good enough. 
If needed I would suggest to lock it on that version to decrease variations.

wfurt:
actually, Ubuntu 18.04 is EOL https://ubuntu.com/blog/ubuntu-18-04-eol-for-devices
Instead of trying to wrestle unsupported OS we should perhaps move to 20.04 or 22.04.

lbussell:
It'd be great to drop 18.04 - but don't we still need to keep this image buildable to support .NET 6 and 7?

lbussell:
Actually, this 18.04 image isn't even being updated in this PR, the image exists in MCR already, it won't get any more security updates, so we should just delete it. If Raspbian 10 is blocking any other PRs we should consider a similar approach there.

lbussell:
@wfurt, @CarnaViire, @ManickaP, would moving the failing runtime tests to 20.04/22.04 unblock your PR?

lbussell:
So @wfurt, @CarnaViire, @ManickaP, this is building now, but I would still probably recommend moving the 8.0/main tests off 18.04 especially if we aren't supporting it for 8.0/9.0. These Dockerfiles should all be removed soon.

CarnaViire:
@lbussell Debian 10 and Debian 11 Arm images are [also failing](https://dev.azure.com/dnceng/internal/_build/results?buildId=2295639&view=results) -- I assume the same changes should be applied to them? I'll try to put up a PR with that

PR #924, opened 2023-10-18
Remove EOL Ubuntu 18.04 images

lbussell:
Per https://ubuntu.com/blog/ubuntu-18-04-eol-for-devices, Ubuntu 18.04 is "ESL" and won't receive publicly available support.


wfurt:
It may be worth of waiting until we are sure everything is ported to updated version. While one can search history, copying files is easiest IMHO.

PR #925, opened 2023-10-19
Fix Debian Arm32 build

CarnaViire:
Port changes from Ubuntu Arm32 (https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/921) to Debian 10 and 11.

Contributes to https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/903

cc @wfurt @lbussell @mthalman

PR #927, opened 2023-10-25
Unblock Debian build by disabling arm32 images

lbussell:
- Disable arm32 debian images
- Also disable raspbian

Context: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/925#issuecomment-1771743023

Debian builds stopped working because the python Cryptography wheel doesn't ship binaries for arm32. It tries to compile itself from source. In Ubuntu, we can do that by installing Rust. Debian's Rust version isn't new enough to compile Cryptography. The fix will be more involved so for now we need to disable Debian builds by removing the images from the manifest. This will unblock builds and PRs in this repo.

### DIFF
--- a/src/debian/manifest.json
+++ b/src/debian/manifest.json
@@ -20,21 +20,6 @@
         {
           "platforms": [
             {
-              "architecture": "arm",
-              "dockerfile": "src/debian/10/helix/arm32v7",
-              "os": "linux",
-              "osVersion": "buster",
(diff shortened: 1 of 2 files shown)


wfurt (approved):
LGTM. 
I still plan to look at the ARM32 builds as time permits later this or next week.

mthalman (approved):

PR #923, opened 2023-10-18
refresh alpine images

wfurt:
this change removes everything bellow 3.15 and adds missing variants for 3.16 and 3.18. 
3.15 will be EOS very soon but I decided to keep it for now. 

3.16 images are copied from 3.15 and 3.18 from 3.17. 

This will allow runtime tests to run on oldest and newest supported Alpine releases. 

cc: @CarnaViire

PR #931, opened 2023-11-08
fix msquic build on Alpine 3.16

wfurt:
this is follow-up on #923.
It seems like the image I cloned from was locked on `release/7.0` so we are building really old msquic. 
Since I'm touching this, I also removed dependency on PowerShell to get it on par with other architectures. 
(we had to do it for ARM as PowerShell does not offer Alpine ARM binaries)

### DIFF
--- a/src/alpine/3.16/helix/amd64/Dockerfile
+++ b/src/alpine/3.16/helix/amd64/Dockerfile
@@ -55,16 +55,45 @@ RUN ln -sf /usr/bin/python3 /usr/bin/python && \
     pip install ./helix_scripts-*-py3-none-any.whl
 
 # build MsQuic as we don't have packages
-RUN cd /tmp && \
-    mkdir pwsh && \
-    cd pwsh && \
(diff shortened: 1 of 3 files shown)


wfurt:
I added `ldd` as extra step. It should fail the build if the produced library is not loadable
```
bash-5.1$ ldd /usr/lib/libmsquic.so.2
	/lib/ld-musl-x86_64.so.1 (0x7fc4d84f9000)
	libcrypto.so.1.1 => /lib/libcrypto.so.1.1 (0x7fc4d8120000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7fc4d84f9000)
Error relocating /usr/lib/libmsquic.so.2: EVP_seed_cbc: symbol not found
bash-5.1$ echo $?
127
```

It should availed late surprised like https://github.com/dotnet/runtime/pull/94487

CarnaViire (approved):
LGTM, thanks!

lbussell (approved):

PR #933, opened 2023-11-24
Pin msquic to pre-regression revision on Alpine Arm32

CarnaViire:
Contributes to https://github.com/dotnet/runtime/issues/91757

While pinning to a docker tag temporarily mitigated the issue above, it doesn't allow us to consume any of the docker image updates, that are unrelated to msquic, e.g. see https://github.com/dotnet/dnceng/issues/1423#issuecomment-1823466181.

Changing it to pin to a pre-regression msquic revision instead of main, so we can "unlock" the image in runtime CI again.

cc @wfurt

### DIFF
--- a/src/alpine/3.15/helix/arm32v7/Dockerfile
+++ b/src/alpine/3.15/helix/arm32v7/Dockerfile
@@ -54,7 +54,7 @@ RUN apk update && apk add --no-cache && \
         openssl-dev \
         perl && \
     cd /tmp && \
-    git clone --depth 1 --single-branch --branch alpine-arm32-pin-2023-08 --recursive https://github.com/dotnet/msquic && \
+    git clone --depth 1 --single-branch --branch msquic-v2.2.2 --recursive https://github.com/dotnet/msquic && \
     cd msquic/src/msquic && \
     cmake -B build/linux/arm_openssl \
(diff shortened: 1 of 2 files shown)


src/alpine/3.15/helix/arm32v7/Dockerfile
```diff
@@ -54,7 +54,7 @@ RUN apk update && apk add --no-cache && \
         openssl-dev \
         perl && \
     cd /tmp && \
-    git clone --depth 1 --single-branch --branch main --recursive https://github.com/dotnet/msquic && \
+    git clone --depth 1 --single-branch --branch alpine-arm32-pin-2023-08 --recursive https://github.com/dotnet/msquic && \
```
wfurt, line 57: I would probably much rather have branch that tracks MsQuic v2.2.4 (or whatever release) than some random point in time. That would also give us coverage on code used by our customers.

CarnaViire: We can pin to 2.2.2 (June 2023), 2.2.3+ already has the regression. Let's try

wfurt:
it looks like all checks finished but the dotnet-buildtools-prereqs-docker-windowsservercore did not get updated.

wfurt:
can you override the checks @mthalman @lbussell ??? It seems like everything passed.

PR #940, opened 2023-12-11
Added in docker image for ML.Net for centos-stream8

michaelgsharp:
Adding in a docker image for ML.NET for Centos-stream8. It also sets the base centos-stream8 to local so that I can directly use it as the starter point.


ericstj (approved):
LGTM

PR #957, opened 2024-01-16
Attempt to fix sourcebuild issues

wtgodbe:
Attempt to fix https://github.com/dotnet/source-build/issues/3925, https://github.com/dotnet/source-build/issues/3924

### DIFF
--- a/src/fedora/39/amd64/Dockerfile
+++ b/src/fedora/39/amd64/Dockerfile
@@ -67,3 +67,4 @@ RUN dnf --setopt=install_weak_deps=False install -y \
 RUN cd ~ && \ 
     curl -sL https://rpm.nodesource.com/setup_20.x
 RUN dnf install -y nodejs && dnf clean all
+ENV NO_UPDATE_NOTIFIER
--- a/src/ubuntu/22.04/cross/arm64/Dockerfile
+++ b/src/ubuntu/22.04/cross/arm64/Dockerfile
@@ -20,4 +20,7 @@ RUN cd ~ && \
     curl -sL https://deb.nodesource.com/setup_20.x -o nodesource_setup.sh && \
     bash nodesource_setup.sh && \
     apt-get install nodejs -y && \
-    rm -rf /var/lib/apt/lists/*
+    rm -rf /var/lib/apt/lists/* && \
+    rm -f nodesource_setup.sh
+
+ENV NO_UPDATE_NOTIFIER=true
--- a/src/ubuntu/22.04/cross/arm64/Dockerfile
+++ b/src/ubuntu/22.04/cross/arm64/Dockerfile
@@ -4,13 +4,12 @@ FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-22.04-crossdeps-local
 RUN apt-get update \
     && apt-get install -y \
         binutils-aarch64-linux-gnu \
-        nodejs \
     && rm -rf /var/lib/apt/lists/*
 
 ADD rootfs.arm64.tar crossrootfs
 
-# update node
-ENV NODE_VERSION 20.9.0
+# install node
 RUN cd ~ && \
-    curl -LO https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-arm64.tar.xz && \
-    tar xf node-v${NODE_VERSION}-linux-arm64.tar.xz
+    curl -sL https://deb.nodesource.com/setup_20.x -o nodesource_setup.sh && \
+    sudo bash nodesource_setup.sh && \
+    sudo apt-get install nodejs -y
--- a/src/ubuntu/22.04/cross/arm64/Dockerfile
+++ b/src/ubuntu/22.04/cross/arm64/Dockerfile
@@ -10,12 +10,12 @@ ADD rootfs.arm64.tar crossrootfs
 
 # Remove older version of node & install node 20
 RUN cd /etc/apt/sources.list.d  && \
-    ls && \
     rm -f nodesource.list && \
     apt --fix-broken install && \
     apt update && \
     apt-get remove nodejs -y && \
-    apt-get remove nodejs-doc -y
+    apt-get remove nodejs-doc -y && \
+    apt-get remove libnode-dev -y
 RUN cd ~ && \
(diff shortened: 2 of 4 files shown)


MichaelSimons (approved):

src/ubuntu/22.04/cross/arm64/Dockerfile
```diff
@@ -4,13 +4,13 @@ FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-22.04-crossdeps-local
...
-ENV NODE_VERSION 20.9.0
+# install node
 RUN cd ~ && \
-    curl -LO https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-arm64.tar.xz && \
-    tar xf node-v${NODE_VERSION}-linux-arm64.tar.xz
\ No newline at end of file
+    curl -sL https://deb.nodesource.com/setup_20.x -o nodesource_setup.sh && \
+    sudo bash nodesource_setup.sh && \
+    sudo apt-get install nodejs -y && \
+    rm -rf /var/lib/apt/lists/*
```
ViktorHofer, line 16: Should we also rm the `nodesource_setup.sh` or doesn't it matter?

ViktorHofer (approved):
Out of curiosity, what's the node&npm version on Fedora 39? And I assume you tested both changes?

src/fedora/39/amd64/Dockerfile
```diff
@@ -66,4 +66,5 @@ RUN dnf --setopt=install_weak_deps=False install -y \
 # Dependencies for Aspnetcore
 RUN cd ~ && \ 
     curl -sL https://rpm.nodesource.com/setup_20.x
-RUN dnf install -y nodejs && dnf clean all
\ No newline at end of file
+RUN dnf install -y nodejs && dnf clean all
+ENV NO_UPDATE_NOTIFIER
```
mthalman, line 70: I think this needs to be set to a value.
```suggestion
ENV NO_UPDATE_NOTIFIER=true
```

ViktorHofer: According to the docs it doesn't need to be set to any value but it doesn't harm: https://www.npmjs.com/package/update-notifier#user-settings

wtgodbe:
> Out of curiosity, what's the node&npm version on Fedora 39? And I assume you tested both changes?

I don't work with Docker much, so normally I use the images that get produced by the build of this repo - how can I test them out from the sources?

src/ubuntu/22.04/cross/arm64/Dockerfile
wtgodbe: @mthalman what do you think?

mthalman: Might as well. It won't be needed in the final image. I'm a fan of image cleanliness.

mthalman:
> > Out of curiosity, what's the node&npm version on Fedora 39? And I assume you tested both changes?
> 
> I don't work with Docker much, so normally I use the images that get produced by the build of this repo - how can I test them out from the sources?

[Instructions for building](https://github.com/dotnet/dotnet-buildtools-prereqs-docker?tab=readme-ov-file#how-to-modify-or-create-a-new-image)

mthalman:
```
#9 7.066 Unpacking nodejs (20.11.0-1nodesource1) ...
#9 8.642 dpkg: error processing archive /var/cache/apt/archives/nodejs_20.11.0-1nodesource1_amd64.deb (--unpack):
#9 8.642  trying to overwrite '/usr/include/node/common.gypi', which is also in package libnode-dev 12.22.9~dfsg-1ubuntu3
#9 8.662 dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
#9 8.671 Errors were encountered while processing:
#9 8.671  /var/cache/apt/archives/nodejs_20.11.0-1nodesource1_amd64.deb
#9 8.690 E: Sub-process /usr/bin/dpkg returned an error code (1)
```

wtgodbe:
That's the same error as https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/957#issuecomment-1894307239, @ViktorHofer it looks like the node uninstall isn't working, or is incomplete?

src/fedora/39/amd64/Dockerfile
```diff
@@ -66,4 +66,5 @@ RUN dnf --setopt=install_weak_deps=False install -y \
 # Dependencies for Aspnetcore
 RUN cd ~ && \ 
     curl -sL https://rpm.nodesource.com/setup_20.x
-RUN dnf install -y nodejs && dnf clean all
\ No newline at end of file
+RUN dnf install -y nodejs && dnf clean all
+ENV NO_UPDATE_NOTIFIER=true
```
mthalman, line 70: This should be applied consistently across all the Node-related Dockerfiles.

ViktorHofer:
The uninstall worked for me locally on my Ubuntu 22.04 machine. Unsure what's required for this dockerfile.

wtgodbe:
Alright, I think this is working now & feedback has been addressed.

src/fedora/39/amd64/Dockerfile
wtgodbe: Done

src/ubuntu/22.04/cross/arm64/Dockerfile
wtgodbe: Done

mthalman (approved):

src/ubuntu/22.04/cross/arm64/Dockerfile
mthalman, line 1: @wtgodbe - Was it intentional to apply this to the crossbuild Dockerfile? This isn't used by source build.

wtgodbe: Oh, I assumed this file corresponded to `ubuntu-22.04-arm64`, but I guess https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/ubuntu/22.04/Dockerfile covers both the regular ubuntu-22.04 image & the arm64 one. Should we revert these changes then?

mthalman: Yes, the change should be reverted and applied to the other Dockerfile. I can take care of it. I just wanted to confirm.

PR #966, opened 2024-01-24
Add llvm-symbolizer to all more helix images

hoyosjs:


jkoritzinsky (approved):
Do we want to/can we copy the live-built llvm-symbolizer from our cross-build image here so we can ensure our toolchain is consistent?

If that's not possible, I understand. LGTM in that case

hoyosjs:
@mthalman are these build issues known? They seem to be in npm and Python, but rebuilds didn't help

mthalman:
> @mthalman are these build issues known? They seem to be in npm and Python, but rebuilds didn't help

@lbussell

lbussell:
> @mthalman are these build issues known? They seem to be in npm and Python, but rebuilds didn't help

it's https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/903. https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/929/files solved this for debian, but there's no binary distribution of python3-cryptography for Alpine. I think we would need to follow the troubleshooting steps in [cryptography's documentation under "Rust"](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/929/files).

lbussell:
Actually I lied and found the package. We should try installing [`py3-cryptography`](https://pkgs.alpinelinux.org/package/v3.16/community/armv7/py3-cryptography).

hoyosjs:
@lbussell adding that package requires gcc, but that might make the container pretty big. Not sure if purging gcc after is an option, or if we should just let it be

JulieLeeMSFT:
Hi @lbussell, we need your help to fix this issue. We are trying to roll out the feature by Feb, so we need this done ASAP.

lbussell:
@JulieLeeMSFT These Dockerfiles are the responsibility of repo owners to maintain (e.g. runtime devs) since they are the devs who need them. I did my best effort to try and fix the problem in my spare time.

I see two blocking issues in this PR:
1. An issue with `npm` in an Ubuntu Dockerfile. I haven't taken a look at this. It looks unrelated to @hoyosjs's change unfortunately.
2. An issue with Helix dependencies in an Alpine arm32 Dockerfile. I tried to fix this based on our own experience with similar issues in the past. /cc @epananth @davfost for help with Helix dependency issues and @wfurt who had helped fix this issue for Debian.

If it's an option, we could disable the images that fail to build. But since @hoyosjs touched the files that are failing to build I'm guessing they're important.

epananth:
Probably @dougbu has some insights on this?

dougbu:
I generally agree w/ @lbussell. we haven't found viable workarounds for the `cffi` on ARM32 issue due to resource constraints. installing C++ and Rust, then removing them _might_ be the best option. right now, it seems we haven't found the perfect set of commands. continuing to experiment here seems like the best option

lbussell:
Submitted https://github.com/dotnet/runtime/pull/98531 so that we can move towards deleting the offending alpine 3.16 image.

hoyosjs:
@radical @radekdoulik I changed the versions to pin things to the last working versions for 18.04. That being said, I can't see a place that uses the 18.04 webassembly image. Can it be removed? the node version and dependencies are no longer maintained.

PR #967, opened 2024-02-12
[wasi] Update wasi-sdk to version 21.

vargaz:
Contributes to https://github.com/dotnet/runtime/issues/98315

### DIFF
--- a/src/ubuntu/20.04/webassembly/Dockerfile
+++ b/src/ubuntu/20.04/webassembly/Dockerfile
@@ -49,7 +49,7 @@ RUN curl -sSL https://netcorenativeassets.blob.core.windows.net/resource-package
     && chmod +x /usr/local/bin/v8
 
 # Install Wasi toolchain
-ENV WASI_SDK_VERSION=20
+ENV WASI_SDK_VERSION=21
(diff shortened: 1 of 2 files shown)


vargaz:
Re: https://github.com/dotnet/runtime/pull/98120

pavelsavara:
cc @lewing

src/ubuntu/20.04/webassembly/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN curl -sSL https://netcorenativeassets.blob.core.windows.net/resource-package
     && chmod +x /usr/local/bin/v8
 
 # Install Wasi toolchain
-ENV WASI_SDK_VERSION=20
+ENV WASI_SDK_VERSION=21
```
akoeplinger, line 52: I don't think we use this container anymore actually. We could probably delete it.

akoeplinger (approved):

akoeplinger:
Note that we're currently pinning an older version of the container here: https://github.com/dotnet/runtime/blob/4f099091aa1dbcc3db0043241f3a4e81ef5abfdd/eng/pipelines/common/templates/pipeline-with-resources.yml#L88-L96

PR #969, opened 2024-02-16
Add an Ubuntu-22.04-based Tizen build image

jkoritzinsky:
The 18.04-based image we use in PRs in dotnet/runtime is causing warnings about the Node version in use. By upgrading to 22.04, we can fix this.

Additionally, other platforms that Samsung supports like RISCV also use 22.04-based images, so this won't break them or test a scenario they wouldn't support.


hoyosjs (approved):

hoyosjs:
Assume we no longer need the hook to the build-scripts since you clone and build the rootfs on it's own layer?

jkoritzinsky:
Yep. We can replace all of the buildrootfs hooks with muti-phase builds.

PR #970, opened 2024-02-22
Remove Alpine 3.16 Arm32v7 Helix image in order to unblock the build

lbussell:
Alpine 3.16 Arm32v7 Helix image does not build due to https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/903. This PR removes it in order to unblock the build since there is no obvious solution.

There is work in progress to move off of the image in https://github.com/dotnet/runtime/pull/98531.


mthalman (commented):
Based on your analysis of the usage data, it seemed like the other two architectures for Alpine 3.16 were unused. If so, those can be deleted too.

lbussell:
> Based on your analysis of the usage data, it seemed like the other two architectures for Alpine 3.16 for Helix were unused. If so, those can be deleted too.

I don't think that's the case. I see usage of alpine3.16-helix Amd64 images in both runtime and aspnetcore, and Arm64 is also used in runtime's outer-loop testing.

epananth (approved):

mthalman (approved):

mthalman:
Now that this is merged, can AutoBuilder be setup to be resumed?

PR #995, opened 2024-04-03
Add azurelinux 3.0 images

sbomer:
Add Azure Linux 3.0 preview images, with the same rootfs setup as the current cbl-mariner-2.0 images, using Ubuntu 16.04. ~We plan to introduce Ubuntu 18.04 rootfs for .NET 9 builds, so the naming scheme includes the rootfs ubuntu version (for example: azurelinux-3.0-cross-amd64-ubuntu-16.04).~

This uses a naming scheme with a -net8.0 suffix. We plan to introduce .NET 9 images with a -net9.0 suffix, that use LLVM 18 and an Ubuntu 16.04 rootfs (if we can find a way to build LLVM 18 there), since we're still aiming to target 16.04 in .NET 9 per https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/995#issuecomment-2036124345:

So image names will look like the following:
- azurelinux-3.0-cross-amd64-net8.0
- azurelinux-3.0-cross-amd64-net9.0

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,8 +2,8 @@
   "registry": "mcr.microsoft.com",
   "includes": [
     "src/almalinux/manifest.json",
-    "src/azurelinux/manifest.json",
     "src/alpine/manifest.json",
+    "src/azurelinux/manifest.json",
     "src/cbl-mariner/manifest.json",
     "src/centos/manifest.json",
     "src/debian/manifest.json",


src/azurelinux/3.0/cross/amd64-ubuntu-16.04/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+        -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" \
+        -DCMAKE_ASM_COMPILER_TARGET="${TARGET_TRIPLE}" \
+        -DCMAKE_C_COMPILER_TARGET="${TARGET_TRIPLE}" \
+        -DCMAKE_CXX_COMPILER_TARGET="${TARGET_TRIPLE}" \
+        -DCMAKE_ASM_FLAGS="${BUILD_FLAGS}" \
+        -DCMAKE_C_FLAGS="${BUILD_FLAGS}" \
+        -DCMAKE_CXX_FLAGS="${BUILD_FLAGS}" && \
+    make -j $(getconf _NPROCESSORS_ONLN)
+
+RUN LLVM_VERSION=16.0.0 LLVM_VERSION_MAJOR="${LLVM_VERSION%%.*}" SANITIZER_RUNTIMES_DIR=/sanitizer-runtimes && \
```
richlander, line 37: LLVM 16 makes sense for .NET 8 since that's what we started with.

18 is already out. We should decide our approach for .NET 9+. Should we definite a peer set of these images that includes LLVM 18? I'm guessing by the time .NET 10 comes around that we'll be looking at LLVM 20.

src/azurelinux/3.0/crossdeps-builder/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+RUN tdnf install -y \
+        # Common utilities
+        ca-certificates \
+        git \
+        tar \
+        wget \
+        # LLVM build dependencies
+        binutils \
+        clang \
+        cmake \
```
richlander, line 12: We should double check with @jkoritzinsky on the CMake version.

src/azurelinux/3.0/crossdeps/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN mkdir /usr/share/node && \
+    curl -O https://nodejs.org/dist/v${NODE_VERSION}/SHASUMS256.txt && \
+    curl -O -sSL https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.xz && \
+    grep node-v${NODE_VERSION}-linux-x64.tar.xz SHASUMS256.txt | sha256sum -c - && \
+    tar xf node-v${NODE_VERSION}-linux-x64.tar.xz -J -C /usr/share/node --strip-components=1 && \
+    ln -s /usr/share/node/bin/node /usr/local/bin/node && \
+    ln -s /usr/share/node/bin/npm /usr/local/bin/npm && \
+    ln -s /usr/share/node/bin/npx /usr/local/bin/npx && \
+    rm SHASUMS256.txt && \
+    rm node-v${NODE_VERSION}-linux-x64.tar.xz
```
jkoritzinsky, line 38: Azure Linux 3.0 should have node 20 available in the `nodejs` package so we shouldn't have to manually download it.

richlander: That'd sure be nice.

src/azurelinux/3.0/crossdeps-builder/Dockerfile
jkoritzinsky: This cmake version should be all good (3.28.2)

src/azurelinux/3.0/crossdeps/Dockerfile
richlander: I see this version: `20.10.0-1.azl3`

richlander:
> We plan to introduce Ubuntu 18.04

We should hold off on that one. That _was_ the original thinking. I think we've concluded that we'll stay the course on 16.04 for at least another release, if not two. We got feedback that made us pause on this one.

@jkotas

src/azurelinux/3.0/crossdeps/Dockerfile
jkoritzinsky: That should be fine. We only needed node20 for some AzDO tasks and for aspnetcore (in the VMR build).

richlander: This is where we should start to partition stuff.

Does ASP.NET Core need to use these images? These images should be EXCLUSIVELY used runtime and other teams who have a native build (with overlapping requirements). We need to break teams using these images who have managed only builds and not accept any of their requests for changes.

I'm going to make a managed only image for teams. It isn't entirely obvious what should be in there yet.

Does `runtime` need those AzDo tasks or is that for other repos/teams?

jkoritzinsky: I think runtime needs it for Component Governance, but I'm not positive.

aspnetcore wouldn't be using this image except for in the VMR. We could have separate images for the VMR builds that include the dependencies for all constituent projects if we wanted to keep the separation.

richlander: Ah yes, VMR. Sven and I talked about that. OK, we probably need it then. Reality is often annoying.

My point about moving repo (non-VMR) builds to another (MUCH) smaller image still stand.

jkotas: Does aspnetcore run any official builds on non-Windows platforms?

> Reality is often annoying.

Right. The current high-level breakdown should be like this:
- If the repo is managed-only, it should not need to run official builds outside Windows at all. Our builds are Windows centric for historic reasons. An exception are source build sanity checks that run on [ubuntu image in CI and 1es-mariner-2 image in internal builds](https://github.com/dotnet/arcade/blob/ce6cb9b392e6a534ffab4e8d9d2dbadcc5491b5a/eng/common/templates-official/job/source-build.yml#L49-L56).
- If the repo builds native components (are there any repos like that outside runtime?), it needs more or less the same image as runtime for official or CI build.

mthalman:
@lbussell - Be aware of the new pipelines in this PR that will need to be wired up.

src/azurelinux/3.0/crossdeps/Dockerfile
jkoritzinsky: With the VMR, we'll now be running official builds for each constituent repo on each platform (as part of the VMR's build command).

Outside the VMR, I believe ASP.NET Core runs the builds that produce the non-Windows runtime packs on those non-Windows platforms. I believe they do so as it matches the innerloop dev experience well and then they don't have to build custom infrastructure for "build on Windows for non-Windows" into their repo (as they do have some Windows-only assets)

lbussell:
> @lbussell - Be aware of the new pipelines in this PR that will need to be wired up.

@mthalman, it seems like the PR validation did run on this PR, so we should merge this and then set up the public and internal pipelines afterwards, right? So that the file is present to create the pipeline against.

src/azurelinux/3.0/cross/amd64-ubuntu-16.04/Dockerfile
jkoritzinsky: I'd prefer -llvmXX over -netXX as it's more descriptive of what's contained in the package.

mthalman:
> > @lbussell - Be aware of the new pipelines in this PR that will need to be wired up.
> 
> @mthalman, it seems like the PR validation did run on this PR, so we should merge this and then set up the public and internal pipelines afterwards, right? So that the file is present to create the pipeline against.

Yep, exactly

src/azurelinux/3.0/cross/amd64-ubuntu-16.04/Dockerfile
jkotas: > I'd prefer -llvmXX over -netXX as it's more descriptive of what's contained in the package.

It is impossible to fully describe what's in the image in the name. I expect that we would need the keep tweaking the naming scheme as new requirements show up.

`-netXX` naming is more durable. It simply means "image used to build .NET XX release" whatever that means.

jkoritzinsky: The `-netXX` scheme is fine with me.

richlander: The advantage of that scheme is that it gives people a hint that a `net80` image probably should NOT still be in use in a .NET 9 or 10 build. Or ... I'm hopeful of that.

src/azurelinux/3.0/crossdeps/Dockerfile
richlander: I started this "spec" to get us going on a plan. There are a TON of patterns in pipelines today. We need to reduce those a fair bit. Some are quite specialized, like in `dotnet/installer` but others seem needlessly DIY.

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/997

manifest.json
```diff
@@ -2,6 +2,7 @@
   "registry": "mcr.microsoft.com",
   "includes": [
     "src/almalinux/manifest.json",
+    "src/azurelinux/manifest.json",
```
mthalman, line 5: Move this down one line for alphabetical order.

mthalman (approved):

PR #1006, opened 2024-04-16
Update mariner build timeout

mthalman:
The latest build finished one minute shy of hitting the timeout (whew!). So I'm increasing it some more.


MichaelSimons (commented):
Is 30 minutes enough?  That's less than 10% variability.

mthalman:
I bumped it up.

MichaelSimons (approved):

PR #1007, opened 2024-04-17
Install libmsquic in CentOS Stream 9 image

richlander:
For https://github.com/dotnet/runtime/pull/101073#issuecomment-2059427197

I couldn't figure out how to use the referenced RHEL 9 image. If there is a good pattern for to that, we can adopt that.

@wfurt


richlander:
Oops. I had added the `libmsquic` package and then started again and forgot to add it back.

Your change fails.

```bash
No such command: https://packages.microsoft.com/rhel/9/prod. Please use /usr/bin/dnf --help
It could be a DNF plugin command, try: "dnf install 'dnf-command(https://packages.microsoft.com/rhel/9/prod)'"
```

wfurt:
yep. cut too much ;(

wfurt (approved):
LGTM

PR #1008, opened 2024-04-18
Remove Ubuntu 22.04 cross build

richlander:
These Dockerfiles were put in place for .NET 7, AFAIK.

This PR should not be merged until .NET 7 goes EOL.


am11:
The haiku, ppcle64le-alpine, riscv, riscv-alpine and s390x-alpine would need to be moved to CBL2 or AZL3. We can leave them until they are moved to the new plan.

richlander:
If community projects are using this, they we can just close this change. I don't think we're trying to force community projects to use mariner/Azure Linux.

PR #1010, opened 2024-04-19
Get centos:7 images updated

### LINKED ISSUE #988: Automate updating images that have infrequent base image updates
Let's take CentOS 7 for example: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/centos/7/Dockerfile

[CentOS 7's EOL is June 30, 2024](https://www.redhat.com/en/topics/linux/centos-linux-eol). Its package feeds are still being updated daily, [see here](https://feeds.centos.org/). However, our base image for that OS, `centos:7` was last updated in 2021 (!).

```
docker inspect centos:7 | grep Created
        "Created": "2021-09-15T18:20:23.99863383Z"
```

By our current auto-re-building standards, this image is "up-to-date". However, there's no doubt the packages we don't install as part of the Dockerfile above are way out of date. We should come up with a system that lets us use the up-to-date packages in our buildtools-prereqs containers so that the images are more secure.

This could possibly be taken care of by https://github.com/dotnet/dotnet-docker/issues/1455, but that is a more general solution.

Off the top of my head, this could be accomplished by:

1. Architecting the (at-risk) buildtools-prereqs Dockerfiles to have a multi-stage build, where all packages on the base system are updated in a separate layer and used as a base layer. The base image layer can be "squashed" so that the image doesn't contain any unnecessary filesystem duplication (not an explicit Docker feature any more, but see https://stackoverflow.com/a/77052183/13771378 for how this can be done now).
2. Make some adjustments to the auto-builder so that it re-builds images that haven't been built in a given amount of time (a week, say).

With the above two features in place, if there aren't frequent updates to a base image, we'd get automatic rebuilds with updated packages.

mthalman:
Fixes #988

* Updates the CentOS 7 Dockerfiles to explicitly upgrade existing packages
* Sets the CentOS 7 pipeline to run on a weekly schedule. I've already configured the pipeline in AzDO to run with caching disabled by default. So that will ensure that all images are rebuilt on a weekly basis.
* Deleted the unused rpmpkg Dockerfile.


MichaelSimons (approved):

src/centos/7/Dockerfile
```diff
@@ -2,7 +2,11 @@ FROM centos:7
 
 # Install dependencies
 
-RUN yum install -y \
+RUN \
+    # Since the base image isn't actively maintained, force an upgrade of existing packages
+    yum update -y \
+    && \
+    yum install -y \
```
lbussell, line 9: This will make the dependency layer really large since the image hasn't been updated in a long time. What is the layer size difference caused by this change, and if it's big (relative to the size of the entire image), will you consider squashing this into a single layer with the base image e.g. https://stackoverflow.com/a/77052183/13771378

PR #1014, opened 2024-04-23
Use Mariner not Ubuntu as armv6 host image

directhex:
This is expected to fail until https://github.com/dotnet/arcade/pull/14730 is merged (it doesn't need to flow to any branches once merged since rootfs generation always uses arcade git main)


mthalman (approved):

mthalman (approved):
@directhex - Are these changes good to merge?

directhex:
I don't see why not. Then we can change the image used in runtime and everything will be just peachy

mthalman:
GitHub is having issues. This is showing and not updating:

image

I'm going to try closing and re-opening.

mthalman:
Ah, now it's fixed. Just needed a comment.

PR #1015, opened 2024-04-23
Add azurelinux 3.0 images with latest LLVM for .NET 9

sbomer:
These images have the same rootfs layout as the .NET 8 images added in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/995, but they have the latest LLVM instead.


richlander:
Thanks!

If you had to do something special to make that happen, it would probably be good to document that in this PR.

jkoritzinsky (approved):

richlander (approved):

richlander:
I see a mix of LLVM versions set. Are we always going to get the one we want?

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1015/files#diff-c80fbd0fa87aeee5ce19605fba6c538b6afc68ed8aba9d01b6c53d4b559dc163R37-R46

Issue #990, opened 2024-03-22
Ensure correct and optimal images are used in pipelines


richlander:
There are a variety of patterns in use that are not correct or could be made more optimal.

## Managed-only builds

We have large images that are intended for native builds. They get used for managed builds, too, which is massive overkill, but we don't offer anything better.

https://github.com/dotnet/arcade/blob/5d8abb43f355edc287d57907967224a0ac5ebb9f/eng/common/templates/jobs/source-build.yml#L17

Referenced image Dockerfile: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/centos/stream8/Dockerfile

## Cross builds

Some cross builds use old patterns.

https://github.com/dotnet/diagnostics/blob/217abce2d2f1a253795231925b049b1171f151ea/eng/pipelines/pipeline-resources.yml#L21

## Varying targets

Some teams need separate targets. Today, the build targets Ubuntu 16.04. The referenced example targets Ubuntu 18.04. Could it just build in plain Mariner instead? Would that provide sufficient compatibility? We should understand the requirements.

https://github.com/dotnet/jitutils/blob/0601313f1d0c30fbfe0143e405d1918c5daed920/coredistools.yml#L42

richlander:
Related: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/869

MichaelSimons:
Related: #805

PR #1020, opened 2024-04-25
Switch from dotnet/msquic to microsoft/msquic

ManickaP:
In order to archive https://github.com/dotnet/msquic we have to get rid of the references. This change replaces the dotnet repo with the official MsQuic one.

### DIFF
--- a/src/alpine/3.16/helix/amd64/Dockerfile
+++ b/src/alpine/3.16/helix/amd64/Dockerfile
@@ -69,8 +69,8 @@ RUN apk update && apk add --no-cache && \
         openssl-dev \
         perl && \
     cd /tmp && \
-    git clone --depth 1 --single-branch --branch main --recursive https://github.com/dotnet/msquic && \
-    cd msquic/src/msquic && \
+    git clone --depth 1 --single-branch --branch main --recursive https://github.com/microsoft/msquic && \
+    cd msquic && \
     cmake -B build/linux/x64_openssl \
        -DQUIC_OUTPUT_DIR=/tmp/msquic/src/msquic/artifacts/bin/linux/x64_Release_openssl \
        -DCMAKE_BUILD_TYPE=Release \
--- a/src/alpine/3.16/helix/amd64/Dockerfile
+++ b/src/alpine/3.16/helix/amd64/Dockerfile
@@ -72,7 +72,7 @@ RUN apk update && apk add --no-cache && \
     git clone --depth 1 --single-branch --branch main --recursive https://github.com/microsoft/msquic && \
     cd msquic && \
     cmake -B build/linux/x64_openssl \
-       -DQUIC_OUTPUT_DIR=/tmp/msquic/src/msquic/artifacts/bin/linux/x64_Release_openssl \
+       -DQUIC_OUTPUT_DIR=/tmp/msquic/artifacts/bin/linux/x64_Release_openssl \
        -DCMAKE_BUILD_TYPE=Release \
        -DQUIC_TLS=openssl \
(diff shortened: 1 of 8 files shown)


CarnaViire:
There seem to be build failures @ManickaP

src/alpine/3.16/helix/amd64/Dockerfile
```diff
@@ -69,10 +69,10 @@ RUN apk update && apk add --no-cache && \
         openssl-dev \
         perl && \
     cd /tmp && \
-    git clone --depth 1 --single-branch --branch main --recursive https://github.com/dotnet/msquic && \
-    cd msquic/src/msquic && \
+    git clone --depth 1 --single-branch --branch main --recursive https://github.com/microsoft/msquic && \
```
wfurt, line 72: One option here is to track MsQuic releases. That would match other platforms where we consume published binaries. 
The `dotnet/msquic` give us some control in the past.  
Pointing at `main` may cause some instabilities and surprises but on the other hand it allows us to see issues sonner.

ManickaP: @CarnaViire suggested the same thing, I'll change it to get this in.

jkoritzinsky:
Related to https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/990, we should move these docker images to use multi-stage images that build msquic in one stage and set up the deployed environment (with the built msquic) in another.

CarnaViire:
@jkoritzinsky The problem is that the build scripts (and therefore the produced binaries) are different, depending on the architecture (x64, arm, arm64) and openssl version (1.1, 3.0). So out of the 8 updated images here, 6 would have _different_ MsQuic, so the environment _cannot_ be reused between them.

Note that Alpine is the only platform where we don't have any official MsQuic packages (tracking issue https://github.com/dotnet/runtime/issues/83789), so we have to resort to the manual build here. All the other images install prebuilt binaries from packages.microsoft.com

jkoritzinsky:
I'm not saying we use a "builder" image or something like that to share between the images. I just mean that we do a second FROM statement and install the tools needed for runtime in the second stage (and copy msquic from the first stage) instead of having the build tools and msquic source on the image that runs in Helix.

ManickaP:
> I'm not saying we use a "builder" image or something like that to share between the images. I just mean that we do a second FROM statement and install the tools needed for runtime in the second stage (and copy msquic from the first stage) instead of having the build tools and msquic source on the image that runs in Helix.

I'm not following this line of thought. And is it necessary for this PR or should it rather be resolved separately? After all, I'm just changing from where we take msquic sources, nothing else.

jkoritzinsky:
It's not required for this PR. I can take a look at doing it after this is merged as an example.

liveans (approved):

rzikm:
Since we opt for consuming the released MsQuic versions, would it make sense to (always) consume the latest published packages where we can? (Ubuntu etc.)

The change does not have to be in this PR, but something to consider for the mentioned future improvements.

rzikm (approved):

CarnaViire:
> Since we opt for consuming the released MsQuic versions, would it make sense to (always) consume the latest published packages where we can? (Ubuntu etc.)

@rzikm We do already:
https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/0b9f68c13626e2e15d916d1565191501b2e0efc5/src/ubuntu/22.04/helix/amd64/Dockerfile#L54-L56
Alpine is the only one where we cannot install the official packages.

PR #1022, opened 2024-04-25
Update pr triggers for eng validation pipeline

lbussell:
This should prevent us from burning 8 hours of compute on PRs like https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1019.

### DIFF
--- a/eng/pipelines/dotnet-buildtools-prereqs-eng.yml
+++ b/eng/pipelines/dotnet-buildtools-prereqs-eng.yml
@@ -4,9 +4,13 @@ pr:
     include:
     - main
   paths:
+    include:
+    - eng/*
+    - manifest.json
+    - build.ps1
+    - .dockerignore
     exclude:
     - src/*
-    - README.md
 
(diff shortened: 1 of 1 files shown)


eng/pipelines/dotnet-buildtools-prereqs-eng.yml
```diff
@@ -4,9 +4,11 @@ pr:
     include:
     - main
   paths:
-    exclude:
-    - src/*
-    - README.md
+    include:
```
MichaelSimons, line 7: exclude was originally used as it was viewed as the safer approach.  Meaning we would rather waste compute than skip running the validation on a new unexpected root file.  I don't feel strongly about this but felt I should point it out.

   exclude:
    - src/*
    - *.md
    - .git*
    - CODEOWNERS
    - LICENSE

mthalman (approved):

MichaelSimons (approved):

PR #1063, opened 2024-05-17
Use XL pools for Mariner/Azure Linux

### LINKED ISSUE #1024: Use XL build pools for CBL Mariner and Azure Linux builds
CBL Mariner and Azure Linux builds can take upwards of 6 hours using the current build agents. We should move those specific pipelines to the XL build pools so they run a little faster:

Internal:
- `NetCore1ESPool-Internal-XL`

Public:
- `NetCore-Public-XL`

/cc @sbomer

mthalman:
Note that the `-all` pipeline has to be configured to run with these as well because it may have to build Mariner/Azure Linux images.

Fixes #1024


MichaelSimons (approved):
Do we know the time impact of this change on the Mariner/Azure Linux legs?  If not, it would be good to report back once merged and we have some runs.

mthalman:
As follow up on this, the job that was timing out after 8 hrs now finished in 6 hrs.

PR #1064, opened 2024-05-17
Build msquic in a separate docker stage

jkoritzinsky:
This allows us to avoid leaking anything from the msquic build into the alpine image, promotes better caching while iterating on docker images, and provides better clarity.

It also means that we don't need to carefully remove files/packages that are only needed for the msquic build.

### DIFF
--- a/src/alpine/3.16/helix/amd64/Dockerfile
+++ b/src/alpine/3.16/helix/amd64/Dockerfile
@@ -19,7 +19,7 @@ RUN git clone --depth 1 --single-branch --branch v2.3.5 --recursive https://gith
 
 WORKDIR /tmp/msquic
 
-RUN cmake -B build/linux/x64_openssl \
+RUN cmake -B build/linux/x64_Release_openssl \
        -DQUIC_OUTPUT_DIR=/tmp/msquic/artifacts/bin/linux/x64_Release_openssl \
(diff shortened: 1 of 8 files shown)


rzikm (approved):
LGTM, thanks!

wfurt:
The change seems to miss 
```
cp artifacts/bin/linux/x64_Release_openssl3/libmsquic.so.* artifacts/bin/linux/x64_Release_openssl3/libmsquic.lttng.so.* /usr/lib && \
     rm -rf /tmp/msquic && \
```
Not sure if that was intentional but unless we install the libraries to standard location or unless we modify library search packages and tests will not be able to find it. 
And I see no point of leaving the build tree behind, we really just need the `.so` files.

jkoritzinsky:
Yes, this change intentionally uses msquic's install layout directly instead of manually doing the install ourselves. The `cmake --install --prefix /msquic` and `COPY --from=msquic /msquic /` install the msquic binaries into the standard layout under `/msquic` and then copy them between the layers into the standard Linux layout, laying them out exactly how the msquic build intends.

jkoritzinsky:
By using separate stages, we do leave the build tree behind. It doesn't get copied into the second stage.

The whole point of using two stages is that we get isolation between them, so we don't need to manually delete files or uninstall packages to get back to the "before we tried to build msquic" state.

PR #1065, opened 2024-05-17
Remove hooks from all images

jkoritzinsky:
Switch all of our images onto the multi-stage build plan and away from using the custom ImageBuilder hooks feature.

This makes our images more consistently designed and makes maintaining them follow a more standard development flow.


src/ubuntu/22.04/cross/arm-alpine/Dockerfile
```diff
@@ -1,4 +1,15 @@
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-22.04-crossdeps-local AS builder
+ARG ROOTFS_DIR=/crossrootfs/arm
```
MichaelSimons, line 2: I think you can define the arg outside a stage scope to have a single shared default value.  You sill have to declare the ARG in each stage it is used.  This will avoid having "/crossrootfs/arm" defined in multiple locations.

Example

ARG ROOTFS_DIR=/crossrootfs/arm

FROM stageA
ARG ROOTFS_DIR

FROM stageB
ARG ROOTFS_DIR

src/raspbian/10/helix/arm32v6/Dockerfile
```diff
@@ -1,13 +1,21 @@
-# dummy F R O M to establish a dependency
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:raspbian-10-crossdeps-local
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:raspbian-10-crossdeps-local as builder
+
+ARG ROOTFS_DIR=/crossrootfs
+
+# Obtain arcade scripts used to build rootfs
+RUN git config --global user.email builder@dotnet-buildtools-prereqs-docker && \
+    git clone --depth 1 --single-branch https://github.com/dotnet/arcade /scripts
```
MichaelSimons, line 7: It doesn't feel right that we are cloning arcade tip.  We really should be using the version of arcade this repo has.  Tip arcade hasn't been validated through it's inner validation process.

mthalman: > We really should be using the version of arcade this repo has.

There is no Arcade dependency flow to this repo.

mthalman: It's also no different than what was previously being done in run-arcade-build-rootfs.sh. So that should be addressed as separate work if it is deemed an issue.

MichaelSimons: I see we are currently doing this in the run-arcade-build-rootfs.sh script.  I thought we utilizing dependency flow for isolation.  I guess that is a separate issue - it obviously has not been a source of problems.  You can ignore my comment.

src/ubuntu/22.04/cross/arm-alpine/Dockerfile
jkoritzinsky: As we follow this pattern in a number of images, I'd like to do the refactoring in a follow-up PR (so as to hit all of the cases, including the cbl-mariner and azurelinux images). Is that okay with you?

MichaelSimons: Yes that is fine.

src/raspbian/10/helix/arm32v6/Dockerfile
```diff
@@ -1,13 +1,21 @@
...
+RUN /scripts/eng/common/cross/build-rootfs.sh armv6 raspbian --skipunmount
 
 FROM scratch
 
 # There doesn't seem to be an official Raspberry Pi OS Docker image
 # But we can use a crossrootfs from our image generation scripts, and
 # extract into the image root
 
-ADD rootfs.armv6.tar /
+COPY --from=builder /crossrootfs /
```
MichaelSimons, line 18: This should make use of an ARG.

MichaelSimons (approved):

MichaelSimons (approved):

PR #1070, opened 2024-05-21
Try 2: Split MSQuic building into a separate stage

jkoritzinsky:
First commit is #1064.

The second commit adds a missing dependency of libmsquic into the image, numactl. This was the only package that was installed as part of the msquic build and then not uninstalled. It's required by msquic to function.


wfurt:
This still does not put the libraries into `/usr/lib`, right? So 1) it is different than all other platforms and 2) I don't think tests will find it. 
If you feel otherwise did you run `Quic` tests on produced image?

jkoritzinsky:
It puts them into `/lib`. I'll update this PR to put them into `usr/lib`.

jkoritzinsky:
@wfurt any more feedback?

ManickaP:
I tested 3.18 amd64 from this PR, `QuicConnection.IsSupported` returned `true`, so it works.

ManickaP (approved):

PR #1079, opened 2024-05-30
Add support for Alpine 3.20

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/4399

### DIFF
--- a/src/alpine/3.20/amd64/Dockerfile
+++ b/src/alpine/3.20/amd64/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.20
+FROM amd64/alpine:3.20
 
 # Install .NET and test dependencies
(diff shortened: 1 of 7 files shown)


src/alpine/3.20/amd64/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM alpine:3.20
```
mthalman, line 1: This should either have `amd64/alpine:3.20` or move it out of the `amd64` directory if we want it to be multi-arch-capable.

MichaelSimons: I updated all of the existing amd64 alpine images given they are currently being built just for amd64.

mthalman (approved):

PR #1080, opened 2024-05-30
Refactor ARG pattern used in cross images

### LINKED ISSUE #1066: Define ROOTFS_ARGs outside of stages
I think you can define the arg outside a stage scope to have a single shared default value.  You sill have to declare the ARG in each stage it is used.  This will avoid having "/crossrootfs/arm" defined in multiple locations.

Example

ARG ROOTFS_DIR=/crossrootfs/arm

FROM stageA
ARG ROOTFS_DIR

FROM stageB
ARG ROOTFS_DIR

_Originally posted by @MichaelSimons in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1065#discussion_r1606782931_

MichaelSimons:
Fixes #1066


jkoritzinsky (approved):

src/ubuntu/20.04/crossdeps/Dockerfile
```diff
@@ -20,17 +20,17 @@ RUN apt-get update \
         debootstrap \
         flex \
         gdb \
-        make \
```
MichaelSimons, line 23: I needed to touch this Dockerfile to force it to rebuild (not pull the cached version) because the key installed at https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1080/commits/19046ff5129a88d598a5e909400a50bb32a6a759#diff-c7d578fdb4b1e1f116216124a43e0ddafdeeabd43843fd305f38f8f1e1a4d186R11 is out of date so when derived Dockerfiles (e.g. ubuntu-cross-arm64) are rebuilt and do an apt-get update, they get the following error:

a GPG error: https://apt.kitware.com/ubuntu focal InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 1A127079A92F09ED"

am11: Touching this fine was a right thing to do. I was getting the same error during testing another PR (https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1102, I just installed `nano` there to keep the ball rolling).

am11:
@MichaelSimons, please merge main one more time (since #1104 is merged). Fingers crossed. 🤞

PR #1083, opened 2024-05-30
Add helix image for armv6 raspbian

sbomer:
The raspbian 10 helix images were removed in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/927, so removing both raspbian/10 Dockerfiles and replacing them with raspbian 12 that has recent enough glibc.


jkoritzinsky (approved):

eng/common/templates/stages/dotnet/build-test-publish-repo.yml
mthalman, line 1: These changes in this file aren't needed. The pipelines in question are using eng/pipelines/stages/build-test-publish-repo.yml (note the lack of `dotnet` directory). This file is also in the eng/common which should have changes applied in the dotnet/docker-tools repo instead and have them flowed to this repo. But regardless, these changes aren't needed.

mthalman: Yep, you're right.

src/raspbian/manifest.json
```diff
@@ -7,19 +7,34 @@
           "platforms": [
             {
               "architecture": "arm",
-              "dockerfile": "src/raspbian/10/crossdeps/arm32v7",
+              "dockerfile": "src/raspbian/12/crossdeps/arm32v7",
               "os": "linux",
-              "osVersion": "buster",
+              "osVersion": "jammy",
```
akoeplinger, line 12: bookworm?

src/raspbian/12/helix/arm32v6/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:raspbian-10-crossdeps-local as builder
+FROM --platform=linux/arm/v7 mcr.microsoft.com/dotnet-buildtools/prereqs:raspbian-12-crossdeps-local as builder
```
akoeplinger, line 1: why are we passing --platform here?

am11: Should it be [`linux/arm/v6`](https://docs.docker.com/build/building/multi-platform/)?

mthalman: It shouldn't be necessary at all. That would only be necessary if you were pulling a multi-arch tag which isn't the case here.

am11: If we are going to replace "armv6" image and new image doesn't provide armv6, then it is not the right image? I think `uname -a` etc. should show `armv6l`. v7 and v6 are very different ISAs; which is why we have ARM and ARMV6 as [separate architectures](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.processorarchitecture?view=msbuild-17-netcore#fields).

mthalman: > Without `--platform` I am getting the following in ci:
> 
> ```
> Dockerfile:1
> --------------------
>    1 | >>> FROM mcr.microsoft.com/dotnet-buildtools/prereqs:raspbian-12-crossdeps-local as builder
>    2 |     
>    3 |     ARG ROOTFS_DIR=/crossrootfs
> --------------------
> ERROR: failed to solve: mcr.microsoft.com/dotnet-buildtools/prereqs:raspbian-12-crossdeps-local: mcr.microsoft.com/dotnet-buildtools/prereqs:raspbian-12-crossdeps-local: not found
>    at Microsoft.DotNet.ImageBuilder.ExecuteHelper.Execute(ProcessStartInfo info, Func`2 executor, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 101
>    at Microsoft.DotNet.ImageBuilder.ExecuteHelper.ExecuteWithRetry(ProcessStartInfo info, Action`1 processStartedCallback, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 60
>    at Microsoft.DotNet.ImageBuilder.ExecuteHelper.ExecuteWithRetry(String fileName, String args, Boolean isDryRun, String errorMessage, String executeMessageOverride) in /image-builder/src/ExecuteHelper.cs:line 45
>    at Microsoft.DotNet.ImageBuilder.DockerService.BuildImage(String dockerfilePath, String buildContextPath, String platform, IEnumerable`1 tags, IDictionary`2 buildArgs, Boolean isRetryEnabled, Boolean isDryRun) in /image-builder/src/DockerService.cs:line 60
>    at Microsoft.DotNet.ImageBuilder.DockerServiceCache.BuildImage(String dockerfilePath, String buildContextPath, String platform, IEnumerable`1 tags, IDictionary`2 buildArgs, Boolean isRetryEnabled, Boolean isDryRun) in /image-builder/src/DockerServiceCache.cs:line 37
>    at Microsoft.DotNet.ImageBuilder.Commands.BuildCommand.BuildImage(PlatformInfo platform, IEnumerable`1 allTags) in /image-builder/src/Commands/BuildCommand.cs:line 483
> ```

This doesn't make sense to me. This happens on the CI in AzDO? I can't find a failed build with that error. I'd like to see the build log to understand what's happening. Getting a `not found` error would mean the image tag doesn't exist at all, regardless of its platform.

mthalman: Ok, I get why `--platform` is needed now. That's fine. I'll leave it to the other reviewers to determine whether consuming a v7 image from a v6 image is the right thing to do.

MichaelSimons:
This PR will need to be updated to respond to the following infra changes that were recently made:

- https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1174
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1169

MichaelSimons:
@sbomer - Friendly ping.  Is this something you intend to complete?  This PR is pretty old now.

mthalman:
> @sbomer - Friendly ping. Is this something you intend to complete? This PR is pretty old now.

@sbomer - Any update here? Note that I've deleted all of the Raspbian 10 Dockerfiles because they were based on Ubuntu 18.04 which is EOL.

src/raspbian/12/helix/arm32v6/Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN git config --global user.email builder@dotnet-buildtools-prereqs-docker && \
     git clone --depth 1 --single-branch https://github.com/dotnet/arcade /scripts
 
 # Build the rootfs
-RUN /scripts/eng/common/cross/build-rootfs.sh armv6 raspbian --skipunmount
+RUN /scripts/eng/common/cross/build-rootfs.sh armv6 bookworm lldb15 --skipunmount
```
akoeplinger, line 12: @am11 we can probably remove the code in https://github.com/dotnet/arcade/blob/d1bc0779143e410ec82696f9865c4eb407db8d67/eng/common/cross/build-rootfs.sh#L170-L183 which sets the __UbuntuRepo, __CodeName etc since that is handled by the `bookworm` argument

src/raspbian/12/helix/arm32v6/Dockerfile
```diff
@@ -45,8 +45,10 @@ RUN apt-get update && \
         libunwind8 \
         locales \
         locales-all \
+        pkg-config \
         python3-dev \
         python3-pip \
+        rustc \
```
akoeplinger, line 51: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1145/files installs rust from rustup instead because apparently the rustc in bookworm is too old?

src/raspbian/12/helix/arm32v6/Dockerfile
```diff
@@ -60,7 +62,7 @@ ENV LANG=en_US.utf8
 RUN ln -sf /usr/bin/python3 /usr/bin/python && \
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
     export CRYPTOGRAPHY_DONT_BUILD_RUST=1 && \
-    pip install ./helix_scripts-*-py3-none-any.whl
+    pip install ./helix_scripts-*-py3-none-any.whl --break-system-packages
```
akoeplinger, line 65: please add a comment why we need --break-system-packages

jkoritzinsky: Alternatively, could we use the multistage docker + python venv approach we've done in all the other images and avoid using `--break-system-packages` (#1142)?

richlander:
We can likely close this PR per https://github.com/dotnet/runtime/pull/115129.

PR #1089, opened 2024-06-12
Do not install emscripten anymore

radekdoulik:
We provision it from nuget packages


mthalman (approved):
Nice. I'm a fan of deleting stuff.

PR #1091, opened 2024-06-13
Add Alpine 3.20 Helix images

### LINKED ISSUE #1093: Alpine 3.20 no longer supports same Python pattern as Alpine 3.18
I took our Alpine 3.18 Dockerfiles, changed the Alpine version number to `3.20` and built the Dockerfile. It fails.

Related: https://github.com/pypa/pip/issues/12429

It appears that Alpine is now a "Python built-in" distro like we've seen of Debian and special handling is now needed. Also, the Helix use of Python appears to be very dated. The Dockerfile appears to use virtual environments but doesn't actually use them.

I'm trying to develop a better pattern in #1091.

Observed error.

```bash
5.656 (18/18) Installing python3-dev (3.12.3-r1)
6.349 Executing busybox-1.36.1-r28.trigger
6.370 OK: 1285 MiB in 96 packages
8.680 error: externally-managed-environment
8.680 
8.680 × This environment is externally managed
8.680 ╰─> 
8.680     The system-wide python installation should be maintained using the system
8.680     package manager (apk) only.
8.680     
8.680     If the package in question is not packaged already (and hence installable via
8.680     "apk add py3-somepackage"), please consider installing it inside a virtual
8.680     environment, e.g.:
8.680     
8.680     python3 -m venv /path/to/venv
8.680     . /path/to/venv/bin/activate
8.680     pip install mypackage
8.680     
8.680     To exit the virtual environment, run:
8.680     
8.680     deactivate
8.680     
8.680     The virtual environment is not deleted, and can be re-entered by re-sourcing
8.680     the activate file.
8.680     
8.680     To automatically manage virtual environments, consider using pipx (from the
8.680     pipx package).
8.680 
8.680 note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
8.680 hint: See PEP 668 for the detailed specification.
------
Dockerfile:55
--------------------
  54 |     # Install Helix Dependencies
  55 | >>> RUN apk update && apk add --no-cache && \
  56 | >>>     apk add \
  57 | >>>         cargo \
  58 | >>>         libffi-dev \
  59 | >>>         linux-headers \
  60 | >>>         python3-dev \
  61 | >>>         openssl-dev && \
  62 | >>>     ln -sf /usr/bin/python3 /usr/bin/python && \
  63 | >>>     curl https://bootstrap.pypa.io/get-pip.py -o ./get-pip.py --fail --silent --show-error && \
  64 | >>>     python3 ./get-pip.py && rm ./get-pip.py && \
  65 | >>>     python3 -m pip install --upgrade pip==22.2.2 && \
  66 | >>>     python3 -m pip install virtualenv==20.16.5 && \
  67 | >>>     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
  68 | >>>     pip install ./helix_scripts-*-py3-none-any.whl && \
  69 | >>>     apk del \
  70 | >>>         cargo \
  71 | >>>         libffi-dev \
  72 | >>>         linux-headers \
  73 | >>>         python3-dev \
  74 | >>>         openssl-dev
  75 |     

```

richlander:
Fixes #1093.

I tried to use `py3-psutil` from Alpine but was unable to do so due to the way dependencies are specified in the helix wheel we install. Instead, the `psutil` is built from source (with `gcc`). Explained in more detail at #1092.

I switched to `venv` from `virtualenv` since the former comes with Python3 distributions. AFAICT, `virtualenv` is more a legacy Python 2-ism.

The way we install `pip` and `virtualenv` is anti-pattern as I've written before. In some Dockerfiles I've seen, we install `pip` three times in a row. "Gotta be sure!"

I've tried to address these problems, however, there is a quite possible chance that we're either need to change the way these images are launched or go back to `--break-system-packages`, like we do in Debian. I really want to get away from that, since it is a clear demonstration that we're unable to use Python as it is designed.

Due to what is discussed in #1093, I actually used the virtual environment for what it is intended for, which is

### DIFF
--- a/src/alpine/3.20/helix/amd64/Dockerfile
+++ b/src/alpine/3.20/helix/amd64/Dockerfile
@@ -48,30 +48,10 @@ RUN apk update && apk add --no-cache \
     numactl \
     openssl \
     python3 \
+    python3-dev \
+    py3-pip \
     sudo \
     tzdata
-
-# Install Helix Dependencies
-RUN apk update && apk add --no-cache && \
-    apk add \
-        cargo \
-        libffi-dev \
-        linux-headers \
-        python3-dev \
-        openssl-dev && \
-    ln -sf /usr/bin/python3 /usr/bin/python && \
-    curl https://bootstrap.pypa.io/get-pip.py -o ./get-pip.py --fail --silent --show-error && \
-    python3 ./get-pip.py && rm ./get-pip.py && \
-    python3 -m pip install --upgrade pip==22.2.2 && \
-    python3 -m pip install virtualenv==20.16.5 && \
-    pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && \
-    pip install ./helix_scripts-*-py3-none-any.whl && \
-    apk del \
-        cargo \
-        libffi-dev \
-        linux-headers \
-        python3-dev \
-        openssl-dev
 
 # Copy msquic from the msquic image into our image that will run on Helix
 COPY --from=msquic /msquic /usr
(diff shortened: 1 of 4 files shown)


mthalman (approved):

src/alpine/3.20/helix/amd64/Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+    llvm \
+    lttng-ust \
+    musl-locales \
+    numactl \
+    openssl \
+    python3 \
+    python3-dev \
+    py3-pip \
+    sudo \
+    tzdata
```
mthalman, line 54: These should be intended to illustrate they're continuations of the previous line's command. Please apply that formatting to other usages in this PR.

mthalman (commented):
And of course, this doesn't seem to work on Arm32. Getting an error when building the cryptography wheel.

richlander:
Yup. I'm still working the Arm32 Dockerfile. Fun!

PR #1104, opened 2024-06-20
Pin debootstrap to tag 1.0.134 and fix azurelinux 3.0 build

am11:
While https://github.com/microsoft/azurelinux/pull/9422 is under review, we can pin it to ubuntu1 release tag.

We also need to add some more dependencies since azurelinux 3.0 rebuilt their packages and now doesn't install some transitive dependencies anymore.

PR #1107, opened 2024-06-21
Add make and gcc to azurelinux 3.0 crossdeps images

akoeplinger:
See https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1104#issuecomment-2183543291


am11:
Was it expected? It seems like they had a major downtime yesterday but other than that, we don’t know if they have recovered from it or still midway through?

akoeplinger:
I'm not sure, I got the information that the 3.0 repository had to be republished and that it was completed. I sent a follow-up email to ask whether the dependency changes are expected.

akoeplinger:
Closing in favor of #1104

PR #1121, opened 2024-07-01
Define common Ubuntu coredeps Dockerfile

mthalman:
The Ubuntu coredeps Dockerfiles reference out-of-date resources:

* `src/ubuntu/20.04/coredeps/Dockerfile`:
  * [Unsupported base image tag](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/c2cea7ad2eb85c1881602df04c0f5f5c3c4d46ef/src/ubuntu/20.04/coredeps/Dockerfile#L1)
  * [Unsupported AZ CLI version](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/c2cea7ad2eb85c1881602df04c0f5f5c3c4d46ef/src/ubuntu/20.04/coredeps/Dockerfile#L12)
* `src/ubuntu/22.04/coredeps/Dockerfile`:
  * [Unsupported base image tag](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/c2cea7ad2eb85c1881602df04c0f5f5c3c4d46ef/src/ubuntu/22.04/coredeps/Dockerfile#L1)

I've fixed this by defining a common Dockerfile that is shared for both Ubuntu versions. Each now references a supported PowerShell base image tag. And the AZ CLI install script is used to ensure the latest version gets installed.


lbussell (approved):

src/ubuntu/manifest.json
```diff
@@ -15,6 +15,9 @@
                 "ubuntu-20.04-coredeps-local": {
                   "isLocal": true
                 }
+              },
+              "buildArgs": {
+                "POWERSHELL_TAG": "lts-7.2-ubuntu-20.04"
```
MichaelSimons, line 20: Would it make sense to define a versions manifest to start putting these types of valuesin?  The benefit is to ease maintenance by getting them next to each other.  Possibly the value could be decomposed so that the pwsh version is a shared value.

src/ubuntu/common/coredeps/Dockerfile
MichaelSimons, line 1: Sharing coredeps across OS versions happens to work here because none of the package names changed.  What is the recommended pattern for when they would change?

I noticed the AzDO build automation and Runtime dependency sections are different in the 24.04 version.  Do you know why?  How do we get to a share set of tools/dependencies that could be defined in something like a shared script/Dockerfile?

mthalman: > Sharing coredeps across OS versions happens to work here because none of the package names changed. What is the recommended pattern for when they would change?

We'll cross that bridge when we get there. I'm just trying to take advantage of content sharing as it exists right now.

> I noticed the AzDO build automation and Runtime dependency sections are different in the 24.04 version. Do you know why?

What difference? The only difference I see is the way that AZ CLI was installed. The script install seems preferred.

> How do we get to a share set of tools/dependencies that could be defined in something like a shared script/Dockerfile?

I see shared scripts as the ideal implementation since they allow for finer-grained scope than an entire Dockerfile. But that's dependent on https://github.com/dotnet/docker-tools/issues/1270.

mthalman: That's not a coredeps Dockerfile though.

MichaelSimons: Gosh it would be nice to get to a place where these sections were shared and made it apparent what the scenarios are where the installed components differ.  Seeing the same section headers across similar dockerfiles yet different components are installed is confusing.

src/ubuntu/manifest.json
mthalman: Actually, I'm kind of inclined to just change it to a non-version floating tag: `lts-ubuntu-20.04`. That's essentially the same as installing it via package manager as far what version you end up getting: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/cbf67da80a106ac8c3114a75d07bf516eebae9a5/src/ubuntu/22.04/debpkg/amd64/Dockerfile#L23

mthalman: I've updated the tag usage to get the latest PowerShell version available. This aligns with the version that gets installed by other Dockerfiles that are installing from PMC. This will give you PowerShell 7.4 in both cases.

There is some cleanup of other PowerShell usage amongst the Dockerfiles but that's outside the scope of coredeps so I'll address those separately.

MichaelSimons (approved):

PR #1095, opened 2024-06-15
Switch from virtualenv to venv for all helix Dockerfiles

richlander:

PR #1122, opened 2024-07-01
Fix remaining Helix Dockerfiles to switch to venv

mthalman:
This applies the changes from https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1095 that were missed for other Helix Dockerfiles.


richlander:
I tried CentOS 7 initially and ran into troubles so quickly gave up. I should have mentioned that at the time. Sorry about that. We'll delete that image at EOY so I decided that the investment wasn't worth it. Also, now that the feed is broken (and would require moving to vault), it feels even less worth it.

mthalman:
ping @dotnet/dotnet-docker-reviewers

MichaelSimons (approved):

PR #1123, opened 2024-07-01
Delete CentOS 7 Dockerfiles

mthalman:
Deleting these Dockerfiles due to CentOS 7 EOL.

The existing Dockerfiles are non-operational as they are due to the mirrorlist host being taken offline:
```
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
0.900 14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error"
```


MichaelSimons (approved):

lbussell (approved):

richlander:
For anyone reading this, https://vault.centos.org/ is apparently an alternative to the mirror that has been disabled. We're not planning on using it, but FYI.

mmitche (approved):

PR #1125, opened 2024-07-03
Remove `zlib-devel` dependency from azurelinux for .NET 9.0

carlossanlop:
Contributes to https://github.com/dotnet/runtime/issues/101465

For .NET 9.0, we brought a copy of zlib-ng into dotnet/runtime that will be used by all platforms, except for the mobile ones (android, ios, tvos, maccatalyst) and armv6: https://github.com/dotnet/runtime/pull/102403

For that reason, and as suggested in [this comment](https://github.com/dotnet/runtime/pull/104399#issuecomment-2207009080), the azurelinux docker images that target .NET 9.0 won't need to install `zlib-devel` as a dependency anymore.

### DIFF
--- a/src/azurelinux/3.0/net9.0/crossdeps-builder/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net9.0/crossdeps-builder/amd64/Dockerfile
@@ -16,7 +16,6 @@ RUN tdnf install -y \
         glibc-devel \
         kernel-headers \
         texinfo \
-        zlib-devel \
         # Rootfs build dependencies
         bzip2-devel \
         debootstrap \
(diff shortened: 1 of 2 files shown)


src/azurelinux/3.0/net9.0/crossdeps-builder/amd64/Dockerfile
```diff
@@ -16,7 +16,6 @@ RUN tdnf install -y \
         glibc-devel \
         kernel-headers \
         texinfo \
```
jkotas, line 18:
```suggestion
        texinfo \
        zlib-devel \
```
The build errors suggest that the builder crossdeps image needs it for other reasons.

jkotas (approved):
We should wait a few weeks for the dust to settle before merging.

carlossanlop:
> We should wait a few weeks for the dust to settle before merging.

@jkotas How about I merge this on Monday August 12th? That's one day before we release P7. On the 13th we will start consuming P7 officially in our own repos.

jkotas:
> That's one day before we release P7. 

This is nice-to-have cleanup. I think that doing this one day *after* we release P7 would make more sense, so that it cannot interfere with P7 release in any way by construction.

carlossanlop:
P7 has already been release. We even snapped 9.0 already. So this should be safe to merge now.

PR #1126, opened 2024-07-08
Add Alpine 3.17 arm64v8 image

mattleibow:


mthalman (commented):
What's the need for this image and why such an old version of Alpine?

mattleibow:
This will be used when I build Alpine things on my Apple Silicon Mac. Right now, the 3.17 is old because I have some older things that I need to first fix and then I will update to a later version. One reason is that I am building and testing on 3.17 since that version of Alpine is still supported: https://learn.microsoft.com/en-us/dotnet/core/install/linux-alpine#supported-distributions

jkotas:
Is this image going to be used in the official builds or CI?

richlander:
Also, for which release is this targeting?

richlander:
The following images are provided for that purpose and will provide better compatibility than Alpine 3.17.

https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/linux-instructions.md#docker-images

richlander:
Ah. Those images are intended to be run on x64. We don't have a good solution for Arm64. Can you use AzDo for the actual building?

mattleibow:
Let me test this rootfs stuff, but still that only helps with CI builds and I am using Apple Silicon half the time so we need to maybe see what I can do. I am a total cross-compile on linux noob, so I am learning as I go :)

richlander:
Why do you need to use Mac hardware if you are building Alpine? Won't an Ubuntu VM work fine?

richlander:
Are you going to continue with this @mattleibow?

MichaelSimons:
This PR will need to be updated to respond to the following infra changes that were recently made:

- https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1174
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1169

MichaelSimons:
Closing due to inactivity.  Feel free to re-open this or a new PR if this is still needed.

PR #1133, opened 2024-07-11
Add MsQuic to Ubuntu 20.04 Dockerfile

rzikm:
Supports https://github.com/dotnet/runtime/pull/104689.


am11:
@rzikm just curious, is it for 'testing on oldest supported Ubuntu' or 'any Ubuntu'? If it is latter, we can probably use 24.04 which will buy us couple of years until EOL.

wfurt:
that may be historical reasons @am11. Msquic several times broke older versions as they focus only on tip of Ubuntu. (and run tests only there) 
Ideally we would run tests on verity of Linux versions IMHO.

PR #1137, opened 2024-07-11
Add azurelinux optimization dockerfiles

LoopedBard3:
Add azurelinux optimization dockerfiles for amd64 and arm64, along with adding them to the manifest.json file, to provide updated LLVM to dotnet-optimization repo. Used the https://github.com/dotnet/dotnet-buildtools-prereqs-docker/tree/main/src/cbl-mariner/2.0/opt files as the base and just updated the base image. Was able to test the build of the Amd64 image and it builds successfully. These will be the new images used for the dotnet optimization runs.


LoopedBard3:
Let me know if there is anything missing or that I should double check as this is my first time add/updating files here.

src/azurelinux/3.0/net9.0/opt/arm64/Dockerfile
mthalman, line 1: This is identical to the `amd64` Dockerfile so let's define just a single file that is used for both architectures. Place the common Dockerfile at `src/azurelinux/3.0/net9.0/opt/Dockerfile` and update the paths in the manifest appropriately.

I understand the related Dockerfiles from Mariner 2.0 don't do this but we're trying to get this repo in better shape for maintainability going forward.

LoopedBard3: Makes sense to me. I updated the PR to reflect this. Let me know if there is anything else. Thank you!

mthalman (approved):

LoopedBard3:
Once this is merged, is there a general amount of time I should suspect for the new images to be usable? Thanks!

mthalman:
> Once this is merged, is there a general amount of time I should expect for the new images to be usable? Thanks!

It's very dependent on the Dockerfile. It could be anywhere from 15 mins to 8 hrs. This one is available now.

PR #1138, opened 2024-07-12
Refactor Apline Dockerfiles to use a consist and efficient apk pattern

MichaelSimons:
The motivation for this PR was the following bad `apk` usage pattern:

```
RUN apk update && apk add --no-cache && \
    apk add \
```

This was causing the package cache to be populated and not cleaned up which introduces image bloat.  Additionally with `apk`, there is no need to make an explicit `apk update` call as `apk add` command will automatically update the package index.  As a result, all Alpine Dockerfiles were updated to use the following simple pattern:

```
RUN apk add --no-cache \
```

### DIFF
--- a/src/alpine/3.17/helix/amd64/Dockerfile
+++ b/src/alpine/3.17/helix/amd64/Dockerfile
@@ -1,7 +1,7 @@
 FROM amd64/alpine:3.17 as msquic
 
 # build MsQuic as we don't have packages
-RUN apk add --no-cache \
+RUN apk add --upgrade --no-cache \
         cmake \
         g++ \
         gcc \
@@ -34,7 +34,7 @@ RUN cmake -B build/linux/x64_Release_openssl3 \
 FROM amd64/alpine:3.17
 
 # Install .NET and test dependencies
-RUN apk add --no-cache \
+RUN apk add --upgrade --no-cache \
         bash \
         coreutils \
         curl \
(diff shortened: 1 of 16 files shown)


src/alpine/3.17/helix/amd64/Dockerfile
```diff
@@ -1,8 +1,7 @@
 FROM amd64/alpine:3.17 as msquic
 
 # build MsQuic as we don't have packages
-RUN apk update && apk add --no-cache && \
-    apk add \
+RUN apk add --no-cache \
```
lbussell, line 4: Is there a reason we aren't using the `--upgrade` option like the dotnet-docker Dockerfiles?

https://github.com/dotnet/dotnet-docker/blob/c5d65d498e7ddacdd6a11410a7a2c5307a3e0bd1/src/runtime-deps/6.0/alpine3.19/amd64/Dockerfile#L11

MichaelSimons: We haven't been using that pattern in this repo.  Is it necessary here?  I don't think the same reasons we do this in dotnet-docker applying here.  I see this change as being outside the scope of these changes.

lbussell: It would match the behavior of other package managers and potentially cut down on CVEs by installing the latest versions of packages if they're already installed from the base distro or another layer - https://github.com/dotnet/dotnet-docker/issues/4776.

If we want to address this in a separate issue or PR that's fine though.

MichaelSimons: Alright you convinced me.  I included the option in this PR.

lbussell (approved):

PR #1139, opened 2024-07-18
Update msquic to v2.3.6

liveans:
Packages has not published yet, but it's fine to have this PR for now.


wfurt (approved):
LGTM. Lack of 24.04 package is unfortunate but we can fix it later.

PR #1142, opened 2024-07-22
Install Helix Python packages into a venv in a separate docker stage

jkoritzinsky:
Install Helix python packages into a separate docker stage into a venv to avoid installing and uninstalling packages from the final image and to avoid using `--break-system-packages`.

### DIFF
--- a/src/alpine/3.17/helix/amd64/Dockerfile
+++ b/src/alpine/3.17/helix/amd64/Dockerfile
@@ -1,4 +1,4 @@
-FROM amd64/alpine:3.17 as msquic
+FROM amd64/alpine:3.17 AS msquic
 
 # build MsQuic as we don't have packages
 RUN apk add --upgrade --no-cache \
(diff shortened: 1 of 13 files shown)


jkoritzinsky:
I've validated that installing the packages into an activated venv does not break the dotnet/runtime CI in https://github.com/dotnet/runtime/pull/105157.

richlander:
Thanks! So much better!

src/alpine/3.17/helix/amd64/Dockerfile
```diff
@@ -85,4 +81,9 @@ RUN /usr/sbin/adduser -D -g '' -G adm -s /bin/bash -u 1000 helixbot && \
 
 USER helixbot
 
-RUN python3 -m venv /home/helixbot/.vsts-env
+# Install Helix Dependencies
+
+ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+COPY --from=venv /venv /home/helixbot/.vsts-env
```
mthalman, line 89:
```suggestion
COPY --from=venv /venv $VIRTUAL_ENV
```

mthalman (approved):

mthalman:
Looks like there are issues with arm32.

richlander:
> Looks like there are issues with arm32.

I thought that was the rule!

PR #1147, opened 2024-07-24
Define Alpine helix images using multi-platform dockerfiles

jkoritzinsky:
Follows the pattern used by the ubuntu 24.04 helix dockerfile


src/alpine/3.17/helix/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM amd64/alpine:3.17 AS msquic
+FROM alpine:3.17 AS msquic
```
mthalman, line 1: All of these Dockerfiles will need to provide the explicit repo name. We have an [infrastructure issue](https://github.com/dotnet/docker-tools/issues/1341) that requires this.
```suggestion
FROM library/alpine:3.17 AS msquic
```

richlander:
Looks good. Great direction.

MichaelSimons (approved):

PR #1156, opened 2024-08-05
Update msquic version and configure distros to use latest package again

### LINKED ISSUE #1155: MsQuic 2.4
Revert https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1154 when working MsQuic 2.4 is published on packages.microsoft.com

liveans:
Fixes #1155
Contributes to dotnet/runtime#82077

This PR also converts Alpine msquic branch to v2.4.1 from main

/cc @ManickaP

### DIFF
--- a/src/alpine/3.17/helix/Dockerfile
+++ b/src/alpine/3.17/helix/Dockerfile
@@ -15,7 +15,7 @@ RUN apk add --upgrade --no-cache \
         openssl-dev \
         perl
 
-RUN git clone --depth 1 --single-branch --branch main --recursive https://github.com/microsoft/msquic /tmp/msquic
+RUN git clone --depth 1 --single-branch --branch v2.4.1 --recursive https://github.com/microsoft/msquic /tmp/msquic
 
 WORKDIR /tmp/msquic
 
--- a/src/almalinux/8/helix/amd64/Dockerfile
+++ b/src/almalinux/8/helix/amd64/Dockerfile
@@ -13,8 +13,7 @@ RUN rpm -Uvh https://packages.microsoft.com/config/centos/8/packages-microsoft-p
         # Required for asp.net core test runs
         sudo \
         libicu \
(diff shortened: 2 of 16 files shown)


ManickaP (approved):

liveans:
I will file issue for libmsquic missing dependencies on Ubuntu 24.04, once we merge this pr

ManickaP (approved):
LGTM, thanks!

src/alpine/3.17/helix/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apk add --upgrade --no-cache \
         openssl-dev \
         perl
 
-RUN git clone --depth 1 --single-branch --branch main --recursive https://github.com/microsoft/msquic /tmp/msquic
+RUN git clone --depth 1 --single-branch --branch v2.4.1 --recursive https://github.com/microsoft/msquic /tmp/msquic
```
mthalman, line 18: This pins to a specific version while the other Dockerfiles are essentially getting the latest available version. Pinning causes maintenance issues. It would be ideal if the Alpine Dockerfiles could have the same behavior. Consider querying the GitHub releases API to determine the latest release and targeting that branch instead. Example of such a pattern: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/75019f244373eeac6238e76dc72ed751a4f9f173/src/nanoserver/1809/helix/amd64/Dockerfile#L6-L8

liveans: It makes sense, but I'll probably do this as a follow-up PR.

mthalman: > It makes sense, but I'll probably do this as a follow-up PR.

That's fine.

> Ah, this might not work in this case, since we're not always using latest release on GitHub, we're actually using latest pre-release build.

That information should still be available from the API to make whatever choice you want.

mthalman (approved):

PR #1167, opened 2024-08-09
Switch to MCR images

richlander:
Time to switch to MCR images for Azure Linux 3.0.

Image to fixup later: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/4cf84c9cb8d9581f2f1f4e330a8199ede451c424/src/azurelinux/3.0/net9.0/android/Dockerfile#L18

Also, is there a reason we're using Java 8? That seems quite old.

@sbomer


jkoritzinsky (approved):
I know we have a few workarounds in some of our images around bugs in the previews. Can we remove any of them?

richlander:
Is that a question of feasibility (like are the underlying issues resolved) or something else?

jkoritzinsky:
Mainly the former (have they been resolved)

richlander:
I don't know what they were. I assume you and @sbomer have that context. Would be good to validate. We can also do that in another PR if  it is going to be a process.

mthalman (approved):

PR #1170, opened 2024-08-12
Add libmsquic to Azure Linux images

richlander:
Followed the pattern from https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1147.

@jkoritzinsky @sbomer

### DIFF
--- a/src/azurelinux/3.0/helix/Dockerfile
+++ b/src/azurelinux/3.0/helix/Dockerfile
@@ -35,7 +35,7 @@ RUN cmake -B build/linux/Release_openssl3 \
     cmake --build build/linux/Release_openssl3  --config Release && \
     cmake --install build/linux/Release_openssl3 --prefix /msquic
 
-FROM msquic as venv
+FROM mcr.microsoft.com/azurelinux/base/core:3.0 as venv
 
 RUN tdnf install --refresh -y \
     build-essential \
--- a/src/azurelinux/3.0/helix/Dockerfile
+++ b/src/azurelinux/3.0/helix/Dockerfile
@@ -5,6 +5,7 @@ ENV LANG=en_US.utf8
 
 # build MsQuic as we don't have packages
 RUN tdnf install --refresh -y \
+        binutils \
         ca-certificates-microsoft \
         cmake \
         gcc-c++ \
--- a/src/azurelinux/3.0/helix/Dockerfile
+++ b/src/azurelinux/3.0/helix/Dockerfile
@@ -1,29 +1,71 @@
-FROM mcr.microsoft.com/azurelinux/base/core:3.0
-
-# Install Helix Dependencies
+FROM mcr.microsoft.com/azurelinux/base/core:3.0 as msquic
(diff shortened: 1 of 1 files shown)


src/azurelinux/3.0/helix/Dockerfile
```diff
@@ -1,29 +1,76 @@
...
+       -DQUIC_TLS=openssl3 \
+       -DQUIC_ENABLE_LOGGING=true \
+       -DQUIC_USE_SYSTEM_LIBCRYPTO=true \
+       -DQUIC_BUILD_TOOLS=off \
+       -DQUIC_BUILD_TEST=off \
+       -DQUIC_BUILD_PERF=off && \
+    cmake --build build/linux/Release_openssl3  --config Release && \
+    cmake --install build/linux/Release_openssl3 --prefix /msquic
+
+FROM msquic as venv
```
jkoritzinsky, line 38: We should be able to extend from the base image for both the venv stage and the msquic stage.

```suggestion
FROM mcr.microsoft.com/azurelinux/base/core:3.0 as venv
```

richlander: I can do that. However, I did it because some of the same packages have already been installed. It seems wasteful to acquire them again, particularly when both the stages are throwaway. Fair or I'm motivated by the wrong thing?

jkoritzinsky: I like keeping the different stages isolated so we understand the dependencies of each. Also, by having these two not have dependencies on each other, we can have them build in parallel when building the image.

richlander: Fair enough. It was a great try to get this one past you! /s

Will make the change.

jkoritzinsky (approved):

src/azurelinux/3.0/helix/Dockerfile
richlander: Sorry. I didn't take your suggestion. I wanted to test it locally since I expected I had missed a dependency but it looks like only that change was required and then just committed and pushed it.

wfurt:
since Azure Linux is based on Centos, I would expect we can find good enough RPM. If not, we should push msquic to add support. cc: @liveans

jperrin:
Fine, I'll be the pedant for this issue. 

Technically we're more closely aligned to Fedora, though a mix of a few releases.  There are some packages in our 1.0 and 2.0 releases (openssl) that originated from CentOS, but it's not the majority. 

Fedora's even talked about this recently! [https://mastodon.social/@fedora@fosstodon.org/112928010076794066](https://mastodon.social/@fedora@fosstodon.org/112928010076794066)

wfurt:
the binary and package are identical for both at the moment. So it is really about feed preference until Azure Linux has it's own.

richlander:
We can update the Dockerfile once the package is available? I think they will have to ready soon.

liveans:
They are working on it microsoft/msquic#4455

Issue #997, opened 2024-04-05
[WIP] Spec for image use


richlander:
The following is the start of a spec for how `dotnet-buildtools-prereqs-docker` images should be used. The intent is that there are a set of approved patterns with the expectation that teams ask for help when the patterns do not work, not that they adopt (seemingly) expedient DIY solutions.

This spec will evolve over time and move to a new location.

## Native build

Azure Linux images are provided for native build. These are designed for dotnet/runtime, but may have utility for other repos. If you don't have any C or C++ code to build, don't use these images. They are also used for VMR builds, since those include runtime code.

These images all have `cross` in the name because we [cross build for all scenarios](https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/linux-instructions.md#docker-images), including x64.

Example images:

- `mcr.microsoft.com/dotnet-buildtools/prereqs:azure-linux-3.0-cross-amd64`
- `mcr.microsoft.com/dotnet-buildtools/prereqs:azure-linux-3.0-cross-arm64-alpine`

## Managed build

- Should build on Windows
- Use official source-build templates to validate that source-build leg runs

Source build images:

- External: Ubuntu 20.04
- Internal: Azure Linux 3.0

## Testing (in helix)

Current Helix images are enormous, for no obvious reason. We should define new images with very minimal components installed (including Python). We should create new Helix images as new distro versions are releases, such as Azure Linux 3.0, Ubuntu 24.04, and Alpine 3.20. We can decide if we want to backport the changes to older versions.

jkoritzinsky:
We have many projects (like sdk, efcore, etc) that are managed-only but want to validate that they build on Linux as they have developers/contributors that are Linux-first. We should have a Linux image for those builds (I think the "source build" images mentioned in your list fit these requirements well).

Additionally, the managed-source-build image is currently an image from a RHEL-based distro, which I think is a reasonable option to ensure that we don't horribly break our RH partners.

richlander:
Agreed that the source build leg solves multiple problems. We currently using CentOS Stream 9 for that purpose. AFAIK, we plan to continue using that distro (with the oldest supported version).

MichaelSimons:
> Source build images:

> External: Ubuntu 20.04
> Internal: Azure Linux 3.0

[...]

PR #1179, opened 2024-08-15
Add sudo to CBL-Mariner and AzureLinux distros

liveans:
Adding `sudo` makes it easier to work on container.

### DIFF
--- a/src/azurelinux/3.0/helix/Dockerfile
+++ b/src/azurelinux/3.0/helix/Dockerfile
@@ -63,6 +63,7 @@ RUN tdnf install --setopt tsflags=nodocs --refresh -y \
         python3 \
         python3-pip \
         shadow-utils \
+        sudo \
         tar \
         tzdata \
         which \
--- a/src/cbl-mariner/2.0/helix/amd64/Dockerfile
+++ b/src/cbl-mariner/2.0/helix/amd64/Dockerfile
@@ -13,6 +13,7 @@ RUN tdnf install --setopt tsflags=nodocs --refresh -y \
          llvm \
          python3-pip \
          shadow-utils \
+         sudo \
          tar \
          tzdata \
          which \


mthalman (approved):

richlander:
Can you provide more detail as to why this is needed? We're trying to limit `sudo` if possible.

@sbomer

liveans:
> Can you provide more detail as to why this is needed? We're trying to limit `sudo` if possible.

In the container itself user is helixbot and without `sudo`, I couldn't easily install some stuff and do some system-wide changes. This is just needed to work on the container itself without needing to rebuild the image.
I'm totally fine to revert it, but it's intention is just make it easier to work with image.

richlander:
> I couldn't easily install some stuff and do some system-wide changes

Can you elaborate on why you needed to do this?

liveans:
> > I couldn't easily install some stuff and do some system-wide changes
> 
> Can you elaborate on why this is needed? If you need to install stuff, why not install those components in the actual image? Or if you are using the image locally, why not launch as `root`?

https://github.com/microsoft/msquic/pull/4466 
We're basically adding our CI docker images into `msquic` pipeline to make sure that packaging works correctly and run simple .NET test on it, for that I'm not rebuilding image but using the publicly exposed one.
My first intention was to use images from this repo, but looks like it would make more sense to use https://hub.docker.com/r/microsoft/dotnet-runtime images.

So, I'll be happy to revert it.

richlander:
If you can get away with using one of our product images, that is much better. They are a ton smaller.

Thanks. It would be best if we didn't include `sudo` in these images.

MichaelSimons:
> It would be best if we didn't include sudo in these images.

I agree with this 100% - it would be good to have some checked in documentation with these types of guidelines.

mthalman:
Be aware that all of the Helix Dockerfiles install `sudo`. That's why I saw no issue with merging this, because these were the only two Helix Dockerfiles that were not doing that yet.

liveans:
> Be aware that all of the Helix Dockerfiles install `sudo`. That's why I saw no issue with merging this, because these were the only two Helix Dockerfiles that were not doing that yet.

I had the same impression as yours, and that's why I thought adding `sudo` into these images should be fine.

In that case, will there be work to remove `sudo` from other helix images?

richlander:
There is a slow path to tightening up our image create process (both content and the model for creating them). If `sudo` isn't needed, we shouldn't add it. Some of the more recent changes (driven by @jkoritzinsky) are really helping to limit the content in the images.

MichaelSimons:
The other thing to keep in mind are the changes @richlander is proposing in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/997 where we separate out native build, managed build, and test images as they each have different requirements.

PR #1197, opened 2024-09-06
Add brotli to sourcebuild container

agocke:
Part of https://github.com/dotnet/runtime/pull/107225

### DIFF
--- a/src/almalinux/8/source-build/amd64/Dockerfile
+++ b/src/almalinux/8/source-build/amd64/Dockerfile
@@ -10,6 +10,7 @@ RUN dnf install -y \
         "perl(Time::HiRes)" \
         autoconf \
         automake \
+        brotli-devel \
         cmake \
         elfutils \
         file \
--- a/src/alpine/3.19/amd64/Dockerfile
+++ b/src/alpine/3.19/amd64/Dockerfile
@@ -6,6 +6,7 @@ RUN apk add --upgrade --no-cache \
         automake \
         bash \
         build-base \
+        brotli-dev \
         ca-certificates \
         clang \
         clang-dev \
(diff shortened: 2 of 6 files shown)


mthalman (commented):
This new dependency will need to be applied equally to all the container images used for source build. Please ensure that all of these contain brotli if they don't already: https://github.com/dotnet/sdk/blob/7dc1e29627f20d47213ef2623b9e74b0e3cb4a4f/eng/pipelines/templates/variables/vmr-build.yml#L15-L28

agocke:
Thanks for the list, thought it was strange we only had one source build image.

agocke:
One thing is: I only need to update non-portable builds. Which ones are those?

mthalman:
> One thing is: I only need to update non-portable builds. Which ones are those?

All of the builds for source build are non-portable.

agocke:
Does source build support cross build? Or does it just share the container?

mthalman:
> Does source build support cross build? Or does it just share the container?

We don't run cross build in these containers for source build. Unified build has its own set of containers for cross build: https://github.com/dotnet/sdk/blob/7dc1e29627f20d47213ef2623b9e74b0e3cb4a4f/eng/pipelines/templates/variables/vmr-build.yml#L29-L48

mthalman (commented):
What about Fedora: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/fedora/40/amd64/Dockerfile. I see it already implicitly has `libbrotli` installed, but not `brotli-devel`.

agocke:
@mthalman Thanks, fixed.

mthalman (approved):

PR #1205, opened 2024-09-30
Change install repo from testing to community on Alpine for msquic

liveans:
As of [this PR](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/72604), libmsquic has been updated to 2.4.5 and moved to `community` repository from `testing` repository.

So, we're updating our images to use community edge repository, also from now on newer alpine versions (3.21 is around corner) will have msquic in `community/3.21` repository without needing to add extra repository (like we do in here `edge`).


ManickaP (approved):
Thanks!

PR #1226, opened 2024-10-30
Add dnf upgrade to Fedora images

MichaelSimons:
This allow the images to pick up available vulnerability fixes prior to a base image update.  This will help make the images more secure.

While doing this I updated the Fedora 39 and 40 images to used the same pattern introduced in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1219.  The one difference is that base dnf installed does not include config-manager so dnf-plugins-core must be installed before the packages.microsoft.com feed can be registered.

### DIFF
--- a/src/fedora/39/amd64/Dockerfile
+++ b/src/fedora/39/amd64/Dockerfile
@@ -1,9 +1,8 @@
-FROM library/fedora:39
+FROM fedora:39
 
-# Install the base toolchain we need to build anything (clang, cmake, make and the like)
-# this does not include libraries that we need to compile different projects, we'd like
-# them in a different layer.
-RUN dnf --setopt=install_weak_deps=False install -y \
+RUN dnf upgrade --refresh -y \
+    && dnf --setopt=install_weak_deps=False install -y \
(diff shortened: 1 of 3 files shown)


src/fedora/39/amd64/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM library/fedora:39
+FROM fedora:39
```
mthalman, line 1: Using `library` is necessary. See https://github.com/dotnet/docker-tools/issues/1341.

MichaelSimons: That was my fault.  I had tested locally and removed to workaround https://github.com/dotnet/docker-tools/issues/1363. 
 I forgot to add it back once done.

MichaelSimons: Fixed in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1226/commits/2ffb559d061bda985f9a6a67b980b8f71a992852

mthalman (approved):

PR #1246, opened 2024-11-05
Add the cpio package to the docker images we build with

jkoritzinsky:
This package is installed by default on regular distro installs, but is not available in our docker containers today. This is needed for the new rpm tooling in Arcade.


src/azurelinux/3.0/net9.0/crossdeps/amd64/Dockerfile
```diff
@@ -16,6 +16,7 @@ RUN tdnf update -y && \
         diffutils \
         icu \
         tar \
+        cpio \
```
mthalman, line 19: Please keep these in alphabetical order for each comment section. So this one should be placed after `awk`. Same applies to the other file.

mthalman (approved):

PR #1247, opened 2024-11-06
Add more missing tools to the images the runtime tooling uses

jkoritzinsky:
Apparently the file tool also isn't available in the base azl3 image by default. Add it as the rpm tooling requires it.

Add cpio to the source-build images as well.

### DIFF
--- a/src/almalinux/8/source-build/amd64/Dockerfile
+++ b/src/almalinux/8/source-build/amd64/Dockerfile
@@ -13,6 +13,7 @@ RUN dnf upgrade --refresh -y \
         automake \
         brotli-devel \
         cmake \
+        cpio \
         elfutils \
         file \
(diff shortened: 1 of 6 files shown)


ViktorHofer (approved):

mthalman (changes requested):
This is the full set of images used by source build: https://github.com/dotnet/dotnet/blob/da7a06999ab65a898fee9ca22f29874740f32f60/src/sdk/eng/pipelines/templates/variables/vmr-build.yml#L15-L28.

I assume that all of these will need to have `cpio` installed. I just tried the Alpine image and don't see it there.

jkoritzinsky:
@mthalman I've added the required packages to the remaining images.

mthalman (approved):

PR #1252, opened 2024-11-09
Fix tag for android-openssl image

akoeplinger:
The tag was missing the `cross` part which made it a bit confusing to map back to the correct Dockerfile. Add another tag so the existing one keeps working.

### DIFF
--- a/src/azurelinux/manifest.json
+++ b/src/azurelinux/manifest.json
@@ -345,7 +345,8 @@
               "os": "linux",
               "osVersion": "azurelinux3.0",
               "tags": {
-                "azurelinux-3.0-net9.0-android-openssl": {}
+                "azurelinux-3.0-net9.0-android-openssl": {},
+                "azurelinux-3.0-net9.0-cross-android-openssl": {}
               }
             }
           ]


mthalman (approved):

src/azurelinux/manifest.json
```diff
@@ -345,7 +345,8 @@
               "os": "linux",
               "osVersion": "azurelinux3.0",
               "tags": {
-                "azurelinux-3.0-net9.0-android-openssl": {}
+                "azurelinux-3.0-net9.0-android-openssl": {},
```
MichaelSimons, line 348: A comment might be useful as a reminder to clean this up during the next version upgrade.  Often times Dockerfiles and their associated manifest entries are copied during version upgrades.  It would be nice to get this cleanup up during that time.

akoeplinger: I thought about that but wasn't sure whether comments were supported in this .json

akoeplinger: hm looks like this file is parsed with json.net and I don't see the comment handling opt-in so I'm not sure it'd work

PR #1253, opened 2024-11-10
Add cpio to the ubuntu-24.04 Dockerfile

akoeplinger:
It was missed in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1246


am11:
@akoeplinger LGTM. I think we could switch to Azure Linux for Microsoft's VMR builds to align with runtime pipelines.

akoeplinger:
@am11 we already do that, the Ubuntu jobs are just to make sure we don't break building on a common developers' distro.

ViktorHofer (approved):

PR #1259, opened 2024-11-21
Schedule image rebuild once a month

### LINKED ISSUE #1224: Rebuild images on a regular basis
In order to keep the images fresh and up-to-date by default, there should be a policy or automation to rebuild images on a regular basis. This will ensure that latest package updates are retrieved when configured to do so. This will help to reduce the need to manually respond to any vulnerability alerts that might pop up.

Proposal: setup the pipeline to run on the 15th of every month, configured to run with caching disabled. That will be the day after the latest possible Patch Tuesday. It'd probably be ok to exclude Windows images as these will automatically be rebuilt on Patch Tuesday.

mthalman:
Fixes #1224

### DIFF
--- a/eng/pipelines/dotnet-buildtools-prereqs-all.yml
+++ b/eng/pipelines/dotnet-buildtools-prereqs-all.yml
@@ -7,6 +7,14 @@ trigger:
     - src/*
 pr: none
 
+schedules:
+- cron: "0 5 15 * *"
+  displayName: Nightly run
+  branches:
+    include:
+    - main
+  always: true
+
 resources:
   repositories:
   - repository: InternalVersionsRepo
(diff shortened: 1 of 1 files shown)


eng/pipelines/dotnet-buildtools-prereqs-all.yml
```diff
@@ -7,6 +7,14 @@ trigger:
     - src/*
 pr: none
 
+schedules:
+- cron: "0 5 15 * *"
+  displayName: Nightly run
```
MichaelSimons, line 12: "Nightly" feels a little odd here since it is only run once a month.  How about something more descriptive like "Monthly rebuild everything run" ... I'm not in love with that particular wording it is more descriptive.

mthalman: That's what I get for using copy and paste.

PR #1047, opened 2024-05-10
Move all Dockerfiles to arch-specific dir

NikolaMilosavljevic:
Fixes: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/87

Moving all Dockerfiles that are not in arch-specific directory to a new arch directory. Keeping Dockerfiles in directories that include architecture in the directory name.

PR #1262, opened 2024-11-21
Fix tag names for Azure Linux images

sbomer:
The tag names should match the directory structure, and should include the architecture for arch-specific images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The images produced from the Dockerfiles in this repository are published to the
 
 ## How to identify an image
 
-There are two tag formats used by the images from this repository. Most of the imagces follow the format `mcr.microsoft.com/dotnet-buildtools/prereqs:<os-name>-<os-version>-<variant>-<architecture>`.
+There are two tag formats used by the images from this repository. Most of the images follow the format `mcr.microsoft.com/dotnet-buildtools/prereqs:<os-name>-<os-version>-<variant>-<architecture>`.
 
 - `<os-name>` - Name of the Linux distribution or Windows OS the image is based on
 - `<os-version>` - Version of the OS
@@ -171,7 +171,7 @@ From this commit of the `image-info.dotnet-dotnet-buildtools-prereqs-docker-main
 
 The folder structure used in [src](./src) aligns with the tagging convention - `<os-name>-<os-version>-<variant>-<architecture>` or `<os-name>-<os-version>-<variant>-cross-<target>`.
 For example, the Dockerfile used to produce the `mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.20-amd64` image is stored in the [src/alpine/3.20/amd64](./src/alpine/3.20/amd64) folder.
-The Dockerfile used to produce the `mcr.`microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net8.0-cross-arm64` image is stored in the [src/azurelinux/3.0/net8.0/cross/arm64](./src/azurelinux/3.0/net8.0/cross/arm64) folder.
+The Dockerfile used to produce the `mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net8.0-cross-arm64` image is stored in the [src/azurelinux/3.0/net8.0/cross/arm64](./src/azurelinux/3.0/net8.0/cross/arm64) folder.
 
 If a Dockerfile is shared across multiple architectures, then the `<architecture>` folder should be omitted.
 For example, the [src\alpine\3.20\helix\Dockerfile](./src/alpine/3.20/helix/Dockerfile) is built for all supported architectures (amd64, arm64 and arm) therefore there is no architecture folder in its path.
--- a/README.md
+++ b/README.md
@@ -30,6 +30,11 @@ The cross-compilation images follow the format `mcr.microsoft.com/dotnet-buildto
 
 - `<target>` - Specifies the target for cross-compilation, including the targeted architecture and libc variant (glibc if not specified).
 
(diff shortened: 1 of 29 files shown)


akoeplinger (approved):

mthalman (commented):
I like these changes. It brings consistency and clarity. But I think we need to update the documentation to reflect this pattern because it is a different pattern than what is currently used for non-cross images.

I'm specifically referring to the pattern documented at https://github.com/dotnet/dotnet-buildtools-prereqs-docker?tab=readme-ov-file#how-to-identify-an-image. The difference with the cross images is that the architecture specified refers to the cross target, not the architecture of the image.

mthalman:
> Thanks! The docs currently say this about the arch:
> 
> > The tag format used by the images from this repository is mcr.microsoft.com/dotnet-buildtools/prereqs:<os-name>-<os-version>-<variant>-<architecture>
> > ...
> > <architecture> - Architecture of the OS (amd64 shall be implied if not specified).
> 
> > The folder structure used in [src](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src) aligns with the tagging convention
> 
> > If a Dockerfile is shared across multiple architectures, then the  folder should be omitted.
> 
> Personally I like the policy of allowing `-amd64` to be omitted from the tag format where it's implied, and keeping the simpler tags like `azurelinux-3.0-net8.0-crossdeps-builder`.
> 
> I propose we:
> 
> * Allow omitting `amd64` from the tag names _and_ folder structure, and undo [Move all Dockerfiles to arch-specific dir #1047](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1047) for images where we take advantage of this option.
> * Update docs to recommend the current pattern (`-<target>`) that we use for the cross-compilation images.
> 
> cc @richlander

I don't like removing it from the directory name. Because there are some Dockerfiles which are truly multi-arch and so they have no architecture directory. Saying that an amd64Dockerfile can get by without being contained in an amd64 directory causes confusion and ambiguity compared to the multi-arch ones.

For tags, I initially was ambivalent. But thinking more about it, I kind of feel it needs to be explicit because of the ambiguity that arises from it. One could imagine have multi-arch tags for our multi-arch Dockerfiles. And that would most definitely not have an arch in the tag name.

So I think the documented pattern should be changed to say that all architectures should be explicit both in tags and directories.

README.md
```diff
@@ -12,20 +12,24 @@ The images produced from the Dockerfiles in this repository are published to the
 
 ## How to identify an image
 
-The tag format used by the images from this repository is `mcr.microsoft.com/dotnet-buildtools/prereqs:<os-name>-<os-version>-<variant>-<architecture>`
+There are two tag formats used by the images from this repository. Most of the imagces follow the format `mcr.microsoft.com/dotnet-buildtools/prereqs:<os-name>-<os-version>-<variant>-<architecture>`.
```
mthalman, line 15:
```suggestion
There are two tag formats used by the images from this repository. Most of the images follow the format `mcr.microsoft.com/dotnet-buildtools/prereqs:<os-name>-<os-version>-<variant>-<architecture>`.
```

README.md
```diff
@@ -12,20 +12,24 @@ The images produced from the Dockerfiles in this repository are published to the
...
 
 Examples:
 
-- mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.20
+- mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.20-amd64
 - mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-helix-arm64v8
 
+The cross-compilation images follow the format `mcr.microsoft.com/dotnet-buildtools/prereqs:<os-name>-<os-version>-<variant>-cross-<target>`. These are all implicitly amd64 images.
+
+- `<target>` - Specifies the target for cross-compilation, including the targeted architecture and libc variant (glibc if not specified).
```
mthalman, line 31: I think providing some examples like the first format would be helpful here.

README.md
```diff
@@ -165,8 +169,9 @@ From this commit of the `image-info.dotnet-dotnet-buildtools-prereqs-docker-main
 
 ### Source Folder Structure
 
-The folder structure used in [src](./src) aligns with the tagging convention - `<os-name>-<os-version>-<variant>-<architecture>`.
-For example, the Dockerfile used to produce the `mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.20` image is stored in the [src/alpine/3.20/amd64](./src/alpine/3.20/amd64) folder.
+The folder structure used in [src](./src) aligns with the tagging convention - `<os-name>-<os-version>-<variant>-<architecture>` or `<os-name>-<os-version>-<variant>-cross-<target>`.
+For example, the Dockerfile used to produce the `mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.20-amd64` image is stored in the [src/alpine/3.20/amd64](./src/alpine/3.20/amd64) folder.
+The Dockerfile used to produce the `mcr.`microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net8.0-cross-arm64` image is stored in the [src/azurelinux/3.0/net8.0/cross/arm64](./src/azurelinux/3.0/net8.0/cross/arm64) folder.
```
mthalman, line 174:
```suggestion
The Dockerfile used to produce the `mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net8.0-cross-arm64` image is stored in the [src/azurelinux/3.0/net8.0/cross/arm64](./src/azurelinux/3.0/net8.0/cross/arm64) folder.
```

mthalman (approved):

PR #1263, opened 2024-11-21
Add Azure Linux images for .NET 10

sbomer:
Contributes to https://github.com/dotnet/runtime/issues/109939. These images have LLVM 19.1.0, but will be updated to LLVM 20 per the plan, once that is released.


mthalman (approved):

richlander:
Nice! There is nothing to do now, but we should we prepared to move to Azure Linux 4.0 at some point during the .NET 10 release.

PR #1270, opened 2024-11-26
Replace alpine tags in Azure Linux images with musl

sbomer:
We want to move to `musl` tags since that naming better reflects the libc compatibility provided by these images. This replaces `alpine` with `musl` in the .NET 10 build images, and adds `musl` variants (keeping the old `alpine` tags for compat) to the .NET 8 and .NET 9 images. I'm not changing the cbl-mariner-2.0 or ubuntu cross images because we're planning to remove those.

Question on the naming: do we prefer `amd64-musl` or `musl-amd64`? `amd64-musl` better matches the format of a target triple (`<arch><sub>-<vendor>-<sys>-<env>`), but `musl-amd64` better matches our RIDs (`linux-musl-x64`).


src/azurelinux/manifest.json
```diff
@@ -598,11 +605,11 @@
         },        {
           "platforms": [
             {
-              "dockerfile": "src/azurelinux/3.0/net10.0/cross/riscv64-alpine",
+              "dockerfile": "src/azurelinux/3.0/net10.0/cross/riscv64-musl",
```
mthalman, line 608: None of the 10.0 Dockerfiles have been moved to the new directory.

mthalman (commented):
I don't have any particular opinion on which naming pattern to use.

richlander (approved):

Issue #1260, opened 2024-11-21
Remove Debian 11 Dockerfiles


mthalman:
Debian 11 was EOL in Aug 2024 so these Dockerfiles should no longer be maintained.

mthalman:
Fixed by https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1606

PR #1276, opened 2024-12-03
Add Debian 12 Dockerfile

mthalman:
This is intended to be the replacement for the Debian 11 Dockerfile that will be removed by https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1260.

### DIFF
--- a/src/debian/12/amd64/Dockerfile
+++ b/src/debian/12/amd64/Dockerfile
@@ -1,0 +1,60 @@
+FROM library/debian:bookworm
+
+# Dependencies for generic .NET Core builds and the base toolchain we need to
+# build anything (clang, cmake, make and the like)
+RUN apt-get update \
+    && apt-get upgrade -y \
+    && apt-get install -y \
+        apt-transport-https \
+        curl \
+        software-properties-common \
+    && curl -sL https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb -o packages-microsoft-prod.deb \
+    && dpkg -i packages-microsoft-prod.deb \
+    && rm packages-microsoft-prod.deb \
+    && apt-get update \
+    && apt-get install -y \
+        autoconf \
+        automake \
+        azure-cli \
+        build-essential \
+        clang \
+        cmake \
+        elfutils \
+        file \
+        g++ \
+        gettext \
+        gdb \
+        git \
+        gnupg \
+        jq \
+        libcurl4-openssl-dev \
+        libgdiplus \
+        libicu-dev \
+        libkrb5-dev \
+        liblldb-dev \
+        liblttng-ust-dev \
+        libnuma-dev \
+        libssl-dev \
+        libssl1.1 \
+        libtool \
+        libunwind8-dev \
+        lldb \
+        llvm \
+        locales \
+        make \
+        pigz \
+        powershell \
+        python-lldb \
+        sudo \
+        tar \
+        uuid-dev \
+        zip \
+        zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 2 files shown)


richlander:
Who is using this style of image at this point. My intuition is that we should have Azure Linux build images and helix test images for everything else. Perhaps I'm missing a scenario.

@sbomer

MichaelSimons:
> Who is using this style of image at this point. My intuition is that we should have Azure Linux build images and helix test images for everything else. Perhaps I'm missing a scenario.

One important instance using this style of image is source-build (although it doesn't use the Debian image specifically).  Azure Linux would not be appropriate for Source Build as it is trying to build as close as possible to how maintainers would.

The PowerBI Docker usage report is useful in answering your question.  Regarding the Debian image specifically, I only see one usage - dotnet-diagnostics.  They are using it for a test leg - https://github.com/dotnet/diagnostics/blob/d12018ba56f7db766272699877908668cba517ef/eng/pipelines/pipeline-resources.yml#L70.  Likely the builder image is overkill for their usage.

richlander:
Last time I looked at the diagnostics repo, it seems like they were using old patterns. I think we need to stop providing them with images perpetuating old patterns. We should be certain we need non-helix images going forward.

MichaelSimons:
> Last time I looked at the diagnostics repo, it seems like they were using old patterns. I think we need to stop providing them with images perpetuating old patterns. We should be certain we need non-helix images going forward.

@hoyosjs and @mikem8361, can you comment on what the Debian requirements are for the [diagnostic test leg](https://github.com/dotnet/diagnostics/blob/d12018ba56f7db766272699877908668cba517ef/eng/pipelines/pipeline-resources.yml#L70) that is using the buildtools Debian image?  It is currently utilizing a "builder" image which carries build tools like cmake, etc.

mikem8361:
> > Last time I looked at the diagnostics repo, it seems like they were using old patterns. I think we need to stop providing them with images perpetuating old patterns. We should be certain we need non-helix images going forward.
> 
> @hoyosjs and @mikem8361, can you comment on what the Debian requirements are for the [diagnostic test leg](https://github.com/dotnet/diagnostics/blob/d12018ba56f7db766272699877908668cba517ef/eng/pipelines/pipeline-resources.yml#L70) that is using the buildtools Debian image? It is currently utilizing a "builder" image which carries build tools like cmake, etc.

The diagnostics repo only uses the Debian image for testing so it only needs lldb installed. We don't build anything on it.

richlander:
Do you need an lldb of a particular version or the one in the distro will typically work?

mikem8361:
The version that contains with the distro should be good enough.

mthalman:
Regarding whether a Helix image should be used by diagnostics for testing...

AFAICT, Helix images will not work for repos that are running their testing as a container job. That's because Helix images define a custom user which messes up the requirements needed by AzDO:

> ##[error]Docker-exec executed: `groupadd azure_pipelines_sudo`; container id: `e5cdaf28a120c102036dcc6b83453c4c0797b94293286c80c71f7cc153081e8b`; exit code: `10`; command output: `groupadd: Permission denied.`, `groupadd: cannot lock /etc/group; try again later.`

It seems that AzDO expects it should be able to run `groupadd` without using `sudo`. That doesn't work with Helix images and I don't see a way to configure them to allow for this.

I have a PR that ran into this at https://github.com/dotnet/diagnostics/pull/5197.

wfurt:
> It seems that AzDO expects it should be able to run `groupadd` without using `sudo`. That doesn't work with Helix images and I don't see a way to configure them to allow for this.
>
Traditionally I saw that solved either with code branches or better yet with variables. let say `${SUDO}`  can be either `sudo` or empty all we would need is replacing offending code with something like `$SUDO grouped ....`

PR #1277, opened 2024-12-04
Add 24.10 helix images

richlander:
Prep for the eventual 26.04 release. I don't think we've typically added the regular Ubuntu releases. I believe we should.

There is no libmsquic available so took the easy path. The `2.4.5` image required more packages so the PR is using `2.4.1`.


mthalman (approved):

richlander:
Hey @wfurt -- Wasn't able to get the 24.04 package to install on Arm32. I suspect it relates to a package rename from `libssl3` to `libssl3t64`. It's not pressing and we don't need Arm32 testing on 24.10. It's something to get resolved by 26.04 or earlier.

PR #1278, opened 2024-12-04
Bumped NDK version to 27

jkurdek:
Updates the container to use LTS Android NDK.


jkoritzinsky:
Can you remove the deletion of clang-tidy now that we're updating the NDK?

jkurdek:
@jkoritzinsky should we also remove the deletion of python site packages? NDK 27 contains python 3.11

jkurdek:
I checked - python needs to be removed

jkoritzinsky:
What's the path of the `npm/cross-spawn` one?

The System.Formats.Asn1 one isn't one I'm worried about (as that's one of our packages, the docker team knows how to fix that one).

jkoritzinsky (approved):

akoeplinger:
npm/cross-spawn is coming from `/usr/lib/node_modules/npm/node_modules/cross-spawn/package.json` so from the upstream Azure Linux npm package.

The System.Formats.Asn1 is coming from `/opt/microsoft/powershell/7/Modules/PSReadLine/_manifest/spdx_2.2/manifest.spdx.json`

src/azurelinux/3.0/net9.0/android/amd64/Dockerfile
```diff
@@ -27,8 +27,7 @@ RUN /usr/local/cmdline-tools/cmdline-tools/bin/sdkmanager --sdk_root="${ANDROID_
 
 # We can't upgrade the NDK version as the runtime repo requires tooling that only exists up to NDK 23
 # Remove all components of NDK 23 that are flagged by security scanners
```
akoeplinger, line 29: comment should be updated since it doesn't apply anymore

jkoritzinsky:
In that case, I think both of those are "fine" in the sense that S360 hasn't flagged them for us to fix.

jkurdek:
@akoeplinger if we remove Java copying here, what runtime changes would we have to implement?

akoeplinger:
none, we'd just install the `msopenjdk-17` package instead of the `COPY --from=mcr.microsoft.com/openjdk/jdk:17-mariner $JAVA_HOME $JAVA_HOME`. I can send a PR later, it's just an optimization.

jkurdek:
Ready to merge, runtime should be ready to work with NDK 27 on main and .NET 9 branches

PR #1280, opened 2024-12-04
Add a Fedora 41 helix image

jkoritzinsky:
We don't currently have any supported docker images for testing on Fedora. Add one for Fedora 41 so we can test with a supported docker image.

### DIFF
--- a/src/fedora/41/helix/Dockerfile
+++ b/src/fedora/41/helix/Dockerfile
@@ -1,4 +1,4 @@
-FROM library/fedora:40 AS venv
+FROM library/fedora:41 AS venv
 
 RUN dnf upgrade --refresh -y && \
     dnf install --setopt tsflags=nodocs -y \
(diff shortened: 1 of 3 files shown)


src/fedora/41/helix/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+    dnf install --setopt tsflags=nodocs -y \
+        dnf-plugins-core && \
+    dnf install --setopt=install_weak_deps=False --setopt tsflags=nodocs -y \
+        openssl \
+        python3-devel \
+        python3-pip \
+        gcc \
+        libatomic \
+        redhat-rpm-config && \
+    dnf clean all
```
mthalman, line 13: Please follow this convention instead which is what we're trying to consolidate on for all the Dockerfiles. This should get applied consistently for the rest of this Dockerfile as well.

```suggestion
RUN dnf upgrade --refresh -y \
    && dnf install --setopt tsflags=nodocs -y \
        dnf-plugins-core \
    && dnf install --setopt=install_weak_deps=False --setopt tsflags=nodocs -y \
        openssl \
        python3-devel \
        python3-pip \
        gcc \
        libatomic \
        redhat-rpm-config \
    && dnf clean all
```

richlander (approved):

PR #1282, opened 2024-12-05
Add Alpine 3.21 helix images

richlander:


mthalman (approved):

src/alpine/3.21/helix/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+        numactl \
+        openssl \
+        python3 \
+        python3-dev \
+        py3-pip \
+        sudo \
+        tzdata
+
+# Install libmsquic from testing repository
+RUN apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community/ libmsquic
```
liveans, line 41: We don't need to pull `libmsquic` from testing repository anymore on alpine starting with version 3.21, we should able to pull it directly as

`apk add libmsquic`

richlander: I had noticed that and was wondering if this would be the guidance. Cool!

I'll make a PR for that.

richlander: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1292

PR #1283, opened 2024-12-05
Update EMSDK in net8.0 image

sbomer:
Addresses some of the vulnerabilities in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1281.

Issue #1281, opened 2024-12-04
Vulnerabilities in the azurelinux-3.0-net8.0-webassembly-amd64 image


MichaelSimons:
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/253
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/252
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/251
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/250
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/249
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/248
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/247
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/246
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/245
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/244
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/243
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/242
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/241
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/240
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/92
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/82
- https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/80

MichaelSimons:
@sbomer - can you take a look at these vulnerabilities and try to address them?  TIA.

mthalman:
No, not without going through a big exception approval process. The proper mitigation here is to update the configuration of the Dockerfile so that it doesn't reference vulnerable versions.

PR #1285, opened 2024-12-06
Revert "Update EMSDK in net8.0 image"

sbomer:
Reverts dotnet/dotnet-buildtools-prereqs-docker#1283, see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1283#issuecomment-2523553684.

@akoeplinger do you have guidance on addressing https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1281? Currently runtime is using a stale image for .NET 8:
https://github.com/dotnet/runtime/blob/c8d152ebbc2c97b6cba580fa8079d4e3b2af1516/eng/pipelines/common/templates/pipeline-with-resources.yml#L101 We need to remove the stale reference, but it's for an image that had version 3.1.34 of the EMSDK.


akoeplinger (approved):

akoeplinger:
None of the vulnerabilities are relevant in our case since we don't run untrusted input and aren't using nodejs in a webserver context. Do you know if we can close them as "won't fix"?

MichaelSimons:
> None of the vulnerabilities are relevant in our case since we don't run untrusted input and aren't using nodejs in a webserver context. Do you know if we can close them as "won't fix"?

I am not familiar with the EMSDK support model.  I see the version we are on is well over a year now.  Will we be staying on this version for the lifetime of 8.0?  This feels like a recipe for numerous vulnerabilities over it's lifetime.  Requesting s360 exceptions is not feasible for these.  They are going to tell us to update our dependency.

Issue #1286, opened 2024-12-06
`libmsquic` support for Ubuntu 24.10


richlander:
I tried to install `libmsquic` on Ubuntu 24.10 and was unsuccessful. I'm going to need some help with that. I'm seeing the usual challenges with new OSes.

Dockerfile: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/ubuntu/24.10/helix/Dockerfile

Subset of CI errors:

```bash
+ ./RunTests.sh --runtime-path /root/helix/work/correlation
  Discovering: System.Net.Quic.Functional.Tests (method display = ClassAndMethod, method display options = None)
  Discovered:  System.Net.Quic.Functional.Tests (found 5 of 153 test cases)
  Starting:    System.Net.Quic.Functional.Tests (parallel test collections = on [2 threads], stop on fail = off)
MsQuic not supported and using 'unknown'.
    System.Net.Quic.Tests.MsQuicPlatformDetectionTests.SupportedLinuxPlatformsWithMsQuic_IsSupportedIsTrue [FAIL]
      Assert.True() Failure
      Expected: True
      Actual:   False
      Stack Trace:
        /_/src/libraries/System.Net.Quic/tests/FunctionalTests/MsQuicPlatformDetectionTests.cs(51,0): at System.Net.Quic.Tests.MsQuicPlatformDetectionTests.SupportedLinuxPlatformsWithMsQuic_IsSupportedIsTrue()
        --- End of stack trace from previous location ---
      Output:
        /usr/lib/x86_64-linux-gnu/libmsquic.so.2
        /usr/lib/x86_64-linux-gnu/libmsquic.so.2.4.1
        
    System.Net.Quic.Tests.MsQuicPlatformDetectionTests.SupportedLinuxPlatforms_IsSupportedIsTrue [FAIL]
      Assert.True() Failure
      Expected: True
      Actual:   False
```

As is obvious, I'm currently using `2.4.1`. It looks like more package dependencies came after that. I wasn't able to get the package installed with `2.4.7`.

```bash
root@46a0173c3749:/# dpkg -i libmsquic_2.4.7_amd64.deb 
Selecting previously unselected package libmsquic.
(Reading database ... 4984 files and directories currently installed.)
Preparing to unpack libmsquic_2.4.7_amd64.deb ...
Unpacking libmsquic (2.4.7) ...
dpkg: dependency problems prevent configuration of libmsquic:
[...]

PR #1288, opened 2024-12-07
Add Debian trixie image

richlander:
It's another Debian release year! My assumption is that we'll start shipping Trixie images in Preview 1, in February. We should look at starting testing in January.

I was unable to make Arm32 work per https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1286

I adopted a slightly different pattern, which I think is superior:

- One `apt-get updated` is consumed by all stages.
- Upstream image is mentioned once
- Removed `apt-get upgrade -y`, which seems wasteful.

### DIFF
--- a/src/debian/13/helix/Dockerfile
+++ b/src/debian/13/helix/Dockerfile
@@ -8,8 +8,7 @@ RUN apt-get install -y \
         coreutils \
         python3-dev \
         python3-pip \
-        python3-venv \
-    && rm -rf /var/lib/apt/lists/*
+        python3-venv
 
 RUN python3 -m venv /venv \
     && . /venv/bin/activate \
@@ -20,14 +19,12 @@ FROM base
 ARG TARGETARCH
 
 # Install Helix Dependencies
-RUN apt-get install -y \
+RUN LIBCURL=libcurl4 \
+    && if [ "$TARGETARCH" = "arm" ]; then \
+           LIBCURL="libcurl4t64"; fi \ 
+    && apt-get install -y \
         apt-transport-https \
         curl \
-    && curl -sOL https://packages.microsoft.com/config/debian/12/packages-microsoft-prod.deb \
-    && dpkg -i packages-microsoft-prod.deb \
-    && rm packages-microsoft-prod.deb \
-    && apt-get update \
-    && apt-get install -y \
         autoconf \
         automake \
         at \
@@ -38,9 +35,8 @@ RUN apt-get install -y \
         iputils-ping \
         libcurl4 \
         libffi-dev \
-        libgdiplus \
         libicu-dev \
-        libmsquic \
+        libnuma1 \
         libssl-dev \
         libtool \
         libunwind8 \
@@ -56,6 +52,10 @@ RUN apt-get install -y \
         unzip \
     && rm -rf /var/lib/apt/lists/* \
     && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+
+RUN curl -O https://packages.microsoft.com/ubuntu/24.04/prod/pool/main/libm/libmsquic/libmsquic_2.4.7_$TARGETARCH.deb \
+    && dpkg -i libmsquic* \
+    && rm libmsquic*
 
 ENV LANG=en_US.utf8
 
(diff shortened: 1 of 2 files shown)


richlander:
I forgot that we don't have Trixie images in our mirror.

@lbussell

mthalman:
> I forgot that we don't have Trixie images in our mirror.
> 
> @lbussell

Fixed

src/debian/manifest.json
```diff
@@ -110,7 +110,34 @@
...
+                "debian-13-helix-amd64": {}
+              }
+            }
+          ]
+        },
+        {
+          "platforms": [
+            {
+              "architecture": "arm64",
+              "dockerfile": "src/debian/12/helix/",
```
mthalman, line 131:
```suggestion
              "dockerfile": "src/debian/13/helix/",
```

mthalman (commented):
> Removed apt-get upgrade -y, which seems wasteful.

Running an upgrade was intentional. See https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1226 for an explanation.

richlander:
There is not. I was going to ping @rbhanda about that.

richlander:
> Running an upgrade was intentional. 

I assumed. Why do we have a different practice for this repo than for dotnet/dotnet-docker? Is it that we have a different tradeoff on size vs security/compliance?

rbhanda:
> There is not. I was going to ping @rbhanda about that.

https://github.com/dotnet/core/issues/9649

mthalman:
> > Running an upgrade was intentional.
> 
> I assumed. Why do we have a different practice for this repo than for dotnet/dotnet-docker? Is it that we have a different tradeoff on size vs security/compliance?

Exactly. We don't have the same strictness in these images. It's ok if there's some wasted size in them if it makes our lives easier for maintaining them.

src/debian/13/helix/Dockerfile
```diff
@@ -0,0 +1,75 @@
+FROM library/debian:trixie AS base
+
+RUN apt-get update
```
MichaelSimons, line 3: This adds bloat to the image.  It defeats the purpose of running `rm -rf /var/lib/apt/lists/*` later on.   Having a smaller image is more important that a faster image build.

richlander:
@liveans -- I tested the latest `libmsquic` on Arm64 and it works there. It doesn't work on x64 for some reason.

I see this:

```bash
root@ad82881d09ec:/# dpkg -i libmsquic_2.4.7_amd64.deb 
Selecting previously unselected package libmsquic.
(Reading database ... 5854 files and directories currently installed.)
Preparing to unpack libmsquic_2.4.7_amd64.deb ...
Unpacking libmsquic (2.4.7) ...
dpkg: dependency problems prevent configuration of libmsquic:
 libmsquic depends on libxdp1; however:
  Package libxdp1 is not installed.
 libmsquic depends on libnl-route-3-200; however:
  Package libnl-route-3-200 is not installed.

dpkg: error processing package libmsquic (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
 libmsquic
```

richlander:
Once the libmsquic issues are resolved, I'm going to do some quick size analysis based on the other feedback.

liveans:
Can you also install these packages on x64?

`libnuma1 libelf1t64 libbpf1 libxdp1 libnl-3-200 libnl-route-3-200`

After than that, it should work

richlander:
Done. I tested on x64 and Arm64 and that worked. I split the images into their own section so that they are easy to remove when we choose to install the libmsquic package.

richlander:
This one passes! Thanks @liveans!

We're now down to more stylistic Dockerfile review.

src/debian/13/helix/Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+    && apt-get install -y \
+        cargo \
+        coreutils \
+        libffi-dev \
+        libssl-dev \
+        pkg-config \
+        python3-dev \
+        python3-pip \
+        python3-venv \
+        && rm -rf /var/lib/apt/lists/*
```
mthalman, line 13:
```suggestion
    && rm -rf /var/lib/apt/lists/*
```

mthalman (approved):

PR #1293, opened 2024-12-10
Update Ubuntu 24.10 libmsquic version

richlander:
Related to #1288

@liveans

### DIFF
--- a/src/ubuntu/24.10/helix/Dockerfile
+++ b/src/ubuntu/24.10/helix/Dockerfile
@@ -1,8 +1,8 @@
 FROM ubuntu.azurecr.io/ubuntu:oracular AS venv
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && \
-    apt-get install -y \
+RUN apt-get update \
+    && apt-get install -y \
         cargo \
         pkg-config \
         libffi-dev \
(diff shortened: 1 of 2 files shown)


liveans:
Seems like on ARM32, it's trying to grab _arm postfix package which does not exist, it should grab armhf instead.
For x64 it has the same symptoms as Debian 13, which means it requires XDP and it's dependencies but not installed on the system, I'll give it a look and will help you with workaround

liveans:
@richlander Can you also install these packages on x64?

libnuma1 libelf1t64 libbpf1 libxdp1 libnl-3-200 libnl-route-3-200

After than that, it should work

src/ubuntu/24.10/helix/Dockerfile
```diff
@@ -1,32 +1,31 @@
...
+    && apt-get install -y \
         cargo \
         pkg-config \
         libffi-dev \
         coreutils \
         python3-dev \
         python3-pip \
         python3-venv \
-        libssl-dev \
-    && rm -rf /var/lib/apt/lists/*
```
mthalman, line 14: This cleanup should be added back.

src/ubuntu/24.10/helix/Dockerfile
```diff
@@ -1,32 +1,31 @@
...
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN LIBCURL=libcurl4 && \
-    if [ "$TARGETARCH" = "arm" ]; then \
-        LIBCURL="libcurl4t64"; fi && \ 
-    apt-get update && \
-    apt-get install -qq -y \
+RUN LIBCURL=libcurl4 \
+    && if [ "$TARGETARCH" = "arm" ]; then \
+           LIBCURL="libcurl4t64"; fi \
```
mthalman, line 26: The extra indentation here seems unnecessary.

src/ubuntu/manifest.json
```diff
@@ -476,6 +476,20 @@
...
+              "dockerfile": "src/ubuntu/24.10/helix",
+              "os": "linux",
+              "osVersion": "oracular",
+              "tags": {
+                "ubuntu-24.10-helix-arm32v7": {}
+              },
+              "variant": "v7"
+            }
+          ]
+        },
```
mthalman, line 492: Nit: to follow the existing convention, this should be listed after arm64.

src/ubuntu/24.10/helix/Dockerfile
richlander: It has no benefit. It is in a stage that doesn't contribute to the end product.

src/ubuntu/manifest.json
richlander: Oops. Yes.

src/ubuntu/24.10/helix/Dockerfile
MichaelSimons: It is beneficial IMO for two reasons.  First, the cleanup is a Dockerfile best practice.  Folks copy these patterns all over so following the best practices everywhere is beneficial.  Second, from a perf perspective you are still writing the cache in the layer.  In my experience deleting the cache is faster that writing it in the layer.  Admittingly this is trivial.

richlander:
This PR now passes! Thanks @liveans!

We're now down to more stylistic Dockerfile review.

src/ubuntu/24.10/helix/Dockerfile
```diff
@@ -1,32 +1,32 @@
...
         coreutils \
+        libffi-dev \
+        libssl-dev \
+        pkg-config \
         python3-dev \
         python3-pip \
         python3-venv \
-        libssl-dev \
-    && rm -rf /var/lib/apt/lists/*
+        && rm -rf /var/lib/apt/lists/*
```
mthalman, line 14:
```suggestion
    && rm -rf /var/lib/apt/lists/*
```

mthalman (approved):

PR #1295, opened 2024-12-11
Floating tag support doc

mthalman:
Contributes to https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1279

This provides a policy that is discussed in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1279. I want to get consensus on the policy first and then will make tag updates in a later PR.

### DIFF
--- a/lifecycle.md
+++ b/lifecycle.md
@@ -57,7 +57,7 @@ Repos that are susceptible to breaking changes in the distro should use these ta
 [dotnet/runtime](https://github.com/dotnet/runtime) is an example of such a repo.
 
 > [!NOTE]  
-> There are plans to automate the creation of PRs to update image references to new distro versions as they become available: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1321.
+> There are plans to [automate the creation of PRs to update image references](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1321) to new distro versions as they become available.
 
 ### Floating Tags
 
@@ -73,8 +73,8 @@ If a new floating tag is desired, log an issue requesting it.
 The maintainers of this repo follow a workflow before a floating tag gets moved to a newer distro version:
 1. First, version-specific tags for the new distro version are provided.
 1. After a one month evaluation period, the new distro version is ready to be rolled out to floating tags according to the schedule, assuming there are no issues found.
-  1. For the .NET version currently in development, the floating tag is moved to reference the new distro version as soon as the evaluation period has been met.
-  1. For servicing versions of .NET, the floating tag is moved to reference the new distro version one month after the in-development .NET version has been updated. This is done as soon as the branch is open for a servicing release.
+   1. For the .NET version currently in development, the floating tag is moved to reference the new distro version as soon as the evaluation period has been met.
+   1. For servicing versions of .NET, the floating tag is moved to reference the new distro version one month after the in-development .NET version has been updated. This is done as soon as the branch is open for a servicing release.
 
 #### Stability Period
 
--- a/lifecycle.md
+++ b/lifecycle.md
@@ -48,22 +48,44 @@ Note: this content should be moved to another location as it is not lifecycle re
 
 Build and test images are referenced in repo infra files, across a variety of `main` and `release/*` branches. Updating these references is a multi-step detail-oriented task. It is a pain, but necessary.
 
-Two types of tag styles are available:
-* version-specific: The tag name includes the distro's version. Example: `alpine-3.21-helix-amd64`
-* floating (optional): The tag name has no version. Instead it is routinely updated to reference a new version as the distro's lifecycle progresses. There are two sub-types of floating tags:
-  * latest: References the latest validated version of the distro. Example: `alpine-latest-helix-amd64`
-  * oldest: References the oldest supported version of the distro. Example: `alpine-oldest-helix-amd64`
+Two types of tag styles are available: [distro version-specific](#distro-version-specific-tags) and [floating](#floating-tags).
 
-FLoating tags are provided on an as-needed basis. They are beneficial for repos that are not susceptible to breaking changes that occur from new distro versions because the source that references the tag doesn't need to be updated in order to make use of the new version.
-Conversely, some repos may be susceptible to distro breaking changes in which case the version-specific tags should be used ([dotnet/runtime](https://github.com/dotnet/runtime) is an example of such a repo).
+### Distro Version-Specific Tags
 
-There is an evaluation period before a `latest` floating tag gets moved to a newer distro version.
-First, version-specific tags for the new distro version are provided.
-After a one month evaluation period, the `latest` floating tag is updated to reference the new version, assuming there are no issues found.
+Distro version-specific tags include the distro's version (e.g. `alpine-3.21-helix-amd64`).
+Repos that are susceptible to breaking changes in the distro should use these tags.
+[dotnet/runtime](https://github.com/dotnet/runtime) is an example of such a repo.
 
-At times, it may be necessary to use a [fixed image reference](https://github.com/dotnet/runtime/pull/110199#discussion_r1859075989) for build reliability.
-This is done by referencing the digest of the specific image that is needed (e.g. `mcr.microsoft.com/dotnet-buildtools/prereqs@sha256:56feee03d202e008a98f3c92784f79f3f0b3a512074f7f8ee2b1ba4ca4c08c6e`).
-If this is ever done, a tracking issue should be created (before the PR is merged) so that we remember to resolve the underlying issue and update the image reference.
+### Floating Tags
+
+Floating tags have no distro version indicated in the name and are scoped to a .NET version (e.g. `alpine-net9.0-helix-amd64`).
+It is routinely updated to reference a new version as the distro's and .NET's lifecycles progress.
+
+Floating tags are provided on an as-needed basis.
+They are beneficial for repos that are not susceptible to breaking changes that occur from new distro versions because the source that references the tag doesn't need to be updated in order to make use of the new version.
+
+#### Moving the Floating Tag
+
+There is a workflow that is following before a floating tag gets moved to a newer distro version:
+1. First, version-specific tags for the new distro version are provided.
+1. After a one month evaluation period, the floating tag is ready to reference the new version, assuming there are no issues found.
+1. The floating tag is moved to reference the new version on a C week (week containing the 3rd Tuesday of the month).
+
+#### Stability Period
+
+Floating tags are scoped to a specific .NET version.
+This ensures they are stable as the release moves to maintenance phase, not getting bumped to a newer distro version during that phase.
+This time period starts 6 months before the EOL date of the .NET version.
+In other words, for the last 6 months of servicing for that .NET version, the floating tag is guaranteed to not be moved to a new distro version.
+Repos consuming these tags should reference the .NET version associated with the release branch (e.g. sources in the `release/9.0` branch should reference the `net9.0` tag).
+
+### Image Pinning
+
+At times, it may be necessary to use a [pinned image reference](https://github.com/dotnet/runtime/pull/110199#discussion_r1859075989) for build reliability.
+This is done by appending the digest of the specific image that is needed (e.g. `mcr.microsoft.com/dotnet-buildtools/prereqs:<tag-name>@sha256:56feee03d202e008a98f3c92784f79f3f0b3a512074f7f8ee2b1ba4ca4c08c6e`).
+If the reference was pinned in response to a break that occurred, a tracking issue should be created (before the PR is merged) so that we remember to resolve the underlying issue and update the image reference back to the original value.
+
+### Example References
 
 The following locations are examples of infra that gets updated when new images are available.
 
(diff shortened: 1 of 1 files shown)


lifecycle.md
```diff
@@ -48,9 +48,24 @@ Note: this content should be moved to another location as it is not lifecycle re
 
 Build and test images are referenced in repo infra files, across a variety of `main` and `release/*` branches. Updating these references is a multi-step detail-oriented task. It is a pain, but necessary.
 
-We use version specific references in infra to ensure that our CI builds are reliable. One can imagine using floating OS tags (such as `debian-oldest` and `debian-latest`), however such an approach would be guaranteed to break our build. We know that since we often see build and test breaks that need addressing in PRs where we update build and test images.
+Several types of tag styles are provided:
```
MichaelSimons, line 51: Would these tags be created regardless if they are being used or would they be created on demand?  I question if they are going to be used for every distro/image variant.

lifecycle.md
```diff
@@ -48,9 +48,24 @@ Note: this content should be moved to another location as it is not lifecycle re
...
+  * oldest: References the oldest supported version of the distro. Example: `alpine-oldest-helix-amd64`
 
-At times, it may be necessary to use a [fixed image reference](https://github.com/dotnet/runtime/pull/110199#discussion_r1859075989) for build reliability. If this is ever done, a tracking issue should be created (before the PR is merged) so that we remember to resolve the underlying issue and update the image reference.
+FLoating tags are beneficial for repos that are not susceptible to breaking changes that occur from new distro versions because the source that references the tag doesn't need to be updated in order to make use of the new version.
+Conversely, some repos may be susceptible to distro breaking changes in which case the version-specific tags should be used ([dotnet/runtime](https://github.com/dotnet/runtime) is an example of such a repo).
+
+There is a vetting process before a `latest` floating tag gets moved to a newer distro version:
+
+1. Version-specific tags for the new distro version are provided.
+1. The [dotnet/runtime](https://github.com/dotnet/runtime) repo is updated to reference the new version-specific tags.
```
MichaelSimons, line 63: Aren't there going to be cases where images aren't referenced by the runtime?

I don't like this vetting process in a way.  If vetting is required you shouldn't be using the version specific tags.  I tend to learn towards the .NET Docker Alpine floating tag policy.  The new distro is made available one month, the next month the floating tag is updated.  This is a more scalable model.

lifecycle.md
richlander: I also lean towards on-demand. In fact, I'm not sure we want all distros being tested with floating tags.

lifecycle.md
richlander: That works as a policy. If we found an issue in runtime, we'd pause roll-out in dotnet-docker. We'd do the same with this system. It does mean that we'll have to be intentional w/ensuring we get coverage in runtime pretty quickly.

lifecycle.md
richlander: > Several types

Several of two?

lifecycle.md
```diff
@@ -48,20 +48,18 @@ Note: this content should be moved to another location as it is not lifecycle re
 
 Build and test images are referenced in repo infra files, across a variety of `main` and `release/*` branches. Updating these references is a multi-step detail-oriented task. It is a pain, but necessary.
 
-Several types of tag styles are provided:
+Two types of tag styles are available:
 * version-specific: The tag name includes the distro's version. Example: `alpine-3.21-helix-amd64`
-* floating: The tag name has no version. Instead it is routinely updated to reference a new version as the distro's lifecycle progresses. There are two sub-types of floating tags:
+* floating (optional): The tag name has no version. Instead it is routinely updated to reference a new version as the distro's lifecycle progresses. There are two sub-types of floating tags:
   * latest: References the latest validated version of the distro. Example: `alpine-latest-helix-amd64`
```
richlander, line 54: I'm concerned about this tags being unbounded. At some point in a product release, it is time to turn up the dial on stability and to not move OS versions forward. In a year, we may want to stop updating .NET tags for that reasons. Similarly, "oldest" for .NET 8 and .NET 10 do not match, now. I've already moved .NET 10 to use Alpine 3.21 in dotnet/runtime as "oldest".

I'm a big fan of the model we used for our build images. They are version-specific. I think we should use the same model for build and test images. That approach would be significantly simplifying for teams. Just one model to understand.

https://github.com/dotnet/runtime/blob/main/docs/workflow/using-docker.md#the-official-runtime-docker-images

I'm also not convinced that we need oldest and latest. I think we should just offer one image variant to teams, just like we do for build images. If they need more versions, they can go with DIY.

In dotnet/runtime, the CoreCLR build targets the trailing edge of OS versions we support. Libraries biases to newest.

I think the best model is to:

- Continue to use runtime to hold the baton on "oldest"
- Use libraries to validate that "newest" is OK.
- Push other repos (that use the floating tag) to newest after that, with a single tag (that doesn't say "newest").

Today, these tags are (effectively) exclusively referenced in our own infra. As we move to VMR/UB, there may be more users of these images. We need to be cognizant of that and ensure we don't break them when they attempt to build and test old .NET versions.

mthalman: > I'm a big fan of the model we used for our build images. They are version-specific. I think we should use the same model for build and test images.

That gets outside the scope of what is being addressed here. That's going to move us into producing new images, not just tagging changes. I would encourage you to create a separate issue for this.

> I think the best model is to:
> 
> * Continue to use runtime to hold the baton on "oldest"
> * Use libraries to validate that "newest" is OK.
> * Push other repos (that use the floating tag) to newest after that, with a single tag (that doesn't say "newest").

Let me check with you that I'm interpreting this correctly. This means we would have a floating tag that represents the latest version. If repos want an older version, they should use a version-specific tag. And libraries would still use a version-specific tag, targeting the latest, that we can use for vetting.

richlander: I don't see why we need new images. I'm only talking about tagging.

richlander: Let's make this concrete. We should model this for .NET 8 and 10 for Alpine. As I said, I'd like to retain the ability to stop taking Alpine changes for .NET 8 late in the .NET 8 lifecycle (last 6 months).

mthalman: > I don't see why we need new images. I'm only talking about tagging.

I thought you were referring to using net8.0 and net9.0 tagging like we do for the build images.

richlander: Yes. For our build images, it's completely different content. For this, I'm suggesting just different tags, all handled in `manifest.json`.

Imagine just before .NET 9 EOL:

- `alpine-net9.0` -- same as `alpine-3.23-helix-amd64`
- `alpine-net10.0` -- same as `alpine-3.24-helix-amd64`

richlander: Some thoughts on users of these tags:

- Should only update tags (from Alpine 3.23 to 3.24) in the middle weeks of the month. If Patch Tuesday is the B week, then we should update change these tags  on a C or D week, ideally also on a Tuesday. We don't want to disrupt servicing activities.
- If a change breaks one or a few teams, it is their responsibility to fix their code. That's part of the terms of use of these tags. If a change breaks most users, then the tag change should be reverted.
- The tags should only be updated after they test cleanly in runtime, otherwise, we're in YOLO mode.
- Not appropriate for repos that have `-staging` branches, like https://github.com/dotnet/runtime/tree/release/8.0-staging. That pattern places a high premium on a commit encoding/representing all aspects of code and config. Floating tags are antithetical to that approach.

MichaelSimons: > If Patch Tuesday is the B week, then we should update change these tags on a C or D week, ideally also on a Tuesday. We don't want to disrupt servicing activities.

Timing is a good point.  Wouldn't we want to make these prior to code complete?

richlander: Yes. Perhaps that's just the C week. I don't know what that timing is so really just bringing up the topic. I am certain the B week is a bad choice.

mthalman: Made updates in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1295/commits/cae40fca0eb118d24ee859c08ad5b5e4867d1b43. This gets rid of `oldest` tags, defines the C week policy, and .NET version scoping.

lifecycle.md
```diff
@@ -48,9 +48,44 @@ Note: this content should be moved to another location as it is not lifecycle re
...
+Distro version-specific tags include the distro's version (e.g. `alpine-3.21-helix-amd64`).
+Repos that are susceptible to breaking changes in the distro should use these tags.
+[dotnet/runtime](https://github.com/dotnet/runtime) is an example of such a repo.
+
+### Floating Tags
+
+Floating tags have no distro version indicated in the name and are scoped to a .NET version (e.g. `alpine-net9.0-helix-amd64`).
+It is routinely updated to reference a new version as the distro's and .NET's lifecycles progress.
+
+Floating tags are provided on an as-needed basis.
```
richlander, line 64: We could use some text in this section that we'll stop re-building these tags once the given release goes EOL.

lifecycle.md
```diff
@@ -48,9 +48,44 @@ Note: this content should be moved to another location as it is not lifecycle re
...
+#### Moving the Floating Tag
+
+There is a workflow that is following before a floating tag gets moved to a newer distro version:
+1. First, version-specific tags for the new distro version are provided.
+1. After a one month evaluation period, the floating tag is ready to reference the new version, assuming there are no issues found.
+1. The floating tag is moved to reference the new version on a C week (week containing the 3rd Tuesday of the month).
+
+#### Stability Period
+
+Floating tags are scoped to a specific .NET version.
```
richlander, line 76: Maybe add some text that says "after EOL, the version-specific tag is no longer maintained".

richlander:
LGTM. One remaining small comment. Thanks for doing this!

richlander (approved):

lifecycle.md
```diff
@@ -48,9 +48,44 @@ Note: this content should be moved to another location as it is not lifecycle re
...
+
+Floating tags have no distro version indicated in the name and are scoped to a .NET version (e.g. `alpine-net9.0-helix-amd64`).
+It is routinely updated to reference a new version as the distro's and .NET's lifecycles progress.
+
+Floating tags are provided on an as-needed basis.
+They are beneficial for repos that are not susceptible to breaking changes that occur from new distro versions because the source that references the tag doesn't need to be updated in order to make use of the new version.
+
+#### Moving the Floating Tag
+
+There is a workflow that is following before a floating tag gets moved to a newer distro version:
```
MichaelSimons, line 69:
```suggestion
There is a workflow that is followed before a floating tag gets moved to a newer distro version:
```

lifecycle.md
```diff
@@ -48,9 +48,44 @@ Note: this content should be moved to another location as it is not lifecycle re
...
+1. First, version-specific tags for the new distro version are provided.
+1. After a one month evaluation period, the floating tag is ready to reference the new version, assuming there are no issues found.
```
MichaelSimons, line 71: Would it be worthwhile to update the in-development version ahead of servicing versions of .NET?  e.g. updating the .NET 10 alpine image before updating the .NET 9.0 + 8.0 images.  This could help provide stability in the servicing versions.

lifecycle.md
MichaelSimons, line 64: Should this state what the process is if needed?  Self-service?  Who is responsible for onboarding the OS versions and moving the floating tag accordingly?

lifecycle.md
richlander: That is an interesting idea. That likely makes sense.

richlander: Can you write up a proposed timeline? I'm not following why we need two months but I might be missing something.

mthalman: It was based on Michael's suggestion:

> updating the .NET 10 alpine image before updating the .NET 9.0 + 8.0 images.

If not 2 months, then what is the timeframe between updating .NET 10 and the others?

richlander: Here is what I interpreted by the suggestion, for a timeline:

- July: Alpine 3.22 ships
- August:
  - .NET 10 Preview -- Switches to Alpine 3.22 as only option (for both version-specific and floating tags)
  - .NET 8/9 -- Both 3.21 and 3.22 are available as version-specific options; floating stays on 3.21
- September: 
  - .NET 8/9 -- Both 3.21 and 3.22 are available as version-specific options; floating stays is updated to 3.22 

Perhaps @MichaelSimons meant something different.

MichaelSimons: I was originally thinking a shorter timeline (2 week in main).  I don't see a particular driving reason for that.  The two month rollout it amenable to me.

mthalman: @richlander - Where does the runtime evaluation period fit into your timeline? Are you suggesting that floating tags for .NET 10 Preview get updated prior to a 1 month evaluation period in runtime? Or does that evaluation period start in July?

richlander: Good question. I forgot about that. Let's assume that July (in this example) is "runtime signoff".

lifecycle.md
mthalman: I think we want to have an issue filed if a floating tag is desired so that we can ensure it's appropriate.

Moving the floating tag should be done centrally and not delegated, I think, since there needs to be a close schedule followed. So that would be done by us, the maintainers.

richlander (approved):
LGTM -- Let's try this plan and adapt if needed.

MichaelSimons (approved):

PR #1298, opened 2024-12-12
Add CentOS Stream 10 test image

richlander:
Added x64 and Arm64 images.

I renamed the directories to something more matching our style.

- "centos" -> "centos-stream"
- "stream10" -> "10.0"

https://blog.centos.org/2024/12/introducing-centos-stream-10/


src/centos-stream/manifest.json
```diff
@@ -0,0 +1,72 @@
...
+        },
+        {
+          "platforms": [
+            {
+              "architecture": "amd64",
+              "dockerfile": "src/centos-stream/10.0/helix/",
+              "os": "linux",
+              "osVersion": "centos-stream10",
+              "tags": {
+                "centos-stream9-helix": {}
```
mthalman, line 50: Two things here.

First is that this should be 10, not 9. This also needs to be applied to the arm64 Dockerfile. 

Second is to add a separator for the version to follow the existing pattern of other distros. This will actually be necessary if we want to use Renovate to automatically update tag references; it needs a common pattern to recognize. Since you're in here, can you apply this pattern to the version 9 tags as well (by adding a new tag, not replacing the existing one)?

```suggestion
                "centos-stream-10-helix": {}
```

richlander: I would be happy to make that change. I think that's a good direction.

richlander:
Looks like we need a stream10 image pushed.

mthalman:
> Looks like we need a stream10 image pushed.

Fixed. Restarting build.

mthalman (approved):

richlander:
FYI @omajid

PR #1301, opened 2024-12-16
Azure Linux cross riscv464: Update debootstrap to 1.0.138

mthalman:
The previously referenced 1.0.137 version of debootstrap is no longer available from the package repository. Updating to 1.0.138 instead.

### DIFF
--- a/src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile
@@ -5,8 +5,8 @@ ARG ROOTFS_DIR
 
 # Install the latest debootstrap
 RUN tdnf remove -y debootstrap && \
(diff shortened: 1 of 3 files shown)


MichaelSimons (approved):

am11:
Any chance we get attention from Azure Linux team to update it upstream https://github.com/microsoft/azurelinux/pull/9422 so we don't have to keep maintaining the version? We only need 'a' version which recognizes new architectures runtime is supporting.

mthalman:
> Any chance we get attention from Azure Linux team to update it upstream [microsoft/azurelinux#9422](https://github.com/microsoft/azurelinux/pull/9422) so we don't have to keep maintaining the version? We only need 'a' version which recognizes new architectures runtime is supporting.

I followed up on that: https://github.com/microsoft/azurelinux/pull/9422#issuecomment-2548555339

PR #1309, opened 2025-01-03
Dummy change to retrigger the new arch build

am11:
https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1306#issuecomment-2569739177

cc @akoeplinger @jkoritzinsky

PR #1306, opened 2024-12-22
Add linux-loongarch64 images

am11:
Upstream https://github.com/dotnet/arcade/pull/15370

Fixes https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/856

PR #1313, opened 2025-01-06
Touch crossdeps-builder clone step

sbomer:
Test to see what gets rebuilt when touching the arcade clone step

### DIFF
--- a/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
@@ -81,4 +81,4 @@ RUN mkdir llvm-project.src && \
 
 # Obtain arcade scripts used to build rootfs
 RUN git config --global user.email builder@dotnet-buildtools-prereqs-docker && \
-    git clone --depth 1 --single-branch https://github.com/dotnet/arcade /scripts 
+    git clone --depth 1 --single-branch https://github.com/dotnet/arcade /scripts


am11:
When we only touch `crossdeps-builder` and the target cross dockerfiles, the change in llvm command does not reflect. e.g. when I added LoongArch in this file and touched cross/loongarch, it didn't located the new architecture. It required us to also touch `crossdeps-llvm`. Touching `crossdeps-llvm` builds everything under that netXX/cross.

Ideally, we should do this:

* Swap naming: crossdeps-llvm is more suitable name for crossdeps-builder; where we are actually building llvm, and vice-versa.
* Move arcade clone step to cross images.

am11:
> crossdeps-builder is a helper that contains build-time dependencies of the cross images that don't need to be part of the final images (llvm sources, llvm build dependencies, debootstrap, and the rootfs build scripts). crossdeps-llvm contains the runtime dependencies, including llvm, and serves as a base for the cross images - it's just the crossdeps image plus an install of llvm that we built in crossdeps-builder.

I think there is a misunderstanding. Naming seem inverted. If the one with "llvm" in its name is building the actual llvm toolchian, it will remedy the confusion. The cross dockerfiles can use the new name (by updating their first line).

> I'd like to keep the current architecture where one arcade clone is shared by all of the cross images, otherwise they could get out of sync and have different arcade versions. If we really need to work around the caching limitations, maybe we can separate the arcade clone into a separate image that is based on crossdeps-builder, so that the ci cache picks it up?

We can still achieve that by moving clone step to the middle layer (current: crossdeps-llvm, proposed: crossdeps-builder), then we don't need to rebuild llvm toolchain. It will rebuild the cross/ images only, which is better than touching multiple files.

> @mthalman is this expected? I would have thought that if I touch crossdeps-builder, images which depend on it (like crossdeps-llvm) would get rebuilt too. This seems like it might be a bug.

Yes, it does seem unintentional / bug. For clarity, the first PR was https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1306,, then we had to do another to rebuild the middle image https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1309 because it wasn't built.

Issue #1312, opened 2025-01-06
Layring for cross image and /scripts directory


am11:
Arcade is cloned in the layer where llvm is built `src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile`. If we make a change for one platform in rootfs script which resides in arcade repo, then we need to rebuild multiple layers in order to pick up that change:

* `src/azurelinux/3.0/net10.0/crossdeps-llvm/amd64/Dockerfile`
* `src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile`
* `src/azurelinux/3.0/net10.0/cross/<that platform>`

this way we end up rebuilding a lot of unrelated stuff and waste many hours. We should move `git clone` step to cross dockerfiles because that's where we use the scripts.

cc @sbomer

mthalman:
> but lacking that we may be able to fix it by moving the arcade clone to a separate image.

If the Dockerfile of this image was to be left to clone as it does today, we would never rebuild the image since the Dockerfile won't have been changed. So the image would get stale over time. I think we'd want to check out a specific commit of the arcade repo instead. But of course then we would want to have a system in place for updating that commit.

am11:
My observation is:

* Touching crossdeps-llvm rebuilds all cross images.
* Touching crossdeps-builder does not rebuild crossdeps-llvm.

If you observe the same then there is more to it.

am11:
My observation is:

* Touching crossdeps-llvm rebuilds all cross images.
* Touching crossdeps-builder does not rebuild crossdeps-llvm.

If you've observed the same then there is more to it.

mthalman:
[Triage]
@am11 - Do you want to try to make this change?

PR #1318, opened 2025-01-11
Update llvm from 19.1.0 to 19.1.6

am11:


src/azurelinux/3.0/net10.0/cross/loongarch64/Dockerfile
```diff
@@ -3,9 +3,6 @@ ARG ROOTFS_DIR=/crossrootfs/loongarch64
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net10.0-crossdeps-builder-amd64 AS builder
 ARG ROOTFS_DIR
 
-# temporarily bypass cache to fetch the latest arcade version with recent fixes
-RUN rm -rf /scripts && git clone --depth 1 --single-branch https://github.com/dotnet/arcade /scripts
```
am11, line 7: `src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile` is rebuilding which will pull the new arcade. Rebuilding all cross/ images with llvm requires touching `src/azurelinux/3.0/net10.0/crossdeps-llvm/amd64/Dockerfile`, see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1312.

am11:
cc @akoeplinger

I think we should remove "android" from the CI leg name

>  dotnet-buildtools-prereqs-docker-all (Build Linux_amd64 src-azurelinux-3.0-net10.0-android-graph) In progress — This check has started... 

because it covers all platforms, not just Android.

akoeplinger (approved):

akoeplinger:
> I think we should remove "android" from the CI leg name

Sure, sounds good.

am11:
> > I think we should remove "android" from the CI leg name
> 
> Sure, sounds good.

I couldn't figure out where is it coming from. It's some dynamic magic. 😆

akoeplinger:
ah yeah, it gets generated by `Generate platformDependencyGraph Matrix`: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/efebaec78e371412e8537ec7e584e5c7347348ef/eng/common/templates/jobs/generate-matrix.yml#L58

that's some darc magic indeed :D

am11:
Looks like it's coming from https://github.com/dotnet/docker-tools/blob/d455b61a1afaacf125af556e81f69d16d39072f4/src/Microsoft.DotNet.ImageBuilder/src/Commands/GenerateBuildMatrixCommand.cs#L385. The `.First()` path it sees is probably `src/azurelinux/3.0/net10.0/android/amd64/Dockerfile` and if `matrixNameParts` is `["amd64"]` to be excluded, we are seeing the final result as `src-azurelinux-3.0-net10.0-android-graph`.

mthalman:
Yeah, there's not a way to explicitly set the build leg name. As you've already seen, it's dynamic based on how the computed build graph gets composed.

Issue #1333, opened 2025-01-27
The helixbot user does not need to be set as the default user


mthalman:
The Helix Dockerfiles should be updated to not set `helixbot` as the default user. They should be left with `root` as the default. This works because the Helix infrastructure will explicitly set `helixbot` as the user when running the container. And having it set to `root` by default will enable the container image to be used as a container job (see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1276#issuecomment-2605791196).

PR #1334, opened 2025-01-27
Don't set helixbot as the default user

mthalman:
Contributes to https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1333

I've already done some validation that removing the `helixbot` user as the default user works in a test build. But I want to roll this out as a "pilot" image first to validate within the runtime repo. If this is successful, then I'll roll this change out to the rest of the Dockerfiles.

### DIFF
--- a/src/fedora/41/helix/Dockerfile
+++ b/src/fedora/41/helix/Dockerfile
@@ -65,8 +65,6 @@ RUN /usr/sbin/adduser --uid 1000 --shell /bin/bash --group adm helixbot \
     && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers \
     && chmod +s /usr/bin/ping
 
-USER helixbot
-
 # Install Helix Dependencies
 ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
 RUN python3 -m venv $VIRTUAL_ENV


mthalman:
> won't this break any `python3` use in build jobs❓ I realize running as `root` allows overwriting and adding files to the `venv`. however recent Python versions fail when attempting to update an environment the user doesn't own

Isn't that handled by this line, which sets helixbot as the owner? https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1334/files#diff-34a803fd425a619ab16b0a1d4251553be842e640e0a9c84eb5c324445a014346L74

dougbu:
> > won't this break any `python3` use in build jobs❓ I realize running as `root` allows overwriting and adding files to the `venv`. however recent Python versions fail when attempting to update an environment the user doesn't own
> 
> Isn't that handled by this line, which sets helixbot as the owner? https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1334/files#diff-34a803fd425a619ab16b0a1d4251553be842e640e0a9c84eb5c324445a014346L74

only for the `helixbot` user. my question was about other users in _build_ scenarios

mthalman:
Oh, you're asking about the feasibility of using this for container jobs that do testing. Not related to the actual removal of helixbot as the default. Is that correct?

PR #1336, opened 2025-01-31
Remove helix script install (partial set of Dockerfiles)

mthalman:
Removing the Helix Python script install from a handful of Dockerfiles. I want to roll this out to just these Dockerfiles for now to evaluate that things are working. I have run a test build with one of the images but I want to ensure things work in the wild across these platforms.

Contributes to #1333

### DIFF
--- a/src/almalinux/8/helix/amd64/Dockerfile
+++ b/src/almalinux/8/helix/amd64/Dockerfile
@@ -21,9 +21,6 @@ RUN dnf upgrade --refresh -y \
         file \
     && dnf clean all
 
-RUN python3 -m pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple \
-    && python3 -m pip install ./helix_scripts-*-py3-none-any.whl
-
 # Aspnetcore test runs expect python to be available without version suffix
 RUN alternatives --set python /usr/bin/python3 \
     && ln -sf /usr/bin/pip3 /usr/bin/pip
@@ -34,7 +31,3 @@ ENV LANG=en-US.UTF-8
 RUN adduser --uid 1000 --shell /bin/bash --gid adm helixbot \
     && chmod 755 /root \
     && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers
-
-USER helixbot
-
-RUN python3 -m venv /home/helixbot/.vsts-env
--- a/src/almalinux/8/helix/amd64/Dockerfile
+++ b/src/almalinux/8/helix/amd64/Dockerfile
@@ -32,5 +32,7 @@ RUN adduser --uid 1000 --shell /bin/bash --gid adm helixbot \
 
 # Setup helix virtual environment
 ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
+USER helixbot
 RUN python3 -m venv $VIRTUAL_ENV
+USER root
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
--- a/src/almalinux/8/helix/amd64/Dockerfile
+++ b/src/almalinux/8/helix/amd64/Dockerfile
@@ -8,9 +8,7 @@ RUN dnf upgrade --refresh -y \
     #  libmsquic except in the 8.1-specific packages feed.)
     && dnf config-manager --add-repo=https://packages.microsoft.com/centos/8/prod/config.repo \
     && dnf install --setopt tsflags=nodocs -y \
-        # Get recent python3 This is needed to get a pip recent enough
-        # not to fail the build when installing cryptography library as
-        # a dependency of the helix scripts.
+        # Install supported version of Python (default for python3 is 3.6 which is unsupported)
         python39 \
         # Required for asp.net core test runs
         sudo \
--- a/src/almalinux/8/helix/amd64/Dockerfile
+++ b/src/almalinux/8/helix/amd64/Dockerfile
@@ -31,3 +31,8 @@ ENV LANG=en-US.UTF-8
 RUN adduser --uid 1000 --shell /bin/bash --gid adm helixbot \
     && chmod 755 /root \
     && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers
+
+# Setup helix virtual environment
+ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
--- a/src/fedora/41/helix/Dockerfile
+++ b/src/fedora/41/helix/Dockerfile
@@ -1,23 +1,3 @@
-FROM library/fedora:41 AS venv
-
-RUN dnf upgrade --refresh -y \
-    && dnf install --setopt tsflags=nodocs -y \
-        dnf-plugins-core \
-    && dnf install --setopt=install_weak_deps=False --setopt tsflags=nodocs -y \
-        openssl \
-        python3-devel \
-        python3-pip \
-        gcc \
-        libatomic \
-        redhat-rpm-config \
-    && dnf clean all
-
-RUN python3 -m venv /venv \
-    && source /venv/bin/activate \
-    pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple \
-    && pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple \
-    && pip install ./helix_scripts-*-py3-none-any.whl
-
 FROM library/fedora:41
 
 # Install Dependencies
@@ -64,9 +44,3 @@ RUN /usr/sbin/adduser --uid 1000 --shell /bin/bash --group adm helixbot \
     && chmod 755 /root \
     && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers \
     && chmod +s /usr/bin/ping
-
-# Install Helix Dependencies
-ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
-RUN python3 -m venv $VIRTUAL_ENV
-ENV PATH="$VIRTUAL_ENV/bin:$PATH"
-COPY --from=venv --chown=helixbot /venv $VIRTUAL_ENV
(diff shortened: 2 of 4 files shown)


src/almalinux/8/helix/amd64/Dockerfile
```diff
@@ -21,9 +21,6 @@ RUN dnf upgrade --refresh -y \
         file \
     && dnf clean all
 
-RUN python3 -m pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple \
```
MichaelSimons, line 24: Are there any prereqs that aren't needed since the script is no longer run?  For example, there is this comment - https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1336/files#diff-f8d93a264641a0bbae6526f0d72e990bfb0bd1e61b581daa07cf4cd5a84a7371R11

MichaelSimons:
@dougbu - It would be good for you to review these changes.

src/almalinux/8/helix/amd64/Dockerfile
```diff
@@ -34,7 +31,3 @@ ENV LANG=en-US.UTF-8
 RUN adduser --uid 1000 --shell /bin/bash --gid adm helixbot \
     && chmod 755 /root \
     && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers
-
-USER helixbot
-
-RUN python3 -m venv /home/helixbot/.vsts-env
```
dougbu, line 40: removing the helix-scripts packages should be fine b/c we mount the host's installation location when starting Helix containers. however it seems like a **Bad Idea :tm:** since a `venv` isolates test/user code from the system Python environment

src/fedora/41/helix/Dockerfile
```diff
@@ -64,9 +44,3 @@ RUN /usr/sbin/adduser --uid 1000 --shell /bin/bash --group adm helixbot \
     && chmod 755 /root \
     && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers \
     && chmod +s /usr/bin/ping
-
-# Install Helix Dependencies
-ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
-RUN python3 -m venv $VIRTUAL_ENV
-ENV PATH="$VIRTUAL_ENV/bin:$PATH"
-COPY --from=venv --chown=helixbot /venv $VIRTUAL_ENV
```
dougbu, line 72: put another way, I'd remove this line but leave the rest of the `venv` setup alone. bonus points if you use the more normal `VIRTUAL_ENV_DISABLE_PROMPT=true . /etc/helix-prep/venv/bin/activate` pattern

mthalman: There's no point in running `activate` in the Dockerfile if you're not running any subsequent commands to be run within that environment. The context of the activated environment is the shell session and the shell session ends after execution of the `RUN` instruction. So when running the container, that activated environment no longer exists.

dougbu: > There's no point in running `activate` in the Dockerfile if you're not running any subsequent commands to be run within that environment. The context of the activated environment is the shell session and the shell session ends after execution of the `RUN` instruction. So when running the container, that activated environment no longer exists.

do the `ENV` statements persist❓ could you please restore the `venv` regardless❓

src/almalinux/8/helix/amd64/Dockerfile
mthalman: Python still needs to be installed. I've updated the comment to be more appropriate for why we're installing that specific package.

mthalman:
@MichaelSimons, @dougbu - This is ready for another review.

dougbu:
build is failing still. could you let us know when that's cleared up❓

MichaelSimons:
> build is failing still. could you let us know when that's cleared up❓

What build are you referring to?

src/windowsservercore/ltsc2019/helix/amd64/Dockerfile
```diff
@@ -1,22 +1,28 @@
...
-    && md C:\Python C:\PythonTemp `
-    && tar -zxf %TEMP%\python.zip -C C:\PythonTemp `
-    && xcopy /s c:\PythonTemp\tools C:\Python `
-    && rd /s /q c:\PythonTemp `
-    && del /q %TEMP%\python.zip `
-    && setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;"
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
+# Install latest stable version of Python
+RUN $apiUrl = 'https://api.nuget.org/v3-flatcontainer/python/index.json'; `
```
MichaelSimons, line 9: It does feel a little strange the the SHELL is set to powershell here yet later on a direct powershell command is executed directly.  It would be nice to have consistency within this Dockerfile.

dougbu:
> > build is failing still. could you let us know when that's cleared up❓
> 
> What build are you referring to?

I saw failures in a few of the jobs, including a required one, when looking at the list of checks. resolved now &mdash; either someone retried the jobs or my window was out of date

src/almalinux/8/helix/amd64/Dockerfile
dougbu: /fyi this requirement for Helix containers won't go away until and unless we retire the Helix Python code in dotnet/arcade. the Helix SDK regularly adds post-commands relying on Python

src/windowsservercore/ltsc2019/helix/amd64/Dockerfile
```diff
@@ -1,29 +1,30 @@
...
+        Invoke-WebRequest -Uri https://www.nuget.org/api/v2/package/python/$latestVersion -OutFile $env:TEMP\python.zip; `
+        md C:\Python; `
+        md C:\PythonTemp; `
+        tar -zxf  $env:TEMP\python.zip -C C:\PythonTemp; `
+        xcopy /s c:\PythonTemp\tools C:\Python; `
+        Remove-Item -Recurse -Force C:\PythonTemp; `
+        Remove-Item -Force $env:TEMP\python.zip;"
 
-SHELL ["cmd", "/S", "/C"]
+ENV PATH="$PATH;C:\Python;C:\python\scripts" `
```
MichaelSimons, line 21: Does this work?  This used to never be supported.  IIRC you have to use setx to set the path in Windows Dockerfiles.

mthalman: Yes, it works. Otherwise, the build would fail due to python not being available from the PATH. 😄 But yes, I recall this being an issue in the past but more recent Docker versions seem to have resolved this.

mthalman: Oh, I see. There's an issue in being able to reference another variable from this instruction.

```Dockerfile
ENV PATH="$PATH;C:\Python;C:\python\scripts"
```

Gives you a `PATH` of `;C:\Python;C:\python\scripts`. If you change `$PATH` to `%PATH%`, you get a literal value of `%PATH%;C:\Python;C:\python\scripts`.

Using `setx` solves that.

src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -10,30 +10,30 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
...
-    pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && `
-    for %f in (.\helix_scripts-*-py3-none-any.whl) do (pip install %f) && `
+COPY --from=installer [ "C:\\PythonTemp\\tools", "C:\\Python\\" ]
+
+ENV PATH="$PATH;C:\Program Files\PowerShell\;C:\Python;C:\python\scripts" `
+    PYTHONPATH="C:\Python\Lib;C:\Python\DLLs;"
+
+RUN md c:\\helixtmp && pushd c:\\helixtmp && `
+    python -m pip install --upgrade pip && `
+    python -m pip install virtualenv && `
```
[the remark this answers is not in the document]
dougbu: ❓

dougbu (approved):
I continue to think `virtualenv` is the wrong approach in the Windows containers but am guessing that should be considered separate from this PR 😦  is there an issue open to use a consistent environment approach❓

src/nanoserver/1809/helix/amd64/Dockerfile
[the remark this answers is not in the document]
mthalman: Can you provide a suggestion of what you expect? This line is just installing the `virtualenv` package. What should it be doing instead?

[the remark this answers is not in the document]
dougbu: install `venv`

at this point, this is for user/test code and not the Helix code. (there isn't really a system environment on Windows that we need to worry about isolating ourselves from.) mostly a consistency point though `virtualenv` is pretty strongly not recommended these days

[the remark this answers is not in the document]
mthalman: Changing it to `python -m pip install venv` yields this error:

```
ERROR: Could not find a version that satisfies the requirement venv (from versions: none)
ERROR: No matching distribution found for venv
```

mthalman:
I'll wait to merge this until next week when I'll have more availability to monitor the usage of these updates.

dougbu (approved):
looking significantly better 😀

src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -10,31 +10,31 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
     Expand-Archive -Path C:\powershell.zip -DestinationPath C:\powershell; `
     Remove-Item -Path C:\powershell.zip
 
+# Download latest stable version of Python
```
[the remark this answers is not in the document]
mthalman: It's being installed in the 2nd stage: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1336/files#diff-254c9f3bc219087b528b7261fbf8ef179781e80eca82ce301d66e492a0917aa2R28. That's needed because Nano Server doesn't have PowerShell like Windows Server Core does.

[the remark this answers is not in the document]
mthalman: That stage runs on Windows Server Core so PowerShell exists. The `SHELL` instruction tells Docker to use `powershell` as its shell when executing any subsequent `RUN` instructions.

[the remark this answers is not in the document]
dougbu: ah, got it. sorry it took me awhile

mthalman:
I'm going to split these changes into separate PRs to have more focused issues being addressed.

Issue #1331, opened 2025-01-24
Be more selective what gets copied from cross builder stage?


mthalman:
There's an issue with an outdated Python package, Pygments, that is included in the cross container image and causing vulnerabilities: https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/164

Example image:

`mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net10.0-cross-armv6@sha256:9060cc2e7080451bc18aca2c9591d6c3bd8d9a222f83ef46cbce6cb15b7d0d60`

has path: `/crossrootfs/armv6/usr/lib/python3/dist-packages/Pygments-2.14.0.egg-info`

Are these packages even needed in the final images? This one ends up getting installed as a result of this line: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/2ab7ab12be6e15dcaa81af4843ae4240301dfa7e/src/azurelinux/3.0/net10.0/cross/armv6/Dockerfile#L13

If Python packages are not needed in the final image, can we clean things up before they get copied to the final image here: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/2ab7ab12be6e15dcaa81af4843ae4240301dfa7e/src/azurelinux/3.0/net10.0/cross/armv6/Dockerfile#L18

cc @sbomer

mthalman:
[Triage]

@sbomer - Can you take a look at this and provide some feedback on if all of these Python packages are needed in the final image? It would be nice if we could exclude them if they're unnecessary.

PR #1339, opened 2025-02-04
Remove lldb from armv6 images

sbomer:
I think this should help with https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1331 - it looks like lldb is what was pulling in the Pygments python package. We don't need lldb on these images because armv6 isn't officially supported.


jkoritzinsky (approved):

am11:
> because armv6 isn't officially supported.

You mean armv6 is not officially supported by libSOS or it's not official platform and we can break it? If it's latter then I don't understand the point of this PR.

akoeplinger:
iirc only Mono "supported" armv6 and libSOS doesn't apply there

am11:
Only reason we install lldb package in rootfs is for building lldb plugin in SOS project. If it’s unsupported then it make sense to not install it (why did we even bother installing it in the first place)

akoeplinger:
that one isn't for armv6 but from what I can see nobody should be using that image anymore

mthalman:
> that one isn't for armv6 but from what I can see nobody should be using that image anymore

You're right! Our usage report shows no hits on that. Nice. I will delete it.

mthalman:
This is why we need https://github.com/dotnet/dnceng/issues/4360

PR #1346, opened 2025-02-06
Add build-deps image

richlander:
Related: https://github.com/dotnet/aspnetcore/issues/60208

The intent is to provide an image that you can install an arbitrary SDK on, likely from ci.dot.net, and that has Node.js installed. This should be usable by most C# only repos.

I'm wondering if we should adopt a new naming scheme:

- `build-deps`
- `test-deps` (replacing Helix)

Runtime has their own [crazy naming scheme](https://github.com/dotnet/runtime/blob/main/docs/workflow/using-docker.md). No need to reason about that.


src/azurelinux/manifest.json
```diff
@@ -943,8 +941,21 @@
...
+        },
+        {
+          "platforms": [
+            {
+              "architecture": "amd64",
+              "dockerfile": "src/azurelinux/3.0/build-deps",
+              "os": "linux",
+              "osVersion": "azurelinux3.0",
+              "tags": {
+                "azurelinux-3.0-build-deps-amd64": {}
```
mthalman, line 953: I do think the tag and dockerfile path should reflect the .NET version. There may be changes in runtime-deps that would break the usage of an older .NET version.

```suggestion
              "dockerfile": "src/azurelinux/3.0/net10.0/build-deps",
              "os": "linux",
              "osVersion": "azurelinux3.0",
              "tags": {
                "azurelinux-3.0-net10.0-build-deps-amd64": {}
```

mthalman, line 953: I'm not sure the `deps` suffix really adds anything. Every image in this repo is kind of a `deps` image already.

richlander: That's true. I was motivate to ensure people realized there was no SDK inside. However, I have low care factor on that.

richlander: I changed it.

richlander:
Can you help me with codeowners @mthalman? I'm not sure what to do for that.

mthalman:
> Can you help me with codeowners @mthalman? I'm not sure what to do for that.

Fixed in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1346/commits/e0ef84ee47e6f98f5ba3df4520db571aebb84e51

lbussell (approved):

PR #1353, opened 2025-02-08
Add tar to Azure Linux build image

richlander:
- Discovered we also need `tar`.
- Added `curl` as well to ensure we get the latest patches for it.

### DIFF
--- a/src/azurelinux/3.0/net10.0/build/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/build/Dockerfile
@@ -3,9 +3,11 @@ FROM mcr.microsoft.com/dotnet/nightly/runtime-deps:10.0-preview-azurelinux3.0
 
 RUN tdnf install -y \
         awk \
+        curl \
         nodejs \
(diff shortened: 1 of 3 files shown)


hoyosjs (approved):

mconnew (approved):

src/azurelinux/3.0/net10.0/build/Dockerfile
```diff
@@ -3,9 +3,11 @@ FROM mcr.microsoft.com/dotnet/nightly/runtime-deps:10.0-preview-azurelinux3.0
 
 RUN tdnf install -y \
         awk \
+        curl \
         nodejs \
         # provides 'useradd', required by Azure DevOps
         shadow-utils \
         # Provides 'su', required by Azure DevOps
+        tar \
```
akoeplinger, line 11: nit: move this up so the comment still refers to util-linux

richlander: Good catch. Fixed.

PR #1355, opened 2025-02-10
Add dependencies to Azure Linux build image

richlander:


src/azurelinux/3.0/net10.0/build/Dockerfile
```diff
@@ -2,10 +2,13 @@
 FROM mcr.microsoft.com/dotnet/nightly/runtime-deps:10.0-preview-azurelinux3.0
 
 RUN tdnf install -y \
+        azure-cli \
```
mthalman, line 5: Should be placed after `awk`.

mthalman (approved):

PR #1357, opened 2025-02-11
Add CentOS Stream 10 build image

MichaelSimons:
Related to https://github.com/dotnet/source-build/issues/4890

There following changes were made vs the 9 version.

1. NodeJS module not installed - the NodeJS module is no longer available in the CentOS Stream 10 - AppStream  package feed.  If any repo's rely on this, they will need to decide where to install it from.
2. Azure CLI not installed - Azure has not published 10 packages yet.  The version from the 9 feed will not install because of missing requirements.  This will have to be added later when available.  
3. sudo was dropped - sudo is not recommended in container scenarios.
4. ENV OPENSSL_ENABLE_SHA1_SIGNATURES=1 was dropped.  This is an attempt to cleanup.  Not sure if this is still required or not.

### DIFF
--- a/src/centos-stream/10.0/amd64/Dockerfile
+++ b/src/centos-stream/10.0/amd64/Dockerfile
@@ -51,6 +51,7 @@ RUN dnf upgrade --refresh -y \
         python3 \
         python3-devel \
         readline-devel \
+        sudo \
         swig \
         tar \
         wget \
@@ -58,5 +59,3 @@ RUN dnf upgrade --refresh -y \
         xz \
         zlib-devel \
     && dnf clean all
-
-ENV NO_UPDATE_NOTIFIER=true
--- a/src/centos-stream/10.0/amd64/Dockerfile
+++ b/src/centos-stream/10.0/amd64/Dockerfile
@@ -1,0 +1,62 @@
+FROM quay.io/centos/centos:stream10
+
+# Install dependencies
+RUN dnf upgrade --refresh -y \
+    && dnf install --setopt tsflags=nodocs -y \
+        dnf-plugins-core \
+    # rhel/10 packages don't exist yet, so we use the rhel/9 packages
+    && dnf config-manager --add-repo=https://packages.microsoft.com/rhel/9/prod/config.repo \
+    && dnf config-manager --set-enabled crb \
+    && dnf install --setopt tsflags=nodocs -y \
+        "perl(Time::HiRes)" \
+        autoconf \
+        automake \
+        brotli-devel \
+        clang \
+        cmake \
+        cpio \
+        curl-devel \
+        doxygen \
+        elfutils \
+        file \
+        findutils \
+        gcc \
+        gdb \
+        git \
+        glibc-langpack-en \
+        hostname \
+        iproute \
+        jq \
+        krb5-devel \
+        libcurl-devel \
+        libedit-devel \
+        libicu-devel \
+        libidn2-devel \
+        libnghttp2-devel \
+        libtool \
+        libuuid-devel \
+        libxml2-devel \
+        lld \
+        lldb-devel \
+        llvm \
+        lttng-ust-devel \
+        lzma \
+        make \
+        ncurses-devel \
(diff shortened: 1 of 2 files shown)


src/centos-stream/10.0/amd64/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+        readline-devel \
+        swig \
+        tar \
+        wget \
+        which \
+        xz \
+        zlib-devel \
+    && dnf clean all
+
+ENV NO_UPDATE_NOTIFIER=true
```
mthalman, line 62: I think we can get rid of this for now. This was originally added in order to build Node sources for source build. Certainly not needed now/yet since there's no Node in here.

src/centos-stream/10.0/amd64/Dockerfile
```diff
@@ -0,0 +1,62 @@
+FROM quay.io/centos/centos:stream10
+
+# Install dependencies
+RUN dnf upgrade --refresh -y \
+    && dnf install --setopt tsflags=nodocs -y \
+        dnf-plugins-core \
+    # rhel/10 packages don't exist yet, so we use the rhel/9 packages
+    && dnf config-manager --add-repo=https://packages.microsoft.com/rhel/9/prod/config.repo \
```
mthalman, line 8: This must be needed for `powershell`?

src/centos-stream/manifest.json
```diff
@@ -42,6 +42,18 @@
...
           ]
         },
+        {
+          "platforms": [
+            {
+              "dockerfile": "src/centos-stream/10.0/amd64",
+              "os": "linux",
+              "osVersion": "centos-stream10",
+              "tags": {
+                "centos-stream-10": {}
```
mthalman, line 52: The [updated tagging policy](https://github.com/dotnet/dotnet-buildtools-prereqs-docker?tab=readme-ov-file#how-to-identify-an-image) says that tag names are to explicitly define the architecture.

ellahathaway:
> NodeJS module not installed - the NodeJS module is no longer available in the CentOS Stream 10 - AppStream package feed. If any repo's rely on this, they will need to decide where to install it from.

Note that an Azure Pipelines containerization prerequisite is as follows: `Can run Node.js, which the agent provides. Node.js must be pre-installed for Linux containers on Windows hosts.`

So as long as we run containerized centos builds on an agent which is non-Windows and has Node.js, I think dropping node will not impact our VMR centos container builds.

MichaelSimons:
> Note that an Azure Pipelines containerization prerequisite is as follows: `Can run Node.js, which the agent provides. Node.js must be pre-installed for Linux containers on Windows hosts.`
> 
> So as long as we run containerized centos builds on an agent which is non-Windows and has Node.js, I think dropping node will not impact our VMR centos container builds.

We run on Linux build agents so we should be good here.  I will find out shortly... 😃

src/centos-stream/manifest.json
MichaelSimons: Fixed with 8c562c1c1762105e67eef14582fe05631b9e19ac

src/centos-stream/10.0/amd64/Dockerfile
MichaelSimons: Removed with 8c562c1c1762105e67eef14582fe05631b9e19ac

MichaelSimons:
> > sudo was dropped - sudo is not recommended in container scenarios.
> 
> This will negatively impact offline builds... `sudo` is required to unshare the network: https://github.com/dotnet/sdk/blob/62d8a6a7b332a1bf9591c3494edbcb48175610b5/eng/pipelines/templates/jobs/vmr-build.yml#L460

Good catch - Added back with 8c562c1c1762105e67eef14582fe05631b9e19ac.

mthalman (approved):

ellahathaway (approved):

MichaelSimons:
Windows PR validation failure is un-related.  Merging.

PR #1361, opened 2025-02-16
Remove debootstrap dependency from riscv64 images

am11:
This bring riscv64 to the same plan as loongarch64, which uses new mechanism to create rootfs without debootstrap/qemu requirement.

### DIFF
--- a/src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile
@@ -3,15 +3,10 @@ ARG ROOTFS_DIR=/crossrootfs/riscv64
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net10.0-crossdeps-builder-amd64 AS builder
 ARG ROOTFS_DIR
 
-# Install the latest debootstrap
-RUN tdnf remove -y debootstrap && \
-    curl -fsL -o debootstrap.tar.gz https://deb.debian.org/debian/pool/main/d/debootstrap/debootstrap_1.0.140.tar.gz && \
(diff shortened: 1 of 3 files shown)


am11:
@mthalman, with this, we don't need to keep debootstrap up to date any longer, because it's not involved at all.
I tested runtime build locally. Besides, loongarch64 and riscv64 are community platforms using the same build validation pipeline in runtime: https://github.com/dotnet/runtime/blob/8a79637e1454c751c24a984b7fb8af4c4d43394b/eng/pipelines/runtime.yml#L488-L489

jkotas:
> we don't need to keep debootstrap up to date any longer, because it's not involved at all.

Should this be deleted too? https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/1746a09f4ec139b4844413033d7d3cd18614860e/src/azurelinux/3.0/net9.0/crossdeps-builder/amd64/Dockerfile#L32C49-L32C60

am11:
Debootstrap is still used by officially supported linux architectures. The key is needed for package installation (`apt install` step) which happens inside chroot environment running on qemu during build-rootfs script: https://github.com/dotnet/arcade/blob/ae3c938af5df4a2db45ded37dd05608ce59d8e5e/eng/common/cross/build-rootfs.sh#L844

@akoeplinger is adding support for signed packages in new (`--no-emulation`) script https://github.com/dotnet/arcade/pull/15374, so that key will be required there as well. Once signed packages are supported, we can switch official archs to this --no-emulation plan as well and delete things like this script https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/1746a09f4ec139b4844413033d7d3cd18614860e/src/ubuntu/build-scripts/install-cross-build-prereqs.sh (which installs qemu in host layer, that is, debootstrap requirement).

akoeplinger (approved):

PR #1369, opened 2025-02-21
Update to clang 20.1.0

sbomer:
See https://github.com/dotnet/runtime/issues/109939#issuecomment-2651317376. The ppc64le build break has been fixed: https://github.com/dotnet/runtime/pull/112777.

### DIFF
--- a/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
@@ -56,11 +56,11 @@ RUN wget https://releases.llvm.org/release-keys.asc && \
     gpg --import release-keys.asc && \
     rm release-keys.asc && \
 # 2. Download llvm sources and signature, and verify signature
-    LLVM_VERSION=20.1.0-rc2 && \
+    LLVM_VERSION=20.1.0 && \
     wget -O llvm-project.src.tar.xz.sig https://github.com/llvm/llvm-project/releases/download/llvmorg-${LLVM_VERSION}/llvm-project-${LLVM_VERSION}.src.tar.xz.sig && \
-    echo "1862ff5a20a4ec884bfec14dab5c84c766b0b894a9a826978d4f1a52863d5a3c llvm-project.src.tar.xz.sig" | sha256sum -c && \
+    echo "d59a826bd924580c4f186ec2d88774c7ca4ec78bfb6593f68793696f32d09ebb llvm-project.src.tar.xz.sig" | sha256sum -c && \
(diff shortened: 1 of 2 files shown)


jkoritzinsky (approved):

jkoritzinsky:
We should update the version in the optimization images as well.

richlander:
I'm assuming we will not merge this change until we do a test run. We need confidence that we will not break CI, particularly since one break has already been found.

@jkotas

am11:
In the past, we have waited for the official release (and sometimes for update 2; which gives them enough time to fix major issues with the release). llvm 20.0 is scheduled for Mar 11th and 20.1 two weeks after that. Can this not wait?

richlander:
We could wait. What concerns do you have about "major issues"? Do you think the issues might block us shipping a preview?

am11:
> We could wait. What concerns do you have about "major issues"? Do you think the issues might block us shipping a preview?

In the first few weeks after the release, they usually prioritize fixing and backporting bugfixes discovered by large consumer base (most folks do not test the release candidate branches). Empirically speaking, the .1, .2 llvm-toolchain releases are usually more stable than RTM (just like .NET).

https://github.com/llvm/llvm-project/issues?q=is:issue%20is:closed%20label:regression:19

> It's also a question of whether we want to help finding those issues. I think that would be useful to the community.

I think it is a good plan. We can update it during RC period each year.

src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
```diff
@@ -56,11 +56,11 @@ RUN wget https://releases.llvm.org/release-keys.asc && \
     gpg --import release-keys.asc && \
     rm release-keys.asc && \
 # 2. Download llvm sources and signature, and verify signature
-    LLVM_VERSION=19.1.6 && \
+    LLVM_VERSION=20.1.0-rc2 && \
```
am11, line 59: Why not use https://github.com/llvm/llvm-project/releases/tag/llvmorg-20.1.0 now that it's already released?

richlander: Good catch @am11!

richlander (approved):

richlander:
Focusing on RC was an elaborate ploy to get us to adopt the GA release week-of!

richlander:
Thanks @sbomer!

am11:
BTW, new images are not built because the heuristics aren't perfect. We would need to touch `crossdeps-llvm` etc. or someone will need to manually evict the cache in ACR https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1312.

mthalman:
I'll kick the build to disable caching.

PR #1374, opened 2025-03-02
Improve Linux Helix containers usability

dougbu:
- give `helixbot` user UID 1000 in Ubuntu 24.04 and 24.10 images
  - delete the `ubuntu` user to free up that UID
- always provide and activate a `venv` for `helixbot` user
- always install `helix-scripts` module and dependencies in `venv`
  - some images installed this module in the system Python environment
  - where possible, do this installation as `helixbot` user
- upgrade some Python OS package versions
  - avoid some older versions e.g. `python3.8`
  - link new version to /usr/bin/python
- upgrade `pip` Python package in some images
  - work around https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1003252
  - bug caused problems installing `psutil` package
- use `${VIRTUAL_ENV}/bin/pip` if run for `venv` before `$PATH` is changed
  - avoid updating system environment in some cases, defence in depth in others
- install `sudo` package in Mariner 2.0 and Azure Linux 3.0 images
- update /etc/sudoers.d/helixbot instead of /etc/sudoers everywhere
- nits:
  - remove `helixbot2` users; they no longer help anything
  - remove `virtualenv` module and comments it
  - remove `helix-scripts` wheels after installing them
  - make some syntax a bit more consistent


dougbu:
build is green and I've done some local validation. ready for review&hellip;

src/ubuntu/24.10/helix/Dockerfile
```diff
@@ -76,8 +77,14 @@ RUN ARCH=$TARGETARCH \
     && dpkg -i libmsquic* \
     && rm libmsquic*
 
+# Move user (probably ubuntu) with UID 1000 to UID 2000 to avoid conflict with helixbot
+RUN user_id=$(id -un 1000) && \
```
richlander, line 81: This strikes me as an undesirable solution, at least in the long-term. This approach is committing us to including all of this `/etc/passwd` management over the long-term in our Ubuntu Dockerfiles. Ideally, we'd change the uid we expect.

Node images use this same uid. They appear to be skipping this problem by not supporting Ubuntu.

https://hub.docker.com/_/node/

src/alpine/3.18/helix/Dockerfile
```diff
@@ -52,5 +53,5 @@ USER helixbot
 # Install Helix Dependencies
 ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
 RUN python3 -m venv $VIRTUAL_ENV
-ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+ENV PATH="${VIRTUAL_ENV}/bin:$PATH"
```
mthalman, line 56: I notice inconsistent usage of this syntax. In the AlmaLinux Dockerfile, it used `${PATH}` but here it does not. I think they should all be consistent.

src/azurelinux/3.0/helix/Dockerfile
```diff
@@ -29,6 +30,7 @@ RUN tdnf install --setopt tsflags=nodocs --refresh -y \
         python3 \
         python3-pip \
         shadow-utils \
+        sudo \
```
mthalman, line 33: What requires `sudo`? These images have been working for a long time without it.

src/centos-stream/9/mlnet/helix/amd64/Dockerfile
```diff
@@ -44,8 +44,9 @@ RUN dnf upgrade --refresh -y \
         openssl-devel \
         perl-FindBin \
         procps-ng \
-        python3 \
-        python3-devel \
+        python3.12 \
```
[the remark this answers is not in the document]
mthalman: Yeah, I'm not sure why this is being done. The version associated with `python3` is 3.9 which is the same version that exists in some other Helix images. So why does this need to be higher?

src/ubuntu/24.10/helix/Dockerfile
lbussell: @dougbu, for reference, here's the script devcontainers uses to change UIDs... https://github.com/devcontainers/cli/blob/main/scripts/updateUID.Dockerfile

Is anything hardcoded to reference the `helixbot` user, or could you use the built-in ubuntu user instead?

mthalman: It seems a lot easier to just delete the `ubuntu` user. This is what the devcontainer Dockerfile does for .NET: https://github.com/devcontainers/images/blob/4828efe3844f92a9802aba9831b3567d79f9c958/src/dotnet/.devcontainer/Dockerfile#L8

richlander: I'm hoping we can migrate to a new UID over time.

You can see that we tried to avoid this problem with our official images, with our `app` user. For clarity, we developed this solution MANY years later, so had the benefit of better guidance.

```bash
$ docker run --rm mcr.microsoft.com/dotnet/runtime-deps cat /etc/passwd | tail -n 1
app:x:1654:1654::/home/app:/bin/sh
```

src/alpine/3.18/helix/Dockerfile
dougbu: sure. expect in the next iteration (likely tomorrow at best)

src/azurelinux/3.0/helix/Dockerfile
dougbu: `sudo` worked in the majority of the Helix containers and some teams use it in their testing scenarios. I also kept hitting issues in the updated containers as I tried to understand problems while I worked through iterations of this PR. so, consistency, muscle memory, and customer scenarios all play into this

we can discuss removing `sudo` where that's an option later. doing so will make some things more difficult around investigations and enhancements in this repo e.g., on OSes whose package managers can only be run as root &mdash; even when querying but it may not break the world

src/centos-stream/9/mlnet/helix/amd64/Dockerfile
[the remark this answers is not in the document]
dougbu: > Versioning pinning Python is a new practice

not really. it was used already in src/opensuse/15.6/helix/amd64/Dockerfile

the background is fairly simple: OS vendors sometimes have Python dependencies of their own or leave the Python packages alone after a release. this means our software can easily fall behind Python's deprecations and we have to support a broader range of Python versions than is manageable

the versioned packages move forward and are sometimes backported to prior releases, reducing our costs. see below for more about Centos Stream 9 in particular (bumping Python elsewhere in this PR e.g. on Debian 11 was required)

----

here and in src/centos-stream/9/helix/amd64/Dockerfile, this update avoided some messier workarounds for problems installing `helix-scripts` dependencies when using Python 3.9. I don't remember the exact details but believe this was the most straightforward fix

that said
1. 3.9 will be out of support in October IIRC
2. I'll try the `python -m pip install --upgrade pip` workaround used in src/debian/11/helix/arm64v8/Dockerfile (where 3.9 is the newest available Python)

src/ubuntu/24.10/helix/Dockerfile
dougbu: there's a lot to unpack here but I'd like to avoid the long-term considerations for a bit longer. much of the infrastructure in Helix will need to change as we go further and it's not clear whether UID alignment will be required even mid-term. for now, I'm dealing w/ a few exceptions (Azure Linux 3 hosts, Ubuntu 24.* containers) rather than doing anything broader

background here is a number of attempts I made e.g., to expand file permissions in the hosts and containers, update default permissions, et cetera. but kept coming back to the unfortunate conclusion we need to align UIDs for now. without that alignment Helix work items fail or (worse) succeed but mask problems reporting tests and so on

the script you referenced @lbussell is messy and focuses entirely on /etc/passwd and /etc/groups, files that mean less and less these days. plus that script only updates the discovered user's home directory, not other files the user may own. `usermod` handles the user's home automatically

----

deleting the `ubuntu` user sounds great to me. any objections to that approach in another iteration❓

richlander: I think that's fine for the near-term.

src/centos-stream/9/mlnet/helix/amd64/Dockerfile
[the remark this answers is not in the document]
richlander: It would be good to document in this repo (not in this issue):

- Min required Python version
- Encouraged patterns for that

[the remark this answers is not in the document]
dougbu: I agree the words need to land somewhere other than this PR. where would be appropriate❓

----

for future reference and given pushback here, I think the rules are something like:
- try other workarounds b/4 moving to non-default OS packages for Python when problems arise
- if either ^^ is not possible or the default OS packages for Python install a deprecated version, move to the latest available Python 3 packages that are for released (not Beta or RC) versions
- today, minimum Python version should be 3.9

patterns when "pinning" is needed include
- pin version number for all `python3*` packages in affected Dockerfile unless unavailable for an OS e.g. `python312-pip` exists for SUSE 15.6 but only `python3-pip` exists for Ubuntu 20.04 (though `python3.9-dev` is what we're using now)
- don't include two Python versions in the OS packages list i.e., change the OS package list and don't add to it
  - if OS itself depends on deprecated / busted-for-us Python, this will be additive where it wasn't b/4 and that's fine
- link `/usr/bin/python` to specific `/usr/bin/python3.{number}` executable matching "pinned" Python version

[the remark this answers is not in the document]
dougbu: I resolved original comments in current iteration

src/debian/11/helix/arm64v8/Dockerfile
```diff
@@ -43,20 +43,21 @@ RUN apt-get update \
...
-# additionally, preinstall the virtualenv packages used for VSTS reporting to save time
-RUN /usr/sbin/adduser --disabled-password --gecos '' --uid 1001 --shell /bin/bash --ingroup adm helixbot \
-    && /usr/sbin/adduser --disabled-password --gecos '' --uid 1000 --shell /bin/bash --ingroup adm helixbot2 \
+# Create helixbot user and give rights to sudo without password
+RUN /usr/sbin/adduser --disabled-password --gecos '' --uid 1000 --shell /bin/bash --ingroup adm helixbot \
+    && usermod -aG sudo helixbot \
     && chmod 755 /root \
-    && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers \
-    && echo "helixbot2 ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers 
+    && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" | (EDITOR="tee -a" visudo)
```
dougbu, line 52: I just realized the reason `sudo` wasn't working in this image. problem is deleted line above replaced the whole /etc/sudoers file, removing `helixbot` user from it.

I could remove the `usermod` addition and go back to
```suggestion
    && echo "helixbot ALL=(ALL)       NOPASSWD: ALL" > /etc/sudoers
```
if you have strong feelings about this

I'm fine either way though `visudo` is strongly recommended everywhere anyone wants to edit /etc/sudoers and replacing that entire file is unnecessary

regardless of other opinions, I'll remove the `usermod` and see if that works locally

dougbu: removing `usermod` worked

we have two choices from my perspective:
1. consistently use `> /etc/sudoers` and thereby replace the whole file
2. consistently use `| (EDITOR="tee -a" visudo)`

I lean toward the second option. it's somewhat further from what's in 'main' but follows accepted Linux practise. not sure of any real downside. I'll admit we could reliably use `>> /etc/sudoers` too but don't see the advantage over the `visudo` gesture

src/centos-stream/9/mlnet/helix/amd64/Dockerfile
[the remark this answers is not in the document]
richlander: I think start with an issue on what the model should be for https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1374#discussion_r1978413157. After we lock that down, we can leave it as an issue and link to it or move it to a repo doc.

src/debian/11/helix/arm64v8/Dockerfile
dougbu: smallish reasons to prefer my second choice:
- this is one of the few ways the non-Helix containers are a bit more capable
- for example, `docker run ... --user root ...` and attempting to use `sudo` in the container results in `root is not in the sudoers file.  This incident will be reported.` in most Helix containers but not the revised Debian 11 ones

src/ubuntu/24.04/helix/Dockerfile
```diff
@@ -73,8 +74,12 @@ RUN curl -LO https://packages.microsoft.com/keys/microsoft.asc && \
     apt-get install -y libmsquic && \
     rm -rf /var/lib/apt/lists/*
 
+# Remove user (probably ubuntu) with UID 1000 to UID 2000 to avoid conflict with helixbot
+RUN user_id=$(id -un 1000) && \
+    /usr/sbin/userdel --force --remove $user_id
```
mthalman, line 79: Why is this also not done in the older versions of Ubuntu Helix Dockerfiles as well?

src/debian/11/helix/arm64v8/Dockerfile
mthalman: Honestly, I would prefer using `>> /etc/sudoers`. That is a more understood syntax by a broader set of people, IMO. I had no idea what `EDITOR="tee -a" visudo` did without looking it up. The benefits that `visudo` provides aren't necessary in this context.

src/ubuntu/24.04/helix/Dockerfile
dougbu: b/c a user w/ UID `1000` doesn't exist prior to 24.04

mthalman: Well that explains it. 😄

src/debian/11/helix/arm64v8/Dockerfile
dougbu: hadn't noticed b/4 but recent Alpine images write to /etc/sudoers.d/helixbot instead of /etc/sudoers. the two files are essentially equivalent as long as we're appending to /etc/sudoers. I don't have a strong preference either way but suggest we use the same file in all Helix images. thoughts❓

mthalman: Copilot says that writing to `/etc/sudoers.d/helixbot` is considered the better practice. Reference: https://unix.stackexchange.com/questions/375433/etc-sudoers-vs-etc-sudoers-d-file-for-enabling-sudo-for-a-user

dougbu: done in latest iteration. I also rebased on latest main and squashed the commits. will update the PR description to match the latest in a minute

mthalman: Why did you squash? That makes it very hard to discern the latest changes from everything else.

dougbu: sorry. I did it b/c I wanted to ensure my description of the final changes was complete in the simplest way

literally the only changes from the previous iteration were the switch to the /etc/sudoers.d/helixbot

mthalman (approved):

Issue #1371, opened 2025-02-25
Develop policy for adopting LLVM minor version updates


richlander:
It looks like we have an inconsistent approach to adopting LLVM/clang minor versions. We should likely revisit that.

Query: https://github.com/search?q=repo%3Adotnet%2Fdotnet-buildtools-prereqs-docker%20LLVM_VERSION&type=code

Ideally, we'd move forward with minor versions to ensure we pick up any security of performance updates. We also would need a way to validate those updates with runtime before publishing. We likely do that with some fancy (with some definition of that term) `manifest.json` edits to produce a new tag name for a temporary period.

@sbomer

mthalman:
> We also would need a way to validate those updates with runtime before publishing. We likely do that with some fancy (with some definition of that term) manifest.json edits to produce a new tag name for a temporary period.

Gosh, do we need a nightly branch in this repo too? 🫠

richlander:
That's a good question.

Also see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1369.

lbussell:
I think that having two images with different floating tags for a temporary period is preferable to spinning up a second repo. Like alpine images in dotnet-docker.

mthalman:
[Triage]

Assigning to @sbomer to take a look at defining a policy.

richlander:
We should bias towards stability-increasing approaches. Pinned image references enables us to merge with confidence in this repo w/o requiring a nightly branch and improves reliability in dotnet/runtime.

It hadn't occurred to me that pinned references were the solution to bumping LLVM minor versions safely. Sounds great! I was not keen on creating a whole other repo for that w/a lot care and feeding required.

mthalman:
A pinning strategy does decrease the priority of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1373 which I'm in favor of. The use of Renovate is outlined in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1321. I'm still working through the pattern for that to reduce any security concerns.

richlander:
Here's a thought. Let's see if you agree.

- Pinning is a generally useful strategy and is overall cheaper so we should start with it.
- If pinning isn't sufficient on its own, we should add staging to help us update images more safely.

IOW, I see these strategies more as a progression than alternatives. However, we should start with just one.

mthalman:
[...]

PR #1375, opened 2025-03-03
Add git sha tags for net10.0 images

sbomer:
We need to update the LLVM version used in our build images. As discussed in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1371, we would like a long-term solution where dotnet/runtime references images by sha, and they get updated via dependency flow PRs. In the meantime, we can temporarily pin dotnet/runtime images to a specific version to avoid build instability while we roll out new images for https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1369.

This change adds back git sha based tags to make it easier to manually pin the images (not to a specific build, but to a specific sha - should be good enough for our purposes).

### DIFF
--- a/src/azurelinux/manifest.json
+++ b/src/azurelinux/manifest.json
@@ -648,7 +648,8 @@
               "os": "linux",
               "osVersion": "azurelinux3.0",
               "tags": {
-                "azurelinux-3.0-net10.0-cross-riscv64": {}
+                "azurelinux-3.0-net10.0-cross-riscv64": {},
+                "azurelinux-3.0-net10.0-cross-riscv64-$(System:DockerfileGitCommitSha)": {}
               }
             }
           ]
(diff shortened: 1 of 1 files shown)


mthalman:
@sbomer - Do you see this as a temporary solution until we can incorporate dependency flow for image digests? I ask because the commit sha is a fairly crude way of versioning, particularly for Dockerfiles that derive from images produced by other Dockerfiles in the repo. For example, changing the content of [`azurelinux-3.0-net10.0-crossdeps-amd64`](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/azurelinux/3.0/net10.0/crossdeps/amd64/Dockerfile) will not result in a new commit sha value for the derived image [`azurelinux-3.0-net10.0-crossdeps-llvm-amd64`](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/azurelinux/3.0/net10.0/crossdeps-llvm/amd64/Dockerfile).

mthalman (approved):

mthalman:
Ran into this issue: https://github.com/dotnet/docker-tools/issues/1493. I'm wrking to resolve things to get these tags published.

PR #1382, opened 2025-03-06
Update Python version on Windows

mthalman:
The Python version in the Windows Helix Dockerfiles is outdated. This ensures it gets the latest. This is a port of the changes that originally existed in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1336.

### DIFF
--- a/src/nanoserver/1809/helix/amd64/Dockerfile
+++ b/src/nanoserver/1809/helix/amd64/Dockerfile
@@ -10,28 +10,31 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
     Expand-Archive -Path C:\powershell.zip -DestinationPath C:\powershell; `
     Remove-Item -Path C:\powershell.zip
 
+# Download latest stable version of Python
+RUN $apiUrl = 'https://api.nuget.org/v3-flatcontainer/python/index.json'; `
+    $response = Invoke-RestMethod -Uri $apiUrl; `
+    $versions = $response.versions | Where-Object { $_ -notmatch '-' } | Sort-Object { [version]$_ } -Descending; `
+    $latestVersion = $versions[0]; `
+    echo \"Downloading Python $latestVersion\"; `
+    Invoke-WebRequest -Uri https://www.nuget.org/api/v2/package/python/$latestVersion -OutFile $env:TEMP\python.zip; `
+    md C:\PythonTemp; `
+    tar -zxf  $env:TEMP\python.zip -C C:\PythonTemp;
+
 
 FROM mcr.microsoft.com/windows/nanoserver:1809
 
-SHELL ["cmd", "/S", "/C"]
 USER ContainerAdministrator
-ENTRYPOINT C:\Windows\System32\cmd.exe
 
 COPY --from=installer [ "C:\\powershell\\", "C:\\Program Files\\PowerShell\\" ]
+COPY --from=installer [ "C:\\PythonTemp\\tools", "C:\\Python\\" ]
 
-RUN curl -SL --output %TEMP%\python.zip https://www.nuget.org/api/v2/package/python/3.7.3 `
-    && md C:\Python C:\PythonTemp `
-    && tar -zxf %TEMP%\python.zip -C C:\PythonTemp `
-    && xcopy /s c:\PythonTemp\tools C:\Python `
-    && rd /s /q c:\PythonTemp `
-    && del /q %TEMP%\python.zip `
-    && setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;"
+RUN setx /M PATH "%PATH%;C:\Program Files\PowerShell\;C:\Python;C:\python\scripts" && `
+    setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;" && `
+    setx /M VIRTUAL_ENV "C:\Python-env"
 
-RUN setx /M PATH "%PATH%;C:\Program Files\PowerShell\;C:\Python;C:\python\scripts"
-
-RUN md c:\\helixtmp && pushd c:\\helixtmp &&`
-    python -m pip install --upgrade pip==20.2 && `
-    python -m pip install virtualenv==16.6.0 && `
+RUN md c:\\helixtmp && pushd c:\\helixtmp && `
+    python -m pip install --upgrade pip && `
+    python -m venv %VIRTUAL_ENV% && `
     pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && `
     for %f in (.\helix_scripts-*-py3-none-any.whl) do (pip install %f) && `
     popd && rd /s /q c:\\helixtmp && `
@@ -46,4 +49,3 @@ RUN md c:\\helixtmp && pushd c:\\helixtmp &&`
         Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\corerun.exe' -Value 2 -Name DumpCount -Force
 
 WORKDIR C:\\Work
-
--- a/src/windowsservercore/ltsc2019/helix/amd64/Dockerfile
+++ b/src/windowsservercore/ltsc2019/helix/amd64/Dockerfile
@@ -1,22 +1,32 @@
 # escape=`
 FROM mcr.microsoft.com/windows/servercore:ltsc2019
 
-SHELL ["cmd", "/S", "/C"]
-USER ContainerAdministrator
+# Install latest stable version of Python
+RUN powershell -Command "`
+        $ErrorActionPreference = 'Stop'; `
+        $ProgressPreference = 'SilentlyContinue'; `
+        $apiUrl = 'https://api.nuget.org/v3-flatcontainer/python/index.json'; `
+        $response = Invoke-RestMethod -Uri $apiUrl; `
+        $versions = $response.versions | Where-Object { $_ -notmatch '-' } | Sort-Object { [version]$_ } -Descending; `
+        $latestVersion = $versions[0]; `
+        echo \"Downloading Python $latestVersion\"; `
+        Invoke-WebRequest -Uri https://www.nuget.org/api/v2/package/python/$latestVersion -OutFile $env:TEMP\python.zip; `
+        md C:\Python; `
+        md C:\PythonTemp; `
+        tar -zxf  $env:TEMP\python.zip -C C:\PythonTemp; `
+        xcopy /s c:\PythonTemp\tools C:\Python; `
+        Remove-Item -Recurse -Force C:\PythonTemp; `
+        Remove-Item -Force $env:TEMP\python.zip;"
 
-RUN curl -SL --output %TEMP%\python.zip https://www.nuget.org/api/v2/package/python/3.7.3 `
-    && md C:\Python C:\PythonTemp `
-    && tar -zxf %TEMP%\python.zip -C C:\PythonTemp `
-    && xcopy /s c:\PythonTemp\tools C:\Python `
-    && rd /s /q c:\PythonTemp `
-    && del /q %TEMP%\python.zip `
-    && setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;"
+RUN setx /M PATH "%PATH%;C:\Python;C:\python\scripts" && `
+    setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;" && `
+    setx /M VIRTUAL_ENV "C:\Python-env"
 
-RUN md c:\\helixtmp && pushd c:\\helixtmp &&`
-    C:\Python\python.exe -m pip install --upgrade pip==20.2 --no-warn-script-location && `
-    C:\Python\python.exe -m pip install virtualenv==16.6.0 --no-warn-script-location && `
-    C:\Python\python.exe -m pip download --no-deps helix-scripts --index-url https://dnceng.pkgs.visualstudio.com/public/_packaging/helix-client-prod/pypi/simple && `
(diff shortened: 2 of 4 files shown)


src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -10,28 +10,31 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
     Expand-Archive -Path C:\powershell.zip -DestinationPath C:\powershell; `
     Remove-Item -Path C:\powershell.zip
 
+# Download latest stable version of Python
+RUN $apiUrl = 'https://api.nuget.org/v3-flatcontainer/python/index.json'; `
+    $response = Invoke-RestMethod -Uri $apiUrl; `
+    $versions = $response.versions | Where-Object { $_ -notmatch '-' } | Sort-Object { [version]$_ } -Descending; `
+    $latestVersion = $versions[0]; `
```
dougbu, line 17: I believe this means we'll jump from current version (`3.7.3`) to `3.13.2`. am I reading this correctly❓

if yes, this is fine b/c we've tested `helix-scripts` using `3.13.x` on Windows. however we may need to "pin" the version again b/c the https://www.nuget.org/packages/python owners don't treat breaking changes the same way most OS vendors treat their packages. put another way, we may need to pin the version again in the future &mdash; hopefully temporarily

unfortunate downside here is such pinning will be a reaction to problems various teams hit after an image is regenerated. we probably need some way to include Helix SDK e2e testing similar to what's in the [dotnet-helix-machines-pr pipeline](https://dev.azure.com/dnceng/internal/_git/dotnet-helix-machines?path=/azure-pipeline-pr.yaml&version=GCab85f6eddd250fb808bb3c0454a2c6de90eff7aa&line=921&lineEnd=969&lineStartColumn=1&lineEndColumn=36&lineStyle=plain&_a=contents)

src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -10,28 +10,31 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
...
+    setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;" && `
+    setx /M VIRTUAL_ENV "C:\Python-env"
 
-RUN setx /M PATH "%PATH%;C:\Program Files\PowerShell\;C:\Python;C:\python\scripts"
-
-RUN md c:\\helixtmp && pushd c:\\helixtmp &&`
-    python -m pip install --upgrade pip==20.2 && `
-    python -m pip install virtualenv==16.6.0 && `
+RUN md c:\\helixtmp && pushd c:\\helixtmp && `
+    python -m pip install --upgrade pip && `
```
dougbu, line 36: please do this upgrade inside the `venv` i.e., move this down one line

dougbu (commented):
I reviewed the first file. am I correct that the other Dockerfiles are different mainly around base images and such❓

src/nanoserver/1809/helix/amd64/Dockerfile
dougbu: actually, the container will likely build just fine w/o updating `pip` at all. the upgrade only seems to be necessary when using older Python versions

mthalman:
This is getting a hadolint error:

```
/github/workspace/src/nanoserver/1809/helix/amd64/Dockerfile:35 SC1072 error:  Fix any mentioned problems and try again.
```

Caused by this line: `for %f in (.\helix_scripts-*-py3-none-any.whl) do (pip install %f)`

What's notable is that this line hasn't been changed in this PR. It's the other changes in this PR that have now exposed this hadolint parsing issue: https://github.com/hadolint/hadolint/issues/436

src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -10,28 +10,31 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
...
 
-RUN curl -SL --output %TEMP%\python.zip https://www.nuget.org/api/v2/package/python/3.7.3 `
-    && md C:\Python C:\PythonTemp `
-    && tar -zxf %TEMP%\python.zip -C C:\PythonTemp `
-    && xcopy /s c:\PythonTemp\tools C:\Python `
-    && rd /s /q c:\PythonTemp `
-    && del /q %TEMP%\python.zip `
-    && setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;"
+RUN setx /M PATH "%PATH%;C:\Program Files\PowerShell\;C:\Python;C:\python\scripts" && `
+    setx /M PYTHONPATH "C:\Python\Lib;C:\Python\DLLs;" && `
```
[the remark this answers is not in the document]
mthalman: > this line should break the current `helix-scripts` Windows Docker entry point script b/c that `unset`s `%PYTHONPATH%` then writes a new value pointing only at the `helix-scripts` code.

How would it break that? First, this already exists in the images and they're working. Second: if the entrypoint script resets the value of `PYTHONPATH`, that's fine. But having it set beforehand shouldn't be an issue. I'm fine with experimenting with the removal of this in later work.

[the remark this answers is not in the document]
dougbu: it would break the Docker entry point script _if_ the line mattered. since existing images work, I'm fine w/ doing the cleanup later. suggest filing an issue so we get to that sometime soon

src/nanoserver/1809/helix/amd64/Dockerfile
mthalman: Noted. I think getting the latest will be good for now. Eventually, I'd like to get this to be pinned and updated automatically via PR generation with Renovate: https://github.com/dotnet/arcade/pull/15594.

src/nanoserver/1809/helix/amd64/Dockerfile
mthalman: It does more than just upgrades pip. It also ensures that pip is on the `PATH`, for example. So if you take this line out, you can't just run `pip` because it won't be found on the `PATH`.

src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -10,28 +10,31 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
...
+
 
 FROM mcr.microsoft.com/windows/nanoserver:1809
 
-SHELL ["cmd", "/S", "/C"]
 USER ContainerAdministrator
-ENTRYPOINT C:\Windows\System32\cmd.exe
 
 COPY --from=installer [ "C:\\powershell\\", "C:\\Program Files\\PowerShell\\" ]
+COPY --from=installer [ "C:\\PythonTemp\\tools", "C:\\Python\\" ]
```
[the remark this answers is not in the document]
mthalman: fixed

src/nanoserver/1809/helix/amd64/Dockerfile
mthalman: I've updated the pip install to be after venv

src/nanoserver/1809/helix/amd64/Dockerfile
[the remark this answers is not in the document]
mthalman: Logged https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1392

src/nanoserver/1809/helix/amd64/Dockerfile
dougbu: that's good. however you're not adding `pip.exe` to `$env:PATH` at all. `venv` creation places `python.exe`, `pip.exe`, and so on in `$env:VIRTUAL_ENV\\Scripts`  but I don't see that in `$env:PATH` in the current changes

dougbu:
not sure of the include capabilities that may be available here. is it possible to consolidate the many common pieces of these 4 DockerFiles❓

mthalman:
> not sure of the include capabilities that may be available here. is it possible to consolidate the many common pieces of these 4 DockerFiles❓

I would love that and have wanted it for a while now. I thought we had an issue for it, but I guess not. So I logged it: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1393

src/nanoserver/1809/helix/amd64/Dockerfile
mthalman: Local testing of removing the pip install shows this output:

```
[notice] A new release of pip is available: 24.3.1 -> 25.0.1
[notice] To update, run: C:\Python-env\Scripts\python.exe -m pip install --upgrade pip
```

Let me know if the install should be added back because of it.

dougbu: nah, we tend to stick w/ `pip` versions that aren't causing problems or just use what the Python installation provides. can suppress the message w/ `--disable-pip-version-check`. in dotnet-build-machines, we use that on every `python -m pip install ...` command line

src/windowsservercore/ltsc2019/helix/amd64/Dockerfile
```diff
@@ -1,33 +1,69 @@
 # escape=`
 FROM mcr.microsoft.com/windows/servercore:ltsc2019
 
-SHELL ["cmd", "/S", "/C"]
-USER ContainerAdministrator
```
dougbu, line 5: user `ContainerAdministrator` must exist for Helix scenarios. this wasn't removed in src/nanoserver/1809/helix/amd64/Dockerfile but was in src/windowsservercore/*

src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -13,37 +13,73 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
...
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\dotnet.exe' -Value 2 -Name DumpType -Force ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\dotnet.exe' -Value 'C:\cores' -Name DumpFolder -Force ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\dotnet.exe' -Value 2 -Name DumpCount -Force ; `
-        New-Item -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\' -Name 'corerun.exe' -Force -ErrorAction SilentlyContinue ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\corerun.exe' -Value 2 -Name DumpType -Force ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\corerun.exe' -Value 'C:\cores' -Name DumpFolder -Force ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\corerun.exe' -Value 2 -Name DumpCount -Force
+RUN setx /M PATH "%PATH%;C:\Program Files\PowerShell\;C:\Python;C:\python\scripts" && `
```
dougbu, line 20: shouldn't break anything (I hope) but suggest defining `%VIRTUAL_ENV%` first then including `%VIRTUAL_ENV%\\Scripts` in `%PATH%`

src/windowsservercore/ltsc2019/helix/amd64/Dockerfile
mthalman: That's because in Windows Server Core containers, `ContainerAdministrator` is the default user. But in Nano Server it's `ContainerUser`. That means this line is redundant, which is why I removed it.

src/nanoserver/1809/helix/amd64/Dockerfile
```diff
@@ -13,37 +13,73 @@ RUN $apiUrl = 'https://api.github.com/repos/PowerShell/PowerShell/releases/lates
...
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\dotnet.exe' -Value 'C:\cores' -Name DumpFolder -Force ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\dotnet.exe' -Value 2 -Name DumpCount -Force ; `
-        New-Item -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\' -Name 'corerun.exe' -Force -ErrorAction SilentlyContinue ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\corerun.exe' -Value 2 -Name DumpType -Force ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\corerun.exe' -Value 'C:\cores' -Name DumpFolder -Force ; `
-        Set-ItemProperty -Path 'HKLM:\SOFTWARE\Microsoft\Windows\Windows Error Reporting\LocalDumps\corerun.exe' -Value 2 -Name DumpCount -Force
+RUN setx /M VIRTUAL_ENV "C:\Python-env" && `
+    setx /M PATH "%PATH%;C:\Program Files\PowerShell\;C:\Python;C:\python\scripts;%VIRTUAL_ENV%\Scripts" && `
```
dougbu, line 21:
```suggestion
    setx /M PATH "%PATH%;C:\Program Files\PowerShell\;%VIRTUAL_ENV%\Scripts;C:\Python;C:\python\scripts" && `
```

dougbu (approved):
no need to ping me again but do reorder `$env:PATH` so the `venv` "wins" once created b/4 merging

Issue #1384, opened 2025-03-06
Clean up optimization image


sbomer:
The optimization images used to be fairly simple - they would install llvm from the package repositories (see https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/872). Now we're building LLVM from source: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/azurelinux/3.0/net10.0/opt/Dockerfile.

Not a high priority, but it might be worth cleaning up that dockerfile so that the final image doesn't have the llvm release keys, build dependencies, or build intermediates. For reference, here's how we copy llvm into the cross-build images: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/azurelinux/3.0/net10.0/crossdeps-llvm/amd64/Dockerfile.

PR #1399, opened 2025-03-20
Clean up net10.0 opt image

sbomer:
Fixes https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1384

### DIFF
--- a/src/azurelinux/3.0/net10.0/opt/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/opt/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/azurelinux/base/core:3.0
+FROM mcr.microsoft.com/azurelinux/base/core:3.0 AS builder
 
 RUN tdnf update -y && \
     tdnf install -y \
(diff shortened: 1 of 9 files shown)


src/azurelinux/3.0/net10.0/opt/Dockerfile
```diff
@@ -50,3 +48,20 @@ RUN mkdir llvm-project.src && \
...
+        ca-certificates \
+        git \
+        tar \
+        util-linux \
+        wget \
+        # Runtime dependencies
+        icu
+
+# Install LLVM that we built from source
+COPY --from=builder /usr/local /usr/local
```
mthalman, line 67: Can this be more targeted in what it copies? Because of how Docker works, it will end up copying over all existing files (even though they are the same), essentially duplicating this data between the two image layers. So it's a bunch of wasted space that's caused here.

jkoritzinsky: If we set `CMAKE_INSTALL_PREFIX` in the LLVM build command above to another path (can be totally custom), we could copy the `/usr/local` from under that folder and reduce the size hit here.

akoeplinger:
It's failing because of 
> clang: error: invalid linker name in argument '-fuse-ld=lld'

looks like it can't find the lld we build?

akoeplinger:
looks like azurelinux is fixed now but mariner needs the same fix

mthalman (approved):

PR #1407, opened 2025-04-17
Add Fedora 42 test image

richlander:
- https://docs.fedoraproject.org/en-US/releases/f42/
- https://fedoramagazine.org/announcing-fedora-linux-42/


richlander:
@siramvikram

mthalman (approved):

PR #1408, opened 2025-04-22
Add azurelinux3.0-source-build-test image

Winniexu01:
Related to https://github.com/dotnet/source-build/issues/3816

Create a specifially image with .NET SDK and scancode installed.


src/ubuntu/24.04/sdk/amd64/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+        uuid-dev \
+        zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Dependencies for VMR/source-build tests
+RUN apt-get update \
+    && apt-get install -y \
+        elfutils \
+        file \
+    && rm -rf /var/lib/apt/lists/*
```
mthalman, line 48: None of these dependencies should be needed. We're not building .NET from source with this container. We're just building the test project which only requires the .NET installation which is already provided in the base `sdk` image.

src/ubuntu/24.04/sdk/amd64/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+# install scancode
+RUN apt-get update  \
+    && apt-get install -y \
+        python3 \
+        python3-venv \
+        dos2unix \
```
mthalman, line 55: This shouldn't be needed. The script file will have the correct line endings when committed in the repo.

src/ubuntu/24.04/sdk/amd64/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    && rm -rf /var/lib/apt/lists/*
+COPY ./install-scancode.sh /tmp
```
mthalman, line 57: There really isn't a need to have this logic defined in a script file. I'd prefer it to be defined directly in the Dockerfile.

src/ubuntu/24.04/sdk/amd64/Dockerfile
```diff
@@ -0,0 +1,61 @@
+FROM mcr.microsoft.com/dotnet/sdk:10.0-preview
```
mthalman, line 1: The tag should be explicit about which distro version it's expecting. 

```suggestion
FROM mcr.microsoft.com/dotnet/sdk:10.0-preview-noble
```

src/ubuntu/manifest.json
```diff
@@ -302,6 +302,19 @@
...
         },
+        {
+          "platforms": [
+            {
+              "architecture": "amd64",
+              "dockerfile": "src/ubuntu/24.04/sdk/amd64",
+              "os": "linux",
+              "osVersion": "noble",
+              "tags": {
+                "ubuntu-24.04-sdk-amd64": {}
```
mthalman, line 313: The tag name (and file path) should reflect its purpose. The `CODEOWNERS` file will also need to be updated to reflect this.

```suggestion
              "dockerfile": "src/ubuntu/24.04/source-build-test/amd64",
              "os": "linux",
              "osVersion": "noble",
              "tags": {
                "ubuntu-24.04-source-build-test-amd64": {}
```

src/ubuntu/24.04/sdk/amd64/Dockerfile
MichaelSimons, line 1: Why are we not using Azure Linux?

mthalman: Yeah, good point. Should be `10.0-preview-azurelinux3.0` instead.

src/azurelinux/3.0/net10.0/source-build-test/amd64/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN SCANCODE_VERSION="32.3.3" && \
+    #!/bin/bash
+    set -euo pipefail && \
+    python3 -m venv $venv && \
+    source $venv/bin/activate && \
+    pip install scancode-toolkit==$SCANCODE_VERSION && \
+    deactivate
+
+# Setup a script which executes scancode in the virtual environment
+RUN cat > /usr/local/bin/scancode <<EOF
```
MichaelSimons, line 16: Thoughts on checking this in next to the Dockerfile and copying it in?  The benefit is improved maintainability.

src/azurelinux/3.0/net10.0/source-build-test/amd64/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+scancode "\$@"
+deactivate
+EOF
+RUN chmod +x /usr/local/bin/scancode
+
+# Install necessary dependencies
+RUN tdnf update -y && \
+    tdnf install -y \
+        libgomp \
+    && tdnf clean all
```
mthalman, line 29: Installing dependencies should be done before setting up the script file. This takes advantage of Docker's layer caching behavior on dev machines.

src/azurelinux/3.0/net10.0/source-build-test/amd64/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM mcr.microsoft.com/dotnet/sdk:10.0-preview-azurelinux3.0-amd64
+
+# Install scancode
+# Install instructions: https://scancode-toolkit.readthedocs.io/en/latest/getting-started/install.html#installation-as-a-library-via-pip
+# See latest release at https://github.com/nexB/scancode-toolkit/releases
+ENV venv="/tmp/scancode-env"
+RUN SCANCODE_VERSION="32.3.3" && \
```
mthalman, line 7: Preference is for the syntax to use `&&` at the beginning of subsequent lines.

src/azurelinux/3.0/net10.0/source-build-test/amd64/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM mcr.microsoft.com/dotnet/sdk:10.0-preview-azurelinux3.0-amd64
```
mthalman, line 1: We should use a multi-stage Dockerfile instead so that scancode is installed in a separate stage and then copied to the final stage. This saves about 300 MB of disk space in the final image.

mthalman (commented):
Based on all the comments, here's a suggested implementation:

```Dockerfile
FROM mcr.microsoft.com/dotnet/sdk:10.0-preview-azurelinux3.0-amd64 AS installer

# Install scancode
# Install instructions: https://scancode-toolkit.readthedocs.io/en/latest/getting-started/install.html#installation-as-a-library-via-pip
# See latest release at https://github.com/nexB/scancode-toolkit/releases
RUN SCANCODE_VERSION="32.3.3" \
    && python3 -m venv /venv \
    && source /venv/bin/activate \
    && pip install scancode-toolkit==$SCANCODE_VERSION


FROM mcr.microsoft.com/dotnet/sdk:10.0-preview-azurelinux3.0-amd64

COPY --from=installer /venv /venv

# Install necessary dependencies
RUN tdnf update -y \
    && tdnf install -y \
        libgomp \
    && tdnf clean all

# Setup a script which executes scancode in the virtual environment
COPY --chmod=755 ./run-scancode.sh /usr/local/bin/scancode
```

src/azurelinux/3.0/net10.0/source-build-test/amd64/Dockerfile
Winniexu01: Ok, got it.

src/azurelinux/3.0/net10.0/source-build-test/amd64/Dockerfile
Winniexu01: Update: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1408/commits/4db806ca9dcbb77cc4f3fc408b2fb6bf36cb7fbf

MichaelSimons (approved):

mthalman (approved):

PR #1412, opened 2025-04-26
Add Alma 9

richlander:
- Add Alma 9
- Switch to new Dockerfile venv scheme
- Looks like Alma 10 hasn't shipped yet

### DIFF
--- a/src/almalinux/9/helix/amd64/Dockerfile
+++ b/src/almalinux/9/helix/amd64/Dockerfile
@@ -50,9 +50,9 @@ RUN dnf upgrade --refresh -y \
 ENV LANG=en_US.utf8
 
 # create helixbot user and give rights to sudo without password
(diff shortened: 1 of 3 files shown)


richlander:
Unsurprisingly, we need the base image uploaded/configured for our mirror.

@mthalman

src/almalinux/9/helix/amd64/Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    openssl-devel \
+    perl \
+    python3.12 \
+    python3.12-devel \
+    python3.12-pip \
+    sudo \
+    tar \
+    wget \
+    which \
+&& dnf clean all
```
mthalman, line 48: This should be indented.

mthalman (approved):

PR #1413, opened 2025-04-26
Upgrade gcc to v15

am11:
Release: https://gcc.gnu.org/gcc-15/


am11:
@mthalman, it requires `library/gcc:15-bookworm` uploaded to dotnet ACR.

jkoritzinsky (approved):

mthalman (approved):

am11:
Already done https://github.com/dotnet/runtime/pull/115128. :)

mthalman:
It's also referenced in the sdk repo:

https://github.com/dotnet/sdk/blob/a1da6b61d1552406d1a67cbbc19b25880f4b8308/.vsts-pr.yml#L34-L35

https://github.com/dotnet/sdk/blob/a1da6b61d1552406d1a67cbbc19b25880f4b8308/.vsts-ci.yml#L71-L72

am11:
It was a wrong dependency added in https://github.com/dotnet/sdk/commit/df4aa4ca3e40c1eb39e5d5d03d2a670354cccb36. gcc image was introduced to validate "runtime builds with gcc". Debian packaging use-case in SDK should be its own image. It doesn't require gcc, let alone the latest version.

PR #1440, opened 2025-05-29
Add missing dependencies for SDK NativeAOT testing

v-wuzhai:


src/centos-stream/9/helix/amd64/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM quay.io/centos/centos:stream9
 
 # Install dependencies
 RUN dnf upgrade --refresh -y \
+    && dnf install -y epel-release \
```
mthalman, line 5: It would be helpful to indicate this is needed in order to install `libunwind`.

PR #1442, opened 2025-05-30
Add Alpine 3.22 images

richlander:


src/alpine/3.22/helix/Dockerfile
```diff
@@ -0,0 +1,54 @@
+FROM library/alpine:3.22 AS venv
```
lbussell, line 1: Oops, I didn't import this one. I'll do that now.

lbussell (approved):
:shipit:

src/alpine/3.22/helix/Dockerfile
richlander: Which one did you import?

lbussell: This should be done now.

richlander: Ah. Got it.

PR #1443, opened 2025-06-01
Update llvm from version 20.1.4 to version 20.1.6

am11:


am11:
@mthalman there was a bug in the script not installing lldb-dev package which only affects riscv64 https://github.com/dotnet/arcade/pull/15886. Would it be possible to rebuild (only) riscv64 image? `src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile` (riscv64-musl is fine)

Thanks!

PR #1457, opened 2025-06-11
Use Ubuntu 22.04 for SQL Server. 24.04 is not fully supported yet.

cincuranet:
Related to #1455.

### DIFF
--- a/src/ubuntu/22.04/helix/sqlserver/amd64/Dockerfile
+++ b/src/ubuntu/22.04/helix/sqlserver/amd64/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-24.04-helix-amd64
+FROM mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-22.04-helix-amd64
 
 USER root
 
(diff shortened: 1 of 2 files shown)


src/ubuntu/24.04/helix/sqlserver/amd64/Dockerfile
```diff
@@ -8,6 +8,10 @@ RUN apt-get update \
         software-properties-common \
     && curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /usr/share/keyrings/microsoft-prod.gpg \
     && curl -fsSL https://packages.microsoft.com/config/ubuntu/22.04/mssql-server-2022.list | tee /etc/apt/sources.list.d/mssql-server-2022.list \
+    # Manualy install this dependency because SQL Server is not compatible with the version of libldap in Ubuntu 24.04.
+    && curl -fsSL -O http://archive.ubuntu.com/ubuntu/pool/main/o/openldap/libldap-2.5-0_2.5.11+dfsg-1~exp1ubuntu3_amd64.deb \
```
mthalman, line 12: This concerns me. This is a downgrade of the libldap version that exists in Ubuntu 24.04. What happens when there's a vulnerability in this version? In fact, it does look like this one would be affected by https://ubuntu.com/security/CVE-2023-2953.

cincuranet: If we don't want to override this library manually, I would have to switch this to Ubuntu 22.04, because that's the last supported version for SQL Server. 22.04 has still standard support until Apr 2027, so we should be fine for a while.

22.04 is fine for us. Do you want me to do it?

mthalman: Yes, that sounds like a better plan.

cincuranet: OK

mthalman (approved):

PR #1464, opened 2025-06-23
Fix vulnerability NodeJs (Npm) Security Update for cross-spawn (GHSA-3xgq-45jj-v275) · Issue #240

v-firzha:
Update Nodejs version to the latest in below Docker file:

https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/e73d69edc270e45ed7a76e3d2832210fac67a48e/src/cbl-mariner/2.0/crossdeps/amd64/Dockerfile#L30

20.11.1 -> 22.16.0

### DIFF
--- a/src/almalinux/8/helix/amd64/Dockerfile
+++ b/src/almalinux/8/helix/amd64/Dockerfile
@@ -37,6 +37,7 @@ WORKDIR /home/helixbot
 ENV VIRTUAL_ENV=/home/helixbot/.vsts-env
 
 RUN python -m venv $VIRTUAL_ENV && \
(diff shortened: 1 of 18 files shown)


mthalman:
I see that Node 23 is EOL as of a month ago. Node 24 is an option, but that's a large jump and I won't to minimize the risk of breaking changes. I think a safer option is Node 22 (latest is 22.16.0) which is going to support supported for almost 2 more years which is plenty.

v-firzha:
Got it, I have changed the version to 22.16.0.

mthalman (approved):

PR #1475, opened 2025-07-15
add basic Android test container

wfurt:
This still  needs some work. For one, the current helix queue starts the emulator from serviced and that is not available inside of container. Let me know if there are any suggestions @akoeplinger .

### DIFF
--- a/src/ubuntu/24.04/helix/android/amd64/Dockerfile
+++ b/src/ubuntu/24.04/helix/android/amd64/Dockerfile
@@ -1,0 +1,87 @@
+FROM ubuntu.azurecr.io/ubuntu:noble AS venv
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && \
+    apt-get install -y \
+        cargo \
+        pkg-config \
+        libffi-dev \
+        coreutils \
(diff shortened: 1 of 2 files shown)


meghnave (approved):

src/ubuntu/manifest.json
```diff
@@ -245,6 +245,20 @@
...
         },
+        {
+          "platforms": [
+            {
+              "architecture": "amd64",
+              "dockerfile": "src/ubuntu/24.04/helix/android/amd64",
+              "os": "linux",
+              "osVersion": "noble",
+              "tags": {
+                "ubuntu-24.04-helix-android": {},
```
mthalman, line 256: We only want architecture-specific tags. Other tags without architecture are still around for backwards compatibility.

```suggestion
```

src/ubuntu/24.04/helix/android/amd64/Dockerfile
mthalman, line 1: Rather than duplicating the Helix content, you should just be able to use `mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-24.04-helix-amd64` as the base image.

PR #1493, opened 2025-08-25
Add packages required by AzureCLI AzDO task

NikolaMilosavljevic:
https://github.com/dotnet/sdk/pull/49942 is blocked by this work. (Installer's 8.0 branches need similar change to move them off of CBL-Mariner 2.0 based images.)

SDK build of RPMs utilizes the following step:
https://github.com/dotnet/installer/blob/release/8.0.1xx/eng/common/templates/steps/get-delegation-sas.yml

It requires: `azure-cli` and `powershell` packages.

Same change was made for cbl-mariner 2.0 image with https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1111

### DIFF
--- a/src/azurelinux/3.0/net8.0/fpm/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net8.0/fpm/amd64/Dockerfile
@@ -14,5 +14,8 @@ RUN tdnf install -y \
         util-linux \
         # Provides sudo
         sudo \
+        # Provides functionality for AzureCLI AzDO task
+        azure-cli \
+        powershell \
     && tdnf clean all \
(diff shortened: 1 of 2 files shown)


mthalman (commented):
We might as well update the 10.0 version to include these packages as well so we don't run into this issue when we eventually use that.

NikolaMilosavljevic:
> We might as well update the 10.0 version to include these packages as well so we don't run into this issue when we eventually use that.

I can do that, for completeness. We do not use `fpm` tool anymore for RPM build in 10+. I'm not sure who the consumers of `net10.0` `fpm` images are.

PR #1499, opened 2025-09-09
Add net11.0 AzureLinux images

akoeplinger:
Copy of the net10.0 images so we can start having different contents on net11.0


pavelsavara (approved):

mthalman (commented):
I think I'd prefer that we not introduce new Dockerfiles unless they are necessary. Adding these as separate Dockerfiles significantly increases our agent usage and storage cost because they would be distinct images. So assuming that the Dockerfile content is the same, all that needs to be done is add `net11.0` tags to the existing `net10.0` entries in the manifest. Then if there does occur a need to change the content, we can make a copy at that time.

PR #1520, opened 2025-10-17
Add trixie-based gcc image

jkoritzinsky:
This will provide us with a newer CMake for the dotnet/runtime GCC validation leg for .NET 11

cc: @am11


hoyosjs (approved):

lbussell:
@jkoritzinsky There's an issue with the import image pipeline (https://github.com/dotnet/docker-tools/issues/1847). How soon is this PR needed? Is it urgent?

PR #1535, opened 2025-11-03
Install az CLI for source build

mthalman:
In order to support https://github.com/dotnet/source-build/issues/5342, it's necessary to dynamically generate a SAS token for which there is an Arcade pipeline step for this. But it requires that `az` CLI exist. There are a few images that source build relies upon which do not have that installed.


mthalman:
> If the azure cli is only needed for internal builds, can we use [`AzureCLI@2`](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/azure-cli-v2?view=azure-pipelines) instead?

I'll try that.

ellahathaway (approved):

PR #1584, opened 2026-02-06
Add ninja-build to more images

am11:
Noticed config error on Steve's PR: https://github.com/dotnet/runtime/pull/124041.


am11:
@akoeplinger, @jkoritzinsky, I was wondering if it makes sense to just use `eng/common/native/install-dependencies.sh` in these images (as we already use build-rootfs.sh script from common/cross in crossdeps)? That will allow us to dogfood the script and keep things in sync.

PR #1593, opened 2026-02-24
Add Renovate Dockerfile

mthalman:
Defines a Renovate Dockerfile that is tagged with the Renovate version.

Related to https://github.com/dotnet/arcade/pull/16504

### DIFF
--- a/eng/docker-tools/CHANGELOG.md
+++ b/eng/docker-tools/CHANGELOG.md
@@ -1,0 +1,76 @@
+# Docker Tools / ImageBuilder Changelog
+
+All breaking changes and new features in `eng/docker-tools` will be documented in this file.
+
+---
+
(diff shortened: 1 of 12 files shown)


src/azurelinux/3.0/renovate/amd64/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+        nodejs24 \
+        nodejs24-npm \
+        \
+        # Requirements for Azure DevOps container jobs
+        shadow-utils \
+        sudo \
+        util-linux \
+    && tdnf clean all
+
+RUN npm install -g renovate@${RENOVATE_VERSION}
```
lbussell: **Issue**: How does this work with CFSClean/Network isolation? I think we need to mirror + authenticate to Azure DevOps feeds here.

PR #1594, opened 2026-02-25
Install node-ip on top of re-installed nodejs

premun:
We install the npm package but then later remove node and install a newer one so the package gets lost
https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/1186


src/ubuntu/22.04/amd64/Dockerfile
```diff
@@ -94,6 +92,8 @@ RUN cd ~ && \
     bash nodesource_setup.sh && \
     apt-get install nodejs -y && \
     rm -rf /var/lib/apt/lists/* && \
-    rm -f nodesource_setup.sh
+    rm -f nodesource_setup.sh \
+    # npm installs a vulnerable node-ip pkg. Upgrading the node-ip pkg requires Ubuntu PRO subscription. Workaround by using NPM.
+    && npm install -g ip@latest
```
premun:
```suggestion
    rm -f nodesource_setup.sh && \
    # npm installs a vulnerable node-ip pkg. Upgrading the node-ip pkg requires Ubuntu PRO subscription. Workaround by using NPM.
    npm install -g ip@latest
```

mthalman (approved):

Issue #909, opened 2023-08-09
LANG env var should be changed to en_US.utf-8


jander-msft:
Several of the Dockerfiles seem to be setting LANG environment variable to `en_US.utf8`: https://github.com/search?q=repo%3Adotnet%2Fdotnet-buildtools-prereqs-docker%20en_US.utf8&type=code

The doesn't seem to be correct, because it causes first chance exceptions:
```
[xUnit.net 00:02:29.38]         System.ArgumentException: 'utf8' is not a supported encoding name. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method. (Parameter 'name')
[xUnit.net 00:02:29.38]            at System.Text.EncodingTable.InternalGetCodePageFromName(System.String)
[xUnit.net 00:02:29.38]            at System.Text.EncodingTable.GetCodePageFromName(System.String)
[xUnit.net 00:02:29.38]            at System.Text.Encoding.GetEncoding(System.String)
[xUnit.net 00:02:29.38]            at System.Text.EncodingHelper.GetEncodingFromCharset()
[xUnit.net 00:02:29.38]            at System.ConsolePal.GetConsoleEncoding()
[xUnit.net 00:02:29.38]            at System.Console.get_OutputEncoding()
[xUnit.net 00:02:29.38]            at System.Console.CreateOutputWriter(System.IO.Stream)
[xUnit.net 00:02:29.38]            at System.Console.<get_Out>g__EnsureInitialized|26_0()
[xUnit.net 00:02:29.38]            at System.ConsolePal.EnsureInitializedCore()
[xUnit.net 00:02:29.38]            at System.ConsolePal.Write(Microsoft.Win32.SafeHandles.SafeFileHandle,System.ReadOnlySpan`1[[System.Byte, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]][System.Byte],System.Boolean)
[xUnit.net 00:02:29.38]            at System.Text.Json.Utf8JsonWriter.Flush()
```

According to the `System.Text.EncodingTable` class (https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Text/EncodingData.cs#L57), the value should be `en_US.utf-8` (note the hyphen).

mthalman:
@jander-msft - Is this something you'd like to submit a PR for to fix this?

PR #1607, opened 2026-03-18
Fix utf-8 encoding in LANG env variable

pavel-purma:
Fixes https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/909


akoeplinger:
@mthalman the azurelinux/mariner timeouts don't look related, right?

PR #1634, opened 2026-04-06
Don't install node through apt-get

dkurepa:
https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/1186
we were installing node through apt-get, and even tho we were installing it manully afterwards, the old version was still on the image, getting flagged

### DIFF
--- a/src/ubuntu/22.04/amd64/Dockerfile
+++ b/src/ubuntu/22.04/amd64/Dockerfile
@@ -25,7 +25,6 @@ RUN mkdir -p /usr/lib/local/lib/python3.10/dist-packages/lldb \
 RUN apt-get update \
     && apt-get install -y \
         git \
-        npm \
         zip \
         curl \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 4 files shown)


src/ubuntu/22.04/amd64/Dockerfile
```diff
@@ -78,22 +77,15 @@ RUN apt-get update \
...
-# Remove older version of node & install node 20
-RUN cd /etc/apt/sources.list.d  && \
+# Install Node 20 from NodeSource and use the patched npm ip module.
+RUN cd /etc/apt/sources.list.d && \
     rm -f nodesource.list && \
-    apt --fix-broken install && \
-    apt update && \
-    apt-get remove nodejs -y && \
-    apt-get remove nodejs-doc -y && \
-    apt-get remove libnode-dev -y
```
mthalman, line 88: This removal pattern is also being done in 24.04 and 26.04 Dockerfiles. I would expect all the versions to follow this new fixed pattern.

dkurepa: updated the other images to match 22.04. The CI should also work now as the build image has been restored

mthalman (approved):

PR #1635, opened 2026-04-07
Add nodejs24-npm to AZL build images

akoeplinger:
The -build images were missing the npm package compared to the -crossdeps images.

### DIFF
--- a/src/azurelinux/3.0/net10.0/build/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/build/amd64/Dockerfile
@@ -8,6 +8,7 @@ RUN tdnf install -y \
         curl \
         git \
         nodejs24 \
+        nodejs24-npm \
         powershell \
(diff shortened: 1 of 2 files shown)


mthalman (commented):
What is it that's actually requiring this new dependency to exist?

akoeplinger:
I'd like to use the -build image in aspnetcore instead of ubuntu and it needs npm

mthalman:
Then I'm confused because aspnetcore is already configured to use that image: https://github.com/dotnet/aspnetcore/blob/309d9091856b7bf58719d1aea142d6a60b76e8c2/.azure/pipelines/ci-public.yml#L394

akoeplinger:
That's just for the Linux Musl job. And there's an explicit step to install nodejs via NodeTool@0 right now

mthalman (approved):

wtgodbe (approved):

PR #1650, opened 2026-05-07
Explicitly upgrade packages

dkurepa:
https://github.com/dotnet/dotnet-buildtools-prereqs-docker-internal/issues/1512

### DIFF
--- a/src/azurelinux/3.0/docker-testrunner/Dockerfile
+++ b/src/azurelinux/3.0/docker-testrunner/Dockerfile
@@ -3,7 +3,8 @@
 
 FROM mcr.microsoft.com/azurelinux/base/core:3.0
 
-RUN tdnf install -y \
+RUN tdnf update -y && \
+    tdnf install -y \
         ca-certificates-microsoft \
         # Install Docker
         moby-engine \
--- a/src/azurelinux/3.0/helix/Dockerfile
+++ b/src/azurelinux/3.0/helix/Dockerfile
@@ -18,7 +18,8 @@ RUN python3 -m venv /venv && \
 FROM mcr.microsoft.com/azurelinux/base/core:3.0
 
 # Install .NET and test dependencies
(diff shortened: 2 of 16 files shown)


mthalman (commented):
This isn't the right fix, IMO. We don't even rely on the `libarchive` package. It's installed in the base image. The fact that it's not the latest version and has vulnerabilities means it's the Azure Linux team's responsibility to get it updated.

wfurt:
If the package is in base image and if adding it explicitly here fixes the problem that would suggest we do not update the packages or rebuild the images recently, right?

mthalman:
> If the package is in base image and if adding it explicitly here fixes the problem that would suggest we do not update the packages or rebuild the images recently, right? 

The package is in the base image but with an older version.

One could argue that we could upgrade packages as a first step. That's actually what we do for other distros. For example: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/3467879939ff845bc97ec4ddf2ce772abd729040/src/almalinux/10/helix/amd64/Dockerfile#L3

So that would be a better option that explicitly targeting this package name to be installed/upgraded.

mthalman (commented):
We should make this the default pattern for Azure Linux, in which case https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/azurelinux/4.0/helix/Dockerfile should be updated too.

dkurepa:
> We should make this the default pattern for Azure Linux, in which case https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/main/src/azurelinux/4.0/helix/Dockerfile should be updated too.

Thanks, I added it to all Dockerfiles that don't use an image we already produced, so all that don't use a `mcr.microsoft.com/dotnet-buildtools/prereqs` image as a base

mthalman (approved):

PR #1653, opened 2026-05-13
Retrieve Debian ports package signing key for loongarch64

akoeplinger:
Needed for https://github.com/dotnet/arcade/pull/15374

### DIFF
--- a/src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile
@@ -3,6 +3,9 @@ ARG ROOTFS_DIR=/crossrootfs/riscv64
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net10.0-crossdeps-builder-amd64 AS builder
 ARG ROOTFS_DIR
 
+# temp, will be reverted
+RUN rm -rf /scripts && git clone https://github.com/dotnet/arcade --single-branch --depth 1 -b akoeplinger-patch-1 /scripts
+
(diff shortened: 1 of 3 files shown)


src/azurelinux/3.0/net10.0/cross/riscv64/Dockerfile
```diff
@@ -3,6 +3,9 @@ ARG ROOTFS_DIR=/crossrootfs/riscv64
 FROM mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-net10.0-crossdeps-builder-amd64 AS builder
 ARG ROOTFS_DIR
 
+# temp, will be reverted
+RUN rm -rf /scripts && git clone https://github.com/dotnet/arcade --single-branch --depth 1 -b akoeplinger-patch-1 /scripts
+
 RUN rootfsEnv="/usr/local/share/rootfs" && \
```
am11, line 6: Maybe we should promote it from `noble` to `resolute` (26.04) on line 12 while at it. In the script, we would need to add resolute, its llvm/lldb version is 21 in default packages.

akoeplinger: sure but I'd prefer doing that in a separate change

am11: Ah actually, we should stick to the lowest supported version until 24.04 is old enough (based on its glibc version) that we no longer need to care about it. Otherwise, binaries built on a 26.04 image won’t run on 24.04 (the reverse works). All good 👍

mthalman (approved):

Issue #1646, opened 2026-04-30
Azure Linux 4.0 .NET 11 build images — package blockers


richlander:
## Context

We are porting the full `src/azurelinux/3.0/net11.0/` image graph (29 Dockerfiles) to Azure Linux 4.0. Azure Linux 4.0 is based on Fedora 44, which changes several packages.

Most images can be ported with straightforward package renames (see "Resolved" below), but a few packages are missing entirely from AzL 4.0 repos.

### 1. `debootstrap` — not available in AzL 4.0
- **Used by:** `crossdeps-builder/amd64` (Tier 1 image)
- **Impact:** Blocks all 18 `cross/*` images, plus `android`, `webassembly`, and `crossdeps-llvm` (24 of 29 images)
- **Purpose:** Used by `build-rootfs.sh` to bootstrap Ubuntu/Debian rootfs for cross-compilation
- **Possible fix:** Install from source (it is a shell script), or find an alternative repo

### 2. `libbsd-devel` — not available in AzL 4.0
- **Used by:** `crossdeps-builder/amd64`
- **Impact:** Same as above — blocks the cross-compilation chain
- **Note:** The base `libbsd` runtime package *is* available, but the `-devel` headers are not
- **Previous source:** Was installed from the AzL 3.0 extended repo (`packages.microsoft.com/azurelinux/3.0/prod/extended/...`)

### 3. No AzL 4.0 extended repo
- The extended repo URL pattern `https://packages.microsoft.com/azurelinux/4.0/prod/extended/x86_64/config.repo` returns **404**
- **Impact:** Any package that was only available in the extended repo (e.g., `libbsd-devel`) cannot be installed
- **Used by:** `crossdeps-builder/amd64`

### 4. `powershell` — not available in AzL 4.0
- **Used by:** `crossdeps/amd64`, `build/amd64`
- **Impact:** These images need PowerShell for Azure DevOps functionality
- **Workaround available:** COPY from `mcr.microsoft.com/dotnet/sdk:10.0-azurelinux3.0` (validated locally)

### 5. `mcr.microsoft.com/dotnet/runtime-deps` — no AzL 4.0 variant
- **Used by:** `build/amd64` (as base image: `runtime-deps:10.0-azurelinux3.0`)
- **Impact:** The build image cannot use the standard .NET runtime-deps base
[...]

PR #1657, opened 2026-05-25
Reduce the scope of libbsd requirement

am11:
See #<span/>2 in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1646. Lets build libbsd and its dep libmd from source in images which actually require it.

### DIFF
--- a/src/azurelinux/3.0/net11.0/cross/freebsd/14/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net11.0/cross/freebsd/14/amd64/Dockerfile
@@ -5,11 +5,32 @@ ARG ROOTFS_DIR
 
 # Install packages needed by the FreeBSD bootstrap scripts
 RUN tdnf install -y \
-    awk \
-    m4 && \
+        automake \
+        autoconf \
+        awk \
+        libtool \
+        m4 && \
     # The xz package in Azure Linux 3.0 doesn't provide liblzma.so, so we need to create a symlink to liblzma.so.5
     # so the FreeBSD bootstrap can find it. See https://github.com/microsoft/azurelinux/issues/9217
     ln -s /usr/lib/liblzma.so.5 /usr/lib/liblzma.so
+
+RUN wget -qO- "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x4F3E74F436050C10F5696574B972BF3EA4AE57A3" | \
+        gpg --batch --import && \
+    echo "4F3E74F436050C10F5696574B972BF3EA4AE57A3:6:" | gpg --import-ownertrust
+
+RUN cd /tmp && wget -O libmd-1.2.0.tar.xz "https://libbsd.freedesktop.org/releases/libmd-1.2.0.tar.xz" && \
+    wget -O libmd-1.2.0.tar.xz.asc "https://libbsd.freedesktop.org/releases/libmd-1.2.0.tar.xz.asc" && \
+    gpg --batch --verify libmd-1.2.0.tar.xz.asc libmd-1.2.0.tar.xz && \
+    mkdir libmd && tar -xJf libmd-1.2.0.tar.xz -C libmd --strip-components=1 && \
+    cd libmd && ./configure && make -j install
+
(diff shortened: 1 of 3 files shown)


src/azurelinux/3.0/net11.0/cross/freebsd/14/amd64/Dockerfile
```diff
@@ -5,12 +5,33 @@ ARG ROOTFS_DIR
...
+        awk \
+        libtool \
+        m4 && \
     # The xz package in Azure Linux 3.0 doesn't provide liblzma.so, so we need to create a symlink to liblzma.so.5
     # so the FreeBSD bootstrap can find it. See https://github.com/microsoft/azurelinux/issues/9217
     ln -s /usr/lib/liblzma.so.5 /usr/lib/liblzma.so
 
+RUN curl -fsSL "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x4F3E74F436050C10F5696574B972BF3EA4AE57A3" | \
+        gpg --batch --import && \
+    echo "4F3E74F436050C10F5696574B972BF3EA4AE57A3:6:" | gpg --import-ownertrust
```
akoeplinger, line 19: we're using wget instead of curl to fetch from keyserver.ubuntu.com in other places, any reason you're using curl here?

src/azurelinux/3.0/net11.0/cross/freebsd/14/amd64/Dockerfile
```diff
@@ -5,12 +5,33 @@ ARG ROOTFS_DIR
...
+
+RUN cd /tmp && curl -fLo libmd-1.2.0.tar.xz "https://archive.hadrons.org/software/libmd/libmd-1.2.0.tar.xz" && \
+    curl -fLo libmd-1.2.0.tar.xz.asc "https://archive.hadrons.org/software/libmd/libmd-1.2.0.tar.xz.asc" && \
```
akoeplinger: we can use these URLs for consistency:
- https://libbsd.freedesktop.org/releases/libmd-1.2.0.tar.xz
- https://libbsd.freedesktop.org/releases/libmd-1.2.0.tar.xz.asc

src/azurelinux/3.0/net11.0/cross/freebsd/14/amd64/Dockerfile
```diff
@@ -5,12 +5,33 @@ ARG ROOTFS_DIR
...
+    gpg --batch --verify libmd-1.2.0.tar.xz.asc libmd-1.2.0.tar.xz && \
+    mkdir libmd && tar -xJf libmd-1.2.0.tar.xz -C libmd --strip-components=1 && \
+    cd libmd && ./configure && make -j install
+
+RUN cd /tmp && curl -fLo libbsd-0.12.2.tar.xz "https://libbsd.freedesktop.org/releases/libbsd-0.12.2.tar.xz" && \
+    curl -fLo libbsd-0.12.2.tar.xz.asc "https://libbsd.freedesktop.org/releases/libbsd-0.12.2.tar.xz.asc" && \
```
akoeplinger: I've just checked and both libbsd and libmd are part of AZL4 now: https://github.com/microsoft/azurelinux/tree/4.0/specs/l/libbsd

I wonder if we even need this now?

am11: I think it makes sense to scope it to where it's needed. If only FreeBSD needs this dependency, as we can tell from the green CI, then only it should install it. We can change the method of installation when the image is available https://mcr.microsoft.com/en-us/artifact/mar/azurelinux/base/core/tags. Right now, this PR fixes #<span/>2 and #<span/>3 from https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1646.

These steps took less than 20sec to build and install (combined), so that's also not an issue.

akoeplinger: right, I'm not concerned about the time, mostly about duplication/supply-chain risks so it'd be good to use upstream repo packages where possible. I just saw that we might be missing the -devel packages. We can try again when AZL4 is more widely available.

akoeplinger (approved):

PR #1666, opened 2026-06-08
Obtain keyring signatures from the keyring tarball

premun:
- Switched the Ubuntu keyring tarball download from https://mirrors.edge.kernel.org/... to http://archive.ubuntu.com/..., keeping the existing pinned SHA256 and signed SHA512SUMS.txt.asc verification.
- Removed both live keyserver.ubuntu.com fetches.
- Installed ubuntu-archive-removed-keys.gpg directly from the already verified ubuntu-keyring tarball.
- Appended the removed-keyring contents into ubuntu-archive-keyring.gpg so Arcade’s single --keyring /usr/share/keyrings/ubuntu-archive-keyring.gpg path can still verify older Ubuntu releases like xenial.

### DIFF
--- a/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
@@ -57,7 +57,7 @@ COPY dimitri_john_ledkov.asc .
 RUN gpg --import dimitri_john_ledkov.asc && \
     rm dimitri_john_ledkov.asc && \
 # 2. Download the ubuntu keyrings
-    wget https://mirrors.edge.kernel.org/ubuntu/pool/main/u/ubuntu-keyring/ubuntu-keyring_2023.11.28.1.tar.xz && \
+    wget http://archive.ubuntu.com/ubuntu/pool/main/u/ubuntu-keyring/ubuntu-keyring_2023.11.28.1.tar.xz && \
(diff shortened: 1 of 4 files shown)


dkurepa:
why are we making this change? I thought the only problem with this was network isolation, and this change will still need it off

oleksandr-didyk (approved):

src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
```diff
@@ -57,21 +57,20 @@ COPY dimitri_john_ledkov.asc .
 RUN gpg --import dimitri_john_ledkov.asc && \
     rm dimitri_john_ledkov.asc && \
 # 2. Download the ubuntu keyrings
-    wget https://mirrors.edge.kernel.org/ubuntu/pool/main/u/ubuntu-keyring/ubuntu-keyring_2023.11.28.1.tar.xz && \
+    wget http://archive.ubuntu.com/ubuntu/pool/main/u/ubuntu-keyring/ubuntu-keyring_2023.11.28.1.tar.xz && \
```
dkurepa, line 60: should it be https?

premun: I think it can't as you'd need to initialize the whole trust chain

akoeplinger: it can (and should) since the SSL certs are unrelated to the keyring.

PR #1686, opened 2026-06-18
Continue CI if some builds fail

dkurepa:
`--extra-index-url` is not allowed by a CFS policy and I don't think we even need it because of the `--no-deps` we provide.

### DIFF
--- a/eng/docker-tools/templates/stages/build-and-test.yml
+++ b/eng/docker-tools/templates/stages/build-and-test.yml
@@ -54,7 +54,7 @@ parameters:
 ################################################################################
 stages:
 - stage: Build
-  condition: and(succeeded(), contains(variables['stages'], 'build'))
+  condition: and(succeededOrFailed(), contains(variables['stages'], 'build'))
   dependsOn: []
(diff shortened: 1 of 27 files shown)


michalpavelka (approved):

premun (approved):

eng/docker-tools/templates/stages/build-and-test.yml
```diff
@@ -54,7 +54,7 @@ parameters:
 ################################################################################
 stages:
 - stage: Build
-  condition: and(succeeded(), contains(variables['stages'], 'build'))
+  condition: and(succeededOrFailed(), contains(variables['stages'], 'build'))
```
mthalman, line 57: Changes to `eng/docker-tools` will be overwritten by automation PRs like https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1678.

dkurepa: Thanks, we'll make the change in docker-tools if this ends up working

PR #1682, opened 2026-06-16
Install Microsoft.AITools.BinlogMcp from dotnet-public

YuliiaKovalova:
The `Microsoft.AITools.BinlogMcp` package is now published on [nuget.org](https://www.nuget.org/packages/Microsoft.AITools.BinlogMcp) as stable `1.0.0`, so the `binlog-mcp` container image no longer needs the dnceng AzDO `dotnet-tools` feed.

## Changes
- **src/azurelinux/3.0/binlog-mcp/amd64/Dockerfile** — pin `BINLOG_MCP_VERSION` to stable `1.0.0` and remove the `DOTNET_TOOLS_FEED` ARG and `--add-source` (nuget.org is the default source). The previously pinned `1.0.0-preview.*` version only exists on the AzDO feed.
- **eng/renovate.json** — point the `Microsoft.AITools.BinlogMcp` `registryUrlTemplate` to `https://api.nuget.org/v3/index.json` and drop the now-obsolete pre-release allowance packageRule (stable versions are now published).

PR #1692, opened 2026-06-22
Fix binlog-mcp network isolation: replace nuget.org source with dotnet-public

YuliiaKovalova:
## Problem
The `binlog-mcp` builder stage installs `Microsoft.AITools.BinlogMcp` using the `mcr.microsoft.com/dotnet/sdk:10.0-azurelinux3.0` image. That image ships a default `NuGet.Config` whose **only** source is `nuget.org` (`https://api.nuget.org/v3/index.json`), which is unreachable from the network-isolated build environment.

PR #1682 switched to `--add-source ${DOTNET_TOOLS_FEED}`, but `--add-source` *adds* the dotnet-public feed **alongside** nuget.org rather than replacing it. `dotnet tool install` still attempts to reach nuget.org during restore, which fails under network isolation.

## Fix
Before installing the tool, remove the default nuget.org source and add the dnceng `dotnet-public` feed (a nuget.org mirror reachable in isolated builds):

```dockerfile
RUN dotnet nuget remove source nuget.org && \
    dotnet nuget add source ${DOTNET_TOOLS_FEED} --name dotnet-public
```

No version or feed-URL changes — `DOTNET_TOOLS_FEED` already points at dotnet-public from #1682.

### DIFF
--- a/src/azurelinux/3.0/binlog-mcp/amd64/Dockerfile
+++ b/src/azurelinux/3.0/binlog-mcp/amd64/Dockerfile
@@ -7,7 +7,7 @@
 #             mcr.microsoft.com/dotnet-buildtools/prereqs:azurelinux-3.0-binlog-mcp-amd64
 
 ARG BINLOG_MCP_VERSION=1.0.0
-ARG DOTNET_TOOLS_FEED=https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json
+ARG DOTNET_PUBLIC_FEED=https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json
 
 # Stage 1: Use the .NET SDK to install the tool. The SDK is required because
 # `dotnet tool install` is not available in distroless runtime images.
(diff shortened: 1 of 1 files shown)


premun:
FYI there's also `--source` instead of `--add-source`. So maybe you don't need to call `dotnet nuget ...`

src/azurelinux/3.0/binlog-mcp/amd64/Dockerfile
```diff
@@ -18,11 +18,16 @@ FROM mcr.microsoft.com/dotnet/sdk:10.0-azurelinux3.0 AS builder
 ARG BINLOG_MCP_VERSION
 ARG DOTNET_TOOLS_FEED
 
+# The SDK image ships a default NuGet.Config whose only source is nuget.org,
+# which is unreachable from the network-isolated build environment. Replace it
+# with the dnceng dotnet-public feed (a nuget.org mirror) before restoring.
+RUN dotnet nuget remove source nuget.org && \
+    dotnet nuget add source ${DOTNET_TOOLS_FEED} --name dotnet-public
```
dkurepa: could we rename the arg since it's no longer pointing to the tools feed?

YuliiaKovalova: Done — renamed to \DOTNET_PUBLIC_FEED\ in 25e904bf.

YuliiaKovalova:
Good call — switched to \dotnet tool install --source \\ (replaces all configured sources), so the separate \dotnet nuget remove/add\ is no longer needed. Updated in 25e904bf.

dkurepa (approved):

oleksandr-didyk (approved):

oleksandr-didyk:
The package seems to have been missing from the feed. Ran the migration pipeline, should be there now - https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-public/NuGet/Microsoft.AITools.BinlogMcp/overview/1.0.0

oleksandr-didyk (approved):

premun:
You might need to wait for the quarantine if it's been uploaded recently

oleksandr-didyk:
> You might need to wait for the quarantine if it's been uploaded recently

Quarantine is for packages that have just been published to public NuGet, right? If the package already is in `dotnet-public` it should not be quarantined. Or am I miss-understanding the loop here

PR #1696, opened 2026-06-25
Update Ubuntu mirrors to allowed ones

### LINKED ISSUE #1693: Use `https://azure.archive.ubuntu.com/`
We need to use `https://azure.archive.ubuntu.com/` as the package source for ubuntu packages.

We definitely need to have it in internal builds, but it'd be good if we had it in public ones too for better validation

premun:
Resolves #1693

### DIFF
--- a/src/debian/12/gcc13/amd64/Dockerfile
+++ b/src/debian/12/gcc13/amd64/Dockerfile
@@ -1,5 +1,8 @@
 FROM library/gcc:13-bookworm
 
+# Use the approved Debian archive mirror (per Network Isolation team guidance).
+RUN find /etc/apt -type f \( -name '*.sources' -o -name '*.list' \) -exec sed -i 's|//deb\.debian\.org|//debian-archive.trafficmanager.net|g' {} +
+
 # Dependencies for dotnet/runtime native components.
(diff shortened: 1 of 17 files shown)


oleksandr-didyk (approved):

akoeplinger:
this will probably not be enough since we also use Ubuntu mirrors when building the rootfs via https://github.com/dotnet/arcade/blob/b076228a542025c4f879f254d38adb5cf34a2475/eng/common/cross/build-rootfs.sh#L228

the script also hits other endpoints (grep for `https`)

oleksandr-didyk (approved):

oleksandr-didyk:
> this will probably not be enough since we also use Ubuntu mirrors when building the rootfs via [dotnet/arcade@`b076228`/eng/common/cross/build-rootfs.sh#L228](https://github.com/dotnet/arcade/blob/b076228a542025c4f879f254d38adb5cf34a2475/eng/common/cross/build-rootfs.sh#L228)
> 
> the script also hits other endpoints (grep for `https`)

Thanks for pointing it out, I'll take a look and try to get at least the Ubuntu links updated in Arcade

PR #1715, opened 2026-07-22
Pin Arcade scripts and automate updates

sbomer:
## Summary

- pin rootfs-building Dockerfiles to an exact Arcade commit
- fetch only the pinned commit so image builds are reproducible
- configure Renovate to update and group the Arcade pins

## Testing

- `pwsh ./run-tests.ps1`
- Renovate 43 configuration validation
- exact shallow fetch and checkout of the pinned Arcade commit

*This content was created with assistance from AI.*

### DIFF
--- a/eng/renovate.json
+++ b/eng/renovate.json
@@ -96,9 +96,28 @@
       "depNameTemplate": "Microsoft.AITools.BinlogMcp",
       "datasourceTemplate": "nuget",
       "registryUrlTemplate": "https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json"
+    },
+    {
+      "description": "Updates the Arcade commit used by rootfs-building Dockerfiles",
+      "customType": "regex",
+      "managerFilePatterns": [
+        "/(^|/)Dockerfile$/"
+      ],
+      "matchStrings": [
+        "ARG ARCADE_COMMIT=(?<currentValue>[a-f0-9]{40})"
+      ],
+      "depNameTemplate": "dotnet/arcade",
+      "datasourceTemplate": "github-commits"
     }
   ],
   "packageRules": [
+    {
+      "description": "Group Arcade commit updates across all Dockerfiles",
+      "matchDepNames": [
+        "dotnet/arcade"
+      ],
+      "groupName": "Arcade"
+    },
     {
       "description": "Only allow minor/patch updates for LLVM to stay on the respective major version",
       "matchDepNames": [
--- a/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
+++ b/src/azurelinux/3.0/net10.0/crossdeps-builder/amd64/Dockerfile
@@ -116,8 +116,11 @@ RUN \
 ENV PATH="/opt/llvm/bin:$PATH"
 
 # Obtain arcade scripts used to build rootfs
+ARG ARCADE_COMMIT=e5f974ba880c3da69932191c665e90ee64e8da26
 RUN git config --global user.email builder@dotnet-buildtools-prereqs-docker && \
-    git clone --depth 1 --single-branch https://github.com/dotnet/arcade /scripts
+    git init /scripts && \
+    git -C /scripts fetch --depth 1 https://github.com/dotnet/arcade ${ARCADE_COMMIT} && \
+    git -C /scripts checkout --detach FETCH_HEAD
 
 # Clear PIP_INDEX_URL since it contains a short-lived token that should not persist in the final image
 ENV PIP_INDEX_URL=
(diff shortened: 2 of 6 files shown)


akoeplinger:
Fixes https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1312? I sort of wonder whether we should just move build-rootfs.sh and friends into this repo and have some way of syncing its source back to arcade given this repo is the main consumer and it'd be nice to be able to make changes without needing to go through arcade

/cc @am11

akoeplinger (approved):

am11:
Do we need to pin by commit or release branch? It feels commit is a bit set-in-stone than a release branch.

> Fixes #1312? I sort of wonder whether we should just move build-rootfs.sh and friends into this repo and have some way of syncing its source back to arcade given this repo is the main consumer and it'd be nice to be able to make changes without needing to go through arcade

Makes sense to me, I think we would just need to update docs in runtime/diagnostics repos (`git grep` for the rescue) and ping concerned folks (mostly contributors of this script). I think that'd be much better than having to work with one-off commit hash.

sbomer:
> Fixes https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1312?

Not exactly - this won't reduce the redundant work that's done when we touch the arcade layer. But using a stable version of arcade would make it possible to work around caching issues by moving the clone to the cross image builds.

> I sort of wonder whether we should just move build-rootfs.sh and friends into this repo

I like it, not sure what's the best way to flow the changes. I still think this is an improvement in the meantime.

> Do we need to pin by commit or release branch? It feels commit is a bit set-in-stone than a release branch.

My goal is to make the builds more deterministic, so it needs to reference a non-moving target.

akoeplinger:
we do have a github workflow in aspnetcore/runtime that is used to sync some source files, we could add something similar.

I'll get this in for now so we can finally get the AZL4.0 images in too and don't need to constantly keep that PR updated.

am11:
> My goal is to make the builds more deterministic, so it needs to reference a non-moving target.

Docker image _is_ the deterministic snapshot. This is not a unique problem. All arcade powered repos have this requirement, they solved it by inventing eng/common dir which gets sync by codeflow.

am11:
@akoeplinger, IMO, best way is to make it arcade powered repo again (revert parts of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1553) then `COPY eng/common/cross/ /scripts` in dockerfile. This will align with a familiar workflow we have been using in runtime/diagnostics for years:

* Normal circumstances: Make change in arcade, wait for it to flow to the runtime/diagnostics repo.
* Urgent circumstances: Make change in arcade and the target repo in parallel, merge them simultaneously

I think this covers the requirement here without inventing a new approach.

akoeplinger:
@am11 almost none of the files in eng/common except the eng/common/cross are relevant in this repo so there'd be quite some churn if we sync the whole dir again.

We could also do the custom source sync in the other direction, so source of truth still lives in arcade and we'd still have the option of easy testing and urgent override here.

am11:
> @am11 almost none of the files in eng/common except the eng/common/cross are relevant in this repo so there'd be quite some churn if we sync the whole dir again.

None of the consumer repos use 100% of eng/common, but that's the most prominent method. Why reinvent a wheel for few kilobytes which dozen other repos don't care to optimize?

akoeplinger:
It's not about the size of the files but the number of effectively no-op PRs to this repo. But ultimately it's up to the owners of the repo to decide, I'm just providing alternatives :)

premun:
I would lean towards @am11's suggestion. If it becomes noisy, we tone down the frequency and/or trigger on-demand.

It would be good to isolate the breakages caused by the rootfs scripts into PRs of this repo.

premun:
> IMO, best way is to make it arcade powered repo again (revert parts of https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1553)

Are you suggesting we use the `eng/common` which would come with `docker-tools`? Or we just depend on Arcade directly and get updates directly?

am11:
> > IMO, best way is to make it arcade powered repo again (revert parts of #1553)
> 
> Are you suggesting we use the `eng/common` which would come with `docker-tools`? Or we just depend on Arcade directly and get updates directly?

Direct arcade dependency would be better than any type of indirect approach. It's clearer to be able to modify files sitting in the same repo and test in CI etc.

premun:
Both would be direct in this regard - either you'd have `eng/common/...` or `/eng/docker-tools/eng/common/...`.

The difference is whether we'd first flow Arcade to `docker-tools` and then that would get flown here.
Depends on the coupling between those repos and the scripts (I do not understand if there's any impact of the rootfs scripts on docker-tools).

So one way is faster and more direct, the other has an extra hop which might go through more validation in `docker-tools` (if that even happens).

Probably the more direct one is better?

am11:
`eng/common` is the direct one by convention across all arcade repos, that’s why I think it’s better.

PR #1716, opened 2026-07-23
Arcadify the repo and use `eng/common` directly instead of cloning Arcade

premun:
Based on a discussion in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1715


premun:
cc @am11

akoeplinger:
we also need to update the new AZL4.0 images

premun:
@copilot+gpt-5.6-sol can you also bring this change to the Azure Linux 4.0 images?

akoeplinger (approved):

akoeplinger (approved):

akoeplinger:
looks like the CI machine ran out of disk space, but we're already using the XL pool. we might need to just opportunistically merge this.

premun:
Retry ended the same... But somehow I refuse to believe the extra image with `eng/common` got us over the disk limits..

am11:
Attempt 2 also hit the same machine `XL 3`. Maybe try closing/reopening the PR for better shuffling? (or push a dummy change)

lbussell:
Consider turning on System.Debug in the pipeline to see disk usage during the build.

akoeplinger:
isn't it just because we're basically rebuilding all images now and that means it can't reuse the cache?

am11:
Ah right.  We can split it in separate PRs for each netXX dir. https://github.github.com/gh-stack/ can be used for this (see in action e.g. https://github.com/dotnet/runtime/pull/131507)

premun:
So I guess we can just force-merge?

akoeplinger:
> So I guess we can just force-merge?

question is if the internal builders won't hit the same issue. I don't know if they have larger disk space than the public XL pool

premun:
I'd say they will in the first iteration but then those that make it get published and then hopefully the next one gets it right.

I wonder if we could also add the new COPY higher in the image to not rebuild from bottom that often though.

Also I think we should change our infra to be able to rebuild everything.
If we start getting new eng/common updates every week, this will keep popping up.

premun:
Oh, it's already on top

akoeplinger:
> If we start getting new eng/common updates every week, this will keep popping up.

I _think_ only when we get changes in eng/common/cross right?
but I agree that it'd be good if we could rebuild everything

premun:
Let's try

PR #1717, opened 2026-07-23
Use Microsoft public npm proxy when NPM_TOKEN is not set

premun:
## Problem

Every Dockerfile that installs npm packages configures the internal `dotnet-public-npm` Azure Artifacts feed when `NPM_TOKEN` is set, and otherwise falls through to npm's default registry (`registry.npmjs.org`). Dev machines can no longer reach npmjs.org (it's been forbidden), so image builds that run `npm install` fail on the no-token fallback path with an SSL/TLS handshake failure.

## Fix

Add an explicit `else` branch to every Dockerfile that configures npm so that, when `NPM_TOKEN` is **not** provided, the registry is pointed at the Microsoft public npm proxy `https://packagefeedproxy.microsoft.io/npm/` instead of leaving npm's default. The internal-token path is unchanged, so official/internal builds still use the authenticated `dotnet-public-npm` feed.

### Files changed (13)

- **Linux (11):** ubuntu 22.04 (amd64, coredeps, helix), 24.04 (+helix), 26.04 (+helix); azurelinux net8.0/net9.0/net10.0 webassembly; azurelinux renovate.
- **Windows (2):** windowsservercore ltsc2022 & ltsc2025 helix webassembly (`else` now sets the proxy registry instead of just echoing).

Accompanying comments were updated to reflect the proxy fallback.

## Validation

- Reproduced the failure: `docker build src/azurelinux/3.0/net8.0/webassembly/amd64` with no token failed at `npm i -g typescript` -> `registry.npmjs.org` SSL handshake failure.
- After the fix, rebuilt the same image (no token) -> **succeeded**, and `npm config get registry` inside the image returns `https://packagefeedproxy.microsoft.io/npm/`.

> Note: only a Linux Docker engine was available, so the two Windows Dockerfile changes were made by matching the existing pattern but not build-tested.

### DIFF
--- a/eng/pipelines/steps/pip-authenticate.yml
+++ b/eng/pipelines/steps/pip-authenticate.yml
@@ -16,10 +16,11 @@ steps:
     # rather than appending a duplicate --build-arg.
     $options = "$(imageBuilderBuildArgs)"
     $options = $options -replace '--build-arg PIP_INDEX_URL=\S+', "--build-arg PIP_INDEX_URL=$env:PIP_INDEX_URL"
-    # The npm registry (public dotnet-public-npm feed) is baked into the Dockerfiles as the
-    # NPM_REGISTRY default and is read anonymously, so no override is needed here. Internal
-    # builds additionally pass a token so they can ingest new package versions; the token is
-    # optional and unused on anonymous reads.
+    # Dockerfiles default NPM_REGISTRY to the Microsoft public npm proxy (for public/local
+    # builds). On internal builds, point npm at the authenticated dotnet-public-npm feed and
+    # provide the token so the auth key derived from the URL resolves. Keeping the internal
+    # feed URL here means it lives in exactly one place instead of every Dockerfile.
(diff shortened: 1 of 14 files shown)


akoeplinger:
IMO we should just use `dotnet-public-npm` as the fallback, it is publicly available (you just can't ingest new upstream package versions if you're not authenticated)

premun:
> IMO we should just use `dotnet-public-npm` as the fallback, it is publicly available (you just can't ingest new upstream package versions if you're not authenticated)


We discussed this and since there are still some places which install latest available versions of a given package, it might cause intermittent failures in public builds since the unauthenticated requests won't be able to ingest those versions from the public NPM. So I think we can merge as-is as it's just a gradual improvement for the dev UX in this repo

pavel-purma (approved):

oleksandr-didyk (approved):

akoeplinger:
@premun I think this missed updating the AZL4 images

premun:
@copilot+claude-opus-4.8 can you open a PR that brings this change to the Azure Linux 4.0 images in this repo?

PR #1723, opened 2026-07-30
Use different mirror for loongarch64 builds

akoeplinger:
Same approach as we use for arm Debian images: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/65c889c5dc751aa6b05edd4fd5d8e27566275b6a/src/debian/13/helix/Dockerfile#L8-L17

Fixes https://github.com/dotnet/dotnet-docker-internal/issues/11705


oleksandr-didyk (approved):

oleksandr-didyk:
Merged on red since the underlying issue of the PR build getting stuck should get fixed by a green re-build / publish in internal pipeline

Issue #1526, opened 2025-10-21
Build break w/sharing Dockerfiles and only differentiating by arg


MichaelSimons:
The Dockerfiles added in https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1508 are causing build breaks during publishing.  The issue is in ImageBuilder and limitations when sharing Dockerfiles and only differentiating by args.   The tracking issue for this is https://github.com/dotnet/docker-tools/issues/1826.

PR #1730, opened 2026-08-05
Add CI images for dotnet/vscode-csharp integration tests

dibarbet:
Integration tests in vscode-csharp that run on docker images currently have to [install a few additional dependencies](https://github.com/dotnet/vscode-csharp/blob/main/azure-pipelines/test-linux-docker-prereqs.yml) in order to run vscode.   Currently they are installed as part of the pipeline which is usually fast, but can sometimes take [~30 minutes](https://dev.azure.com/dnceng-public/public/_build/results?buildId=1539620&view=logs&j=46c07da2-fe5a-5fce-2ff0-f6ed1ae45ae8&t=79e719ce-129e-50b6-1590-65e5a787a56d) when the repository is slow.

Instead, we can create a docker image that we can re-use that already has the dependencies installed.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -38,6 +38,7 @@ src/ubuntu/22.04/mlnet/ @dotnet/runtime-infrastructure @dotnet/dotnet-docker-rev
 src/ubuntu/22.04/opt/arm64v8/ @dotnet/runtime-infrastructure @dotnet/dotnet-docker-reviewers
 src/ubuntu/22.04/coredeps/amd64/ @dotnet/runtime-infrastructure @dotnet/dotnet-docker-reviewers
 src/ubuntu/22.04/amd64/ @dotnet/source-build @dotnet/dotnet-docker-reviewers
+src/ubuntu/**/vscode-csharp/ @dotnet/roslyn-ide @dotnet/dotnet-docker-reviewers
 src/ubuntu/*/Dockerfile @dotnet/source-build @dotnet/dotnet-docker-reviewers
 
 # common paths
--- a/src/ubuntu/manifest.json
+++ b/src/ubuntu/manifest.json
@@ -259,6 +259,57 @@
           ]
         },
         {
+          "productVersion": "8.0",
+          "platforms": [
+            {
+              "architecture": "amd64",
+              "dockerfile": "src/ubuntu/24.04/vscode-csharp/amd64",
+              "os": "linux",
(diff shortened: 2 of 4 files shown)


JoeRobich (approved):

lbussell:
@dibarbet what's the size of the image (either on disk, or compressed) after it's built? Just curious.

dibarbet:
> @dibarbet what's the size of the image (either on disk, or compressed) after it's built? Just curious.

Locally I'm seeing each image around 1.3 GB disk size - with the incremental usage compared to the base image being ~460-490 MB

dibarbet:
@dotnet/dotnet-docker-reviewers  could use a review here, happy to give more details if needed

src/ubuntu/manifest.json
```diff
@@ -258,6 +258,57 @@
             }
           ]
         },
+        {
+          "productVersion": "8.0",
```
lbussell, line 262: Is `productVersion` needed here? Not sure if there's a reason you added it.

CODEOWNERS
```diff
@@ -38,6 +38,7 @@ src/ubuntu/22.04/mlnet/ @dotnet/runtime-infrastructure @dotnet/dotnet-docker-rev
 src/ubuntu/22.04/opt/arm64v8/ @dotnet/runtime-infrastructure @dotnet/dotnet-docker-reviewers
 src/ubuntu/22.04/coredeps/amd64/ @dotnet/runtime-infrastructure @dotnet/dotnet-docker-reviewers
 src/ubuntu/22.04/amd64/ @dotnet/source-build @dotnet/dotnet-docker-reviewers
+src/ubuntu/**/vscode-csharp/ @dotnet/roslyn-ide @dotnet/dotnet-docker-reviewers
```
lbussell, line 41: @dotnet/prodconsvcs, I think the dotnet/roslyn-ide needs to be given write access to the repo.

lbussell (approved):

src/ubuntu/manifest.json
dibarbet: I'm not the most familiar with this repo, but copilot and this issue https://github.com/dotnet/dotnet-buildtools-prereqs-docker/issues/1526 seem to indicate that it is needed (due to the images only differing by the build args).

CODEOWNERS
premun: Done

PR #1733, opened 2026-08-10
Install missing packages in azl 4.0 crossdeps

am11:
`build.sh` on all arcade repos is failing:

```sh
tar (child): gzip: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
```

It wasn't needed in 3.0 images because base system had it installed.

---

3.0 installed GNU toolchain at `x86_64-pc-linux-gnu`, which is recognized by clang but 4.0 install it at `x86_64-azurelinux-linux`, which isn't recognized.

---

Install missing C++ toolchain.

### DIFF
--- a/src/azurelinux/4.0/net11.0/crossdeps/amd64/Dockerfile
+++ b/src/azurelinux/4.0/net11.0/crossdeps/amd64/Dockerfile
@@ -10,7 +10,6 @@ RUN dnf update -y && \
         # Provides 'su', required by Azure DevOps
         ca-certificates \
         git \
-        pigz \
         shadow-utils \
         util-linux \
         wget2-wget \
@@ -22,8 +21,10 @@ RUN dnf update -y && \
         make \
         cmake \
         diffutils \
+        gzip \
         icu \
         ninja-build \
+        pigz \
         tar \
         # Crosscomponents build dependencies
         glibc-devel \
--- a/src/azurelinux/4.0/net11.0/crossdeps/amd64/Dockerfile
+++ b/src/azurelinux/4.0/net11.0/crossdeps/amd64/Dockerfile
@@ -18,6 +18,7 @@ RUN dnf update -y && \
         cpio \
         file \
         gcc \
+        gcc-c++ \
         make \
         cmake \
         diffutils \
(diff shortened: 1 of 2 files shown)


premun (approved):

premun:
@am11 I queued a full rebuild of all images, should be done tonight

src/azurelinux/4.0/net11.0/crossdeps/amd64/Dockerfile
```diff
@@ -38,6 +40,12 @@ RUN dnf update -y && \
         azure-cli \
     && dnf clean all
 
+# GNU toolchain is installed in 'x86_64-azurelinux-linux' directory which is not recognized by
+# clang (https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChains/Gnu.cpp),
+# so we create a symlink with -pc- triplet.
```
akoeplinger, line 45: @am11 do you have a simple repro for this? I wonder if we should file a bug with azurelinux

akoeplinger: I think we might need to fix this in the clang we build instead.

If I install clang through dnf then it finds the GCC location:
```bash
bash-5.3# clang -v
clang version 21.1.8 (Microsoft Corporation 21.1.8-8.azl4)
Target: aarch64-azurelinux-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Configuration file: /etc/clang/aarch64-azurelinux-linux-gnu-clang.cfg
System configuration file directory: /etc/clang/
Found candidate GCC installation: /usr/bin/../lib/gcc/aarch64-azurelinux-linux/15
Selected GCC installation: /usr/bin/../lib/gcc/aarch64-azurelinux-linux/15
Candidate multilib: .;@m64
Selected multilib: .;@m64
```

am11: before: https://github.com/am11/CrossRepoCITesting/actions/runs/31414507086/job/93540201703
after (with new images): https://github.com/am11/CrossRepoCITesting/actions/runs/31487863070

Note that "before" failed after 31m and it wasn't the first native component build which failed. It failed when the build compiles JIT component for host RID. Every cross build (regardless of official/community) compiles some tools for host which requires host JIT.

> I think we might need to fix this in the clang we build instead.

For runtime/VMR, that's enough to have a symlink, but yes that would be an idiomatic fix: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1734.

PR #1734, opened 2026-08-11
Use branded triplet in llvm build on AZL 4.0

am11:
Removed the workaround and match packages' approach to use branded triplet, per https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/1733#discussion_r3758438232.

### DIFF
--- a/src/azurelinux/4.0/net11.0/crossdeps-builder/amd64/Dockerfile
+++ b/src/azurelinux/4.0/net11.0/crossdeps-builder/amd64/Dockerfile
@@ -119,6 +119,7 @@ RUN \
         -DCMAKE_BUILD_TYPE=Release \
         -DCMAKE_C_COMPILER=clang \
         -DCMAKE_CXX_COMPILER=clang++ \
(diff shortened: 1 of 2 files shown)


am11:
cc @akoeplinger

akoeplinger (approved):
Thank you!

am11:
@akoeplinger, unfortunately, this is not enough. The symlink approach is much simpler than the "matching distro" approach. The symlink is one line vs. this [microsoft/azurelinux@f394783](https://github.com/microsoft/azurelinux/commit/f39478361e6ce8dbab05213e6852be812404d819). They manually create `/etc/clang/{triple}.cfg` (which you can check after `dnf install`, as you did before), then point the LLVM build to use that config file. In our case, that would mean `echo .... > /etc/clang/{triple}.cfg` followed by `COPY` in the other layer, which is a bit more work than the symlink. It's all distro-specific branding that has zero bearing on the product. Honestly, it doesn't make much sense to follow distro branding here; if anything, it should be `dotnet` branding since this is our custom llvm toolchain build.

PR #1738, opened 2026-08-18
Automate .NET preview container tag promotion

mthalman:
## Description

The `11.0-preview` .NET container tags used by these images are at risk of becoming stale after the corresponding `11.0` stable tags become available.

This change configures Renovate to detect .NET preview tags in manifests and Dockerfiles and promote them to stable tags within the same major version.

## Validation

- validated the configuration with Renovate 43.288.0
- ran a full repository dry run against MCR
- verified promotion, suffix preservation, namespace matching, missing stable tags, and cross-major filtering using a local container registry


lbussell (approved):
Cool!

PR #1744, opened 2026-09-17
Use release .NET 11 SDK tag for vscode-csharp CI

dibarbet:
vscode-csharp CI needs to upgrade integration tests to run on .NET rc instead of preview, so switching our image over to use the 11.0-resolute tag


dibarbet:
@premun would you be the right person to look at this change?

mthalman (approved):

PR #1745, opened 2026-09-17
Fix the Android Helix image and parameterize the API level

wfurt:
## Problem

The published `ubuntu-24.04-helix-android-amd64` image cannot start an emulator.

`avdmanager` runs as root with `HOME=/root`, so the AVD is created under `/root/.android/avd`, while `/home/helixbot/.android/avd` receives only a 3 GB `sdcard.img`. No `ANDROID_*` variables are set either. As `helixbot`, which is the user the container runs as, `emulator -list-avds` returns nothing and a boot fails:

```
ERROR | Unknown AVD name [android_emu_86_64], use -list-avds to see valid list.
ERROR | HOME is defined but there is no file android_emu_86_64.ini in $HOME/.android/avd
```

Nothing references this image tag today, so these are latent defects rather than a regression.

## Changes

- Create `helixbot` before the AVD and run `avdmanager` as that user with `ANDROID_AVD_HOME` set, so the `.ini` and `.avd` land together and stay writable by the user that runs the emulator.
- Set `ANDROID_SDK_ROOT`, `ANDROID_HOME`, `ANDROID_AVD_HOME` and `PATH`.
- Copy the helix-scripts venv from the `venv` stage. The stage was built but its result was never used, unlike in the sibling helix images.
- Depend on `libicu` explicitly. XHarness runs on .NET, which aborts at startup without ICU, and the package was only present as a transitive dependency.
- Drop the container-local `kvm` group. It was created with a dynamically allocated gid (1000 in the published image) that does not match the gid owning `/dev/kvm` on the host, so it granted no access.
- Drop the `android-sdk` and `qemu-kvm` packages. The SDK is installed from the command line tools, and the Android emulator bundles its own qemu.

## API level as a build argument

The API level, system image target, ABI and AVD name become build arguments, so one Dockerfile can produce an image per API level instead of needing a copy of the file per level. **Each image still contains exactly one API level**; the level is selected by image tag.

This is what lets a single Helix queue cover every API level, instead of a dedicated queue and VM pool for each one. There are currently ~50 Linux Android queues, one family per API level.

Only **one** image is built for now, at **API 29**, the level used for daily Android testing. The existing unversioned tag is kept as an alias so nothing that references it breaks. Further levels are a manifest entry each and are worth adding once the approach is proven in CI.

## Validation

Built from this Dockerfile on an Ubuntu 24.04 host with KVM, then ran a real test suite against the emulator booted **inside the container**, started **unprivileged**:

| | |
| --- | --- |
| Emulator ready | **21 s** |
| `System.Numerics.Vectors.Tests` via XHarness | **7523 / 7523 passed**, exit 0 |
| Device reported | `"apiLevel": 29` |
| Artifacts | `testResults.xml`, logcat, emulator log, `xharness-result.json` |

The container was started with `--device /dev/kvm` plus the numeric gid owning the device, which is the mechanism proposed for the Helix client separately.

> [!NOTE]
> This pull request description was generated by GitHub Copilot.


src/ubuntu/24.04/helix/android/amd64/Dockerfile
```diff
@@ -32,9 +32,17 @@ RUN find /etc/apt -type f \( -name '*.sources' -o -name '*.list' \) -exec sed -i
 ARG PIP_INDEX_URL
 ENV PIP_INDEX_URL=${PIP_INDEX_URL}
 
+# The API level this image tests. Each API level gets its own image tag, so a single Helix queue
+# (Ubuntu.2204.Amd64.Android.Multi.*) can cover every level by selecting the matching image, and
```
akoeplinger: nit: let's not specify the Helix queue name here, it will just get outdated (e.g. I doubt we'd base new queues off of Ubuntu.2204 at this point).

more generally, the reason why we need a separate queue rather than using the normal ones is because of the kvm/nested virtualization setup. that might be useful for non-android cases as well so it might make sense to name the queue in a more general way.


```suggestion
# can cover every level by selecting the matching image, and
```

src/ubuntu/24.04/helix/android/amd64/Dockerfile
```diff
@@ -32,9 +32,17 @@ RUN find /etc/apt -type f \( -name '*.sources' -o -name '*.list' \) -exec sed -i
...
+# each work item gets an isolated emulator instead of a machine-wide shared one.
+# "default;x86_64" system images are published for every level from 21 to 36.
```
akoeplinger: I'd remove this line, the 21 to 36 isn't relevant here and will also just go stale

```suggestion
```

src/ubuntu/24.04/helix/android/amd64/Dockerfile
```diff
@@ -32,9 +32,17 @@ RUN find /etc/apt -type f \( -name '*.sources' -o -name '*.list' \) -exec sed -i
...
+ARG ANDROID_API_LEVEL=29
+ARG ANDROID_IMAGE_TARGET=default
+ARG ANDROID_ABI=x86_64
+ARG ANDROID_AVD_NAME=android_emu_86_64
```
akoeplinger: given we'll only ever have one emulator running we don't need to disambiguate. It would also be wrong if you e.g. pass in ANDROID_ABI=x86
```suggestion
ARG ANDROID_AVD_NAME=android_emu
```

src/ubuntu/24.04/helix/android/amd64/Dockerfile
```diff
@@ -48,6 +56,7 @@ RUN apt-get update && \
         iputils-ping \
         libnuma1 \
         libtool \
+        libicu74 \
```
akoeplinger: nit: other images just specify the -dev package so we don't need to hardcode the version

```suggestion
        libicu-dev \
```

we could also set the globalization invariant env var so we don't need icu, xharness shouldn't need locale-aware string parsing

akoeplinger (approved):

src/ubuntu/24.04/helix/android/amd64/Dockerfile
wfurt: Done, thanks — applied in e9a9daef.

src/ubuntu/24.04/helix/android/amd64/Dockerfile
wfurt: Done, thanks — applied in e9a9daef.

src/ubuntu/24.04/helix/android/amd64/Dockerfile
wfurt: Done, thanks — applied in e9a9daef.

src/ubuntu/24.04/helix/android/amd64/Dockerfile
wfurt: Done, thanks — applied in e9a9daef.

wfurt:
Thanks for the review — all four suggestions applied in e9a9daef, verified by rebuilding and re-running the suite in the container (7523/7523, emulator up in 21s).

Two notes on the broader points:

**Queue naming.** Agreed, and the KVM side is deliberately generic in the companion change: the queue opts in via a `linux-docker-kvm` artifact that writes `/etc/helix/docker-devices`, and the Helix client passes the device plus the owning gid to any container on that machine. Nothing in that mechanism is Android-specific, so another workload needing nested virtualization would reuse it as is. The queue we are targeting (`Ubuntu.2204.Amd64.Android.Multi`) already exists and is the only one with Docker plus the nested-virt SKU pin, so this change uses it rather than adding one, but the name is a leftover from an earlier consolidation attempt and no longer describes it — worth renaming separately.

**Globalization invariant instead of ICU.** Reasonable, but I kept ICU for now: the image only works today because libicu arrives as a transitive dependency, and the failure mode when it disappears is a startup `FailFast` with no test results, which is exactly what happens on 26.04. `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` would also apply to anything else run in the container, not just XHarness, so I would rather make the dependency explicit here and treat dropping ICU as a separate, measurable change.

> [!NOTE]
> This comment was drafted by GitHub Copilot on my behalf.

akoeplinger (approved):

PR #1746, opened 2026-09-17
Install pam/sudo in Azure Linux 4.0 build images

am11:
Unblocks https://github.com/dotnet/runtime/pull/134049.

AzDO container init runs `su -c ...` and fails with `su: Module is unknown`.

az4 images have pam-libs but not pam, so /etc/pam.d/su (shipped by util-linux)
has no modules to load. az3 got pam transitively via shadow-utils; on 4.0 that
dep was narrowed to pam-libs.

Verified in-image: installing pam -> su works.

fpm already has pam via sudo, hence only the three.


akoeplinger (approved):

akoeplinger:
> fpm already has pam via sudo, hence only the three.

maybe we should install sudo everywhere for consistency?

akoeplinger (approved):

PR #1748, opened 2026-09-18
Install missing deps in android and wasm AZ4 images

am11:
Azure Linux 4.0 base ships `glibc-minimal-langpack`, so `en_US.UTF-8`
does not exist and `javac` falls back to US-ASCII, failing on non-ASCII
source. The wasm image lacks `libatomic.so.1`, so the emsdk `node`
cannot start and every CMake `try_compile` link fails, leaving
`SIZEOF_*` empty in `config.h`.

Next thing blocking https://github.com/dotnet/runtime/pull/134049.


am11:
cc @akoeplinger

akoeplinger (approved):

PR #1751, opened 2026-09-19
Install self-contained PowerShell in AzL 4.0 images

am11:
The AzL 4.0 build and crossdeps images installed pwsh as a
framework-dependent .NET 10 tool. Arcade prepends the repo-local SDK
to PATH for every step after the build, so pwsh resolved its runtime
from there instead and failed to start:

```
  Framework: 'Microsoft.NETCore.App', version '10.0.0' (x64)
  .NET location: /__w/1/s/.dotnet/
```

That breaks every pwsh step in a Linux container job, such as the
cDAC dump and stress legs in dotnet/runtime#134049.

Install the self-contained upstream tarball instead, matching the
ubuntu, alpine and debian images, and drop the .NET 10 runtime that
was copied in only to host pwsh (AzL 3.0 ships no dotnet either).
Add gzip to the build image, which now unpacks a tarball.


am11:
cc @akoeplinger

akoeplinger (approved):

PR #1752, opened 2026-09-20
Validate powershell download with SHA256 checksum

am11:
cc @akoeplinger


akoeplinger:
we should ideally do this everywhere we download PS.

we should also add checksum updating for  the renovate bump PRs, like we do for llvm: https://github.com/dotnet/dotnet-buildtools-prereqs-docker/blob/e65f91a1d4d1db0a002ed6b4746cb39b848f1648/eng/renovate.json#L177

am11:
For renovate automation of this, we can do something like this as the hashes file is surprisingly UTF16 encoded.. :/

```sh
$ wget -O - "https://github.com/PowerShell/PowerShell/releases/download/v${POWERSHELL_VERSION}/hashes.sha256" |\
   iconv -f UTF-16LE -t UTF-8 | grep "powershell-${POWERSHELL_VERSION}-linux-x64.tar.gz" | tr -d '\r'
```

akoeplinger (approved):

